PR #6, opened 2020-03-22
[tests] Updating mock to unittest.mock

nepython:
I just noticed that `mock` wasn't specified in `requirements-test.txt`, so I have updated the tests using `mock` module to now use `unittest.mock` which is built-in. Also, I have narrowed down the scope of import to just `patch` as that was only being used.

### DIFF
--- a/openwisp_monitoring/check/tests/test_ping.py
+++ b/openwisp_monitoring/check/tests/test_ping.py
@@ -1,4 +1,5 @@
-import mock
+from unittest.mock import patch
+
 from django.core.exceptions import ValidationError
 from django.test import TransactionTestCase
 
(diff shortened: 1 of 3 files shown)


nepython:
@nemesisdesign, since travis has not yet been setup, I have manually run `runisort`, `runtests.py` and also doubled checked that there remain no `flake8` error. 
Everything is passing, though one test always fails **test_wifi_hostapd** it has no connection to this PR and was failing prior to it too. Might be a local problem as different data is stored in the database for the list under comparision.
```
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/monitoring/tests/test_graphs.py", line 242, in test_wifi_hostapd
    self.assertEqual(data['traces'][0][1][-10:], [0, 2, 2, 2, 2, 2, 2, 2, 2, 4])
AssertionError: Lists differ: [0, 2, 2, 2, 2, 2, 1, 2, 1, 4] != [0, 2, 2, 2, 2, 2, 2, 2, 2, 4]

First differing element 6:
1
2

- [0, 2, 2, 2, 2, 2, 1, 2, 1, 4]
?                    ---   ^

+ [0, 2, 2, 2, 2, 2, 2, 2, 2, 4]
?                       ^ +++
```
I am exploring the timeseries database more and would get back on this soon, if you know why this may be happening please mention it, it shall be greatly helpful. :smile:

nemesisdesign:
@nepython I don't know why that is happening, I also noticed it occasionally, can you create a dedicated issue for that please?

nepython:
@nemesisdesign, I have created #7 for this. Also, can you please review this PR :smile:

PR #8, opened 2020-03-30
[WIP] Mock few tests in test_api.py

nepython:
#### Testing time for test_device_api.py reduced from around **144s**-->**40s** on my pc
* I have tried to send mock response wherever possible, though I have left few cases where interaction with influxdb was essential for the test. These tests (which are not being mocked, approx. 4) are taking around 6-7s.
* I have also reduced the measurement creation for testing `test_device_admin` and `get_device_metrics_csv`(updated an assert in it accordingly, please check it) as they can be successfully tested even after a single measurement is created.
* ` _post_data` which is being used in almost all the tests was consuming the most time which can be pinpointed to writing increment in `_write` function of **DeviceMetricView**. There are other factors too so I have created a **mock_metric_view** keeping only the relevant lines. 
* **Need to resolve before merging**: Test coverage has fallen from 96% to 95% approx.
* Point of interest: I spent 3 days trying all available means of mocking `httpretty`, `requests/responses`, `model-bakery`, `httmock`, even changed caching methods and what not :smile:, finally to come to the conclusion it's best to mock responses yourself rather than depend on external modules. Nevertheless it was great as I now understand Metric model, DeviceData and how to query/write in **InfluxDB**
> **CC:** @nemesisdesign


openwisp_monitoring/device/tests/mock.py
```diff
@@ -0,0 +1,536 @@
...
+        if created:
+            if not 'wifi_clients' not in monitoring_settings.AUTO_GRAPHS:
+                graph = Graph(metric=metric,
+                              description=('{0} clients').format(metric.key),
+                              query="SELECT COUNT(DISTINCT({field_name})) AS value FROM {key} "
+                                    "WHERE time >= '{time}' AND content_type = '{content_type}' "
+                                    "AND object_id = '{object_id}' GROUP BY time(24h) fill(0)")
+                graph.full_clean()
+                graph.save()
+    return Response(status=200)
```
nemesisdesign: You either make the main code callable and use the main code or you don't do anything, you can't duplicate the logic of the views because that adds maintenance overhead.

PabloCastellano:
@nepython Regarding running tests faster I did read this article some while ago and I like it

https://hackernoon.com/timing-tests-in-python-for-fun-and-profit-1663144571

nepython:
@PabloCastellano, nice article indeed. I had tried out some methods in it back while working on this PR. This PR has gone stale as I started working on other issues. I will get back to this one after #43 and #44, thanks for sharing the article :smile:

PR #9, opened 2020-03-31
[enhancement] Delete monitoring objects when devices are deleted

### LINKED ISSUE #1: Delete monitoring objects when devices are deleted
When a device is deleted, its related metric and check (eg: ping) are not deleted.
Ideally we should delete the related objects automatically ~~and make it easy to recover these via django-reversion~~.

nepython:
Closes #1

### DIFF
--- a/openwisp_monitoring/check/apps.py
+++ b/openwisp_monitoring/check/apps.py
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from django.db.models.signals import post_delete
 from django.utils.translation import ugettext_lazy as _
 
 
@@ -6,3 +7,14 @@ class CheckConfig(AppConfig):
     name = 'openwisp_monitoring.check'
     label = 'check'
     verbose_name = _('Network Monitoring Checks')
+
+    def ready(self):
+        from ..device.models import Device
+        post_delete.connect(self.delete_related_check,
+                            sender=Device,
+                            dispatch_uid='delete_related_check')
+
+    @classmethod
+    def delete_related_check(cls, instance, **kwargs):
+        from .models import Check
+        Check.objects.filter(object_id=instance.pk).delete()
--- a/openwisp_monitoring/check/tests/test_ping.py
+++ b/openwisp_monitoring/check/tests/test_ping.py
@@ -101,26 +101,6 @@ class TestPing(TestDeviceMonitoringMixin, TransactionTestCase):
         result = check.perform_check(store=False)
         self.assertIsNone(result)
 
-    def test_device_deleted(self):
-        d = self._create_device(organization=self._create_org())
-        d.management_ip = '10.40.0.1'
-        d.save()
-        check = Check(name='Ping check', check=self._PING, content_object=d, params={})
-        check.full_clean()
-        check.save()
-        # dev = d
-        # import ipdb; ipdb.set_trace()
-        d.delete()
-        check = Check.objects.get(pk=check.pk)
-        result = check.perform_check(store=False)
-        self.assertIsNone(result)
-        try:
-            check.refresh_from_db()
-        except Check.DoesNotExist:
-            pass
-        else:
-            self.fail('check was not deleted')
-
     def test_content_object_none(self):
         check = Check(name='Ping check', check=self._PING, params={})
         try:
--- a/openwisp_monitoring/monitoring/apps.py
+++ b/openwisp_monitoring/monitoring/apps.py
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 from django.conf import settings
+from django.db.models.signals import post_delete
 from django.utils.translation import ugettext_lazy as _
 
 from .utils import create_database
@@ -14,3 +15,12 @@ class MonitoringConfig(AppConfig):
         # create influxdb database if doesn't exist yet
         create_database()
         setattr(settings, 'OPENWISP_ADMIN_SHOW_USERLINKS_BLOCK', True)
+        from ..device.models import Device
+        post_delete.connect(self.delete_related_metric,
+                            sender=Device,
+                            dispatch_uid='delete_related_metric')
+
+    @classmethod
+    def delete_related_metric(cls, instance, **kwargs):
+        from .models import Metric
+        Metric.objects.filter(object_id=instance.pk).delete()
--- a/openwisp_monitoring/monitoring/tests/test_models.py
+++ b/openwisp_monitoring/monitoring/tests/test_models.py
@@ -48,7 +48,7 @@ class TestModels(TestMonitoringMixin, TestCase):
         om = self._create_object_metric()
         self.assertEqual(om.key, om.codename)
 
-    def test_cutom_get_or_create(self):
+    def test_custom_get_or_create(self):
         m, created = Metric._get_or_create(name='lan', key='br-lan')
         self.assertTrue(created)
         m2, created = Metric._get_or_create(name='lan', key='br-lan')
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -135,9 +135,6 @@ class Ping(object):
         Figures out ip to use or fails raising OperationalError
         """
         device = self.related_object
-        if not device:
-            self.check_instance.delete()
-            return None
         ip = device.management_ip or device.last_ip
         return ip
 
--- a/openwisp_monitoring/check/tests/test_models.py
+++ b/openwisp_monitoring/check/tests/test_models.py
@@ -50,3 +50,10 @@ class TestModels(TestDeviceMonitoringMixin, TransactionTestCase):
         c = Check.objects.first()
         self.assertEqual(c.content_object, d)
         self.assertIn('Ping', c.check)
+
+    def test_device_deleted(self):
+        self.assertEqual(Check.objects.count(), 0)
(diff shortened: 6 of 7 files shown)


nepython:
@nemesisdesign, can you please check the minor change in `TestPing` it was essential after the enhancement.

openwisp_monitoring/check/models.py
```diff
@@ -70,6 +70,16 @@ def perform_check(self, store=True):
...
 
+    @classmethod
+    def delete_related_check(cls, instance, **kwargs):
+        """
+        Deletes check related to a device upon device's deletion
+        """
+        device = instance
+        checks = cls.objects.filter(object_id=device.pk)
+        if checks:
+            checks.delete()
```
nemesisdesign, line 81: keep it simple:

```python
cls.objects.filter(object_id=instance.pk).delete()
```

openwisp_monitoring/check/models.py
```diff
@@ -70,6 +70,16 @@ def perform_check(self, store=True):
         """
         return self.check_instance.check(store=True)
 
+    @classmethod
+    def delete_related_check(cls, instance, **kwargs):
```
nemesisdesign, line 74: Let's put this method in check/apps.py to make the whole flow easier to read.

openwisp_monitoring/monitoring/models.py
```diff
@@ -239,6 +239,16 @@ def _set_extra_notification_opts(self, opts):
...
 
+    @classmethod
+    def delete_related_metric(cls, instance, **kwargs):
+        """
+        Deletes metrics related to a device upon device's deletion
+        """
+        device = instance
+        metrics = cls.objects.filter(object_id=device.pk)
+        if metrics:
+            metrics.delete()
```
nemesisdesign, line 250: same as in checks, simplify the delete operation (can be done in 1 line) and move it to apps.py to make the whole flow easier to read

openwisp_monitoring/monitoring/tests/test_models.py
```diff
@@ -48,7 +50,7 @@ def test_object_key(self):
         om = self._create_object_metric()
         self.assertEqual(om.key, om.codename)
 
-    def test_cutom_get_or_create(self):
+    def test_custom_get_or_create(self):
```
nemesisdesign, line 53: thanks

openwisp_monitoring/monitoring/tests/test_models.py
```diff
@@ -68,6 +70,22 @@ def test_general_write(self):
         measurement = list(query('select * from sync_test').get_points())[0]
         self.assertEqual(measurement['value'], 1)
 
+    def test_device_deleted(self):
```
nemesisdesign, line 73: this test can be moved to `openwisp_monitoring.device`, since it's that app that creates the device metrics.

Those tests have helper methods like `_create_device` which you can use.
To create a metric, use `_create_object_metric`.

openwisp_monitoring/check/tests/test_ping.py
```diff
@@ -109,8 +109,8 @@ def test_device_deleted(self):
         check.save()
         # dev = d
         # import ipdb; ipdb.set_trace()
-        d.delete()
         check = Check.objects.get(pk=check.pk)
+        d.delete()
```
nemesisdesign, line 113: so there was already a test here? Why was it passing, do you have any idea?

nepython:
@nemesisdesign, I have moved the functions into `apps.py` to make the flow easier and tried to recover deleted metrics via django-reversion. I had registered `Metric` model with `reversion` and as per the docs used `reversion.create_revision()` while creating the version. The problem is, the `metric` is not getting saved, I have checked it thoroughly with docs, everything seems fine. Any, suggestion on why it might not be getting saved?
```
    pre_delete.connect(self.set_recover_version,
                        sender=Device,
                        dispatch_uid='set_metric_recovery_version')
    post_delete.connect(self.delete_related_metric,
                        sender=Device,
                        dispatch_uid='delete_related_metric')
    post_save.connect(self.recover_related_metric,
                        sender=Device,
                        dispatch_uid='recover_related_metric')

    @classmethod
    def set_recover_version(cls, instance, **kwargs):
        from .models import Metric
        d = instance
        metrics = Metric.objects.filter(object_id=d.pk)
        with reversion.create_revision():
            for metric in metrics:
                metric.save()
                reversion.set_comment(str(d.pk)) 
                # I have checked right here and my VersionQueryset didn't contain metric object
    
    @classmethod
    def recover_related_metric(cls, instance, **kwargs):
        """
        Recovers metrics related to a device upon device's recovery
        """
        from ..device.models import Device
        d = instance
        metrics = Version.objects.get_deleted(Metric)
        for metric in metrics:
            if metric.revision.comment == str(d.pk):
                metric.revision.revert()

    @classmethod
    def delete_related_metric(cls, instance, **kwargs):
        """
        Deletes metrics related to a device upon device's deletion
        """
        from .models import Metric
        device = instance
        metrics = Metric.objects.filter(object_id=device.pk)
        if metrics:
            metrics.delete()
```

openwisp_monitoring/check/tests/test_ping.py
nepython: I am unsure but it looks like `perform_check` triggered the `check`'s deletion though on backtracing I didn't find anything to trigger check's deletion besides `content_type` :smiley:

nemesisdesign: this is weird. I'll investigate myself once the rest of the PR is ready. Thanks

nepython:
Then the PR is ready if django-reversion is not to be implemented anymore.
> this is weird. I'll investigate myself once the rest of the PR is ready. Thanks

Humble request, if you find any good reason for it please do share!

nepython:
@nemesisdesign, I think this PR is ready. Though I have tested it multiple times please do test it I don't want to break something again :sweat_smile:

nepython:
It's done @nemesisdesign :smiley:

nepython:
@nemesisdesign rebased, hopefully it gets merged now. Accidently closed the PR :sweat_smile:

nepython:
@nemesisdesign, this PR is ready for review.

openwisp_monitoring/check/tests/test_ping.py
```diff
@@ -110,8 +110,8 @@ def test_device_deleted(self):
...
```
nemesisdesign, line 114: I now understood why you had to do this.
Now the related check will be deleted and hence the get query will fail.

The test written here was before passing but it was wrong.

Let's improve this part:

- this test doesn't have reason to be anymore, especially here in the `test_ping` file, 
- let's create a new test in `test_models` with the following code, which fails correctly before applying your patch:

```python
def test_device_deleted(self):
        self.assertEqual(Check.objects.count(), 0)
        self.assertEqual(Check.objects.count(), 1)
        d.delete()
        self.assertEqual(Check.objects.count(), 0)
```

openwisp_monitoring/check/tests/test_ping.py
```diff
@@ -110,8 +110,8 @@ def test_device_deleted(self):
         check.save()
         # dev = d
         # import ipdb; ipdb.set_trace()
```
nemesisdesign, line 112: could you remove these comments please?

openwisp_monitoring/check/tests/test_ping.py
nepython: I now understand why the `Check` was earlier getting deleted and what this test was checking out. There's a check in `_get_ip()` which was deleting the Check if the device had been deleted :blush:. I think that check can be safely removed as the function won't get triggered now due to Check being deleted.

openwisp_monitoring/check/tests/test_ping.py
nepython: Since, the entire test has been removed as per your above suggestion, it's been automatically removed :sweat_smile:

nepython:
Also, test-wifi-hostapd has started failing again, my build failed 2 times so something needs to be done in the short run :sweat_smile:

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,20 @@
...
+    def ready(self):
+        from ..device.models import Device
+        post_delete.connect(self.delete_related_check,
+                            sender=Device,
+                            dispatch_uid='delete_related_check')
+
+    @classmethod
+    def delete_related_check(cls, instance, **kwargs):
+        from .models import Check
+        Check.objects.filter(object_id=instance.pk).delete()
```
nemesisdesign, line 20: this query needs to specify the `content_type` as well, I will handle this change on my own

openwisp_monitoring/monitoring/apps.py
```diff
@@ -14,3 +15,12 @@ def ready(self):
...
         setattr(settings, 'OPENWISP_ADMIN_SHOW_USERLINKS_BLOCK', True)
+        from ..device.models import Device
+        post_delete.connect(self.delete_related_metric,
+                            sender=Device,
+                            dispatch_uid='delete_related_metric')
+
+    @classmethod
+    def delete_related_metric(cls, instance, **kwargs):
+        from .models import Metric
+        Metric.objects.filter(object_id=instance.pk).delete()
```
nemesisdesign, line 26: this query needs to specify the `content_type` as well, I will handle this change on my own

PR #17, opened 2020-04-04
[monitoring] Added migration to create ping for existing devices. #16

### LINKED ISSUE #16: [monitoring] Add a data-migration to create ping for existing devices
Add a data-migration to create ping for existing devices.

Useful when the monitoring module is added afterwards.

This is the pseudo-code to run in the data migration

```python
from openwisp_monitoring.check.tasks import auto_create_ping

for device in Device.objects.all():
    auto_create_ping(
        model=Device.__name__.lower(),
        app_label=Device._meta.app_label,
        object_id=str(device.pk),
        created=True
    )
```

TheOneAboveAllTitan:
Closes #16

`flake8` :heavy_check_mark: 
`isort`. :heavy_check_mark: 

Steps I took to check : 
1. After migration ran `python tests/manage.py migrate monitoring 0004`
2. Create few device using admin interface
3. Applied `0005_create_ping` migration,
4. Verified that ping checks were created.

### DIFF
--- a/openwisp_monitoring/check/migrations/0003_create_ping.py
+++ b/openwisp_monitoring/check/migrations/0003_create_ping.py
@@ -1,0 +1,26 @@
+from django.db import migrations
+from openwisp_monitoring.check.settings import AUTO_PING
+from openwisp_monitoring.check.tasks import auto_create_ping
+
+
+def create_device_ping(apps, schema_editor):
+    if AUTO_PING:
+        Device = apps.get_model('config', 'Device')
+        for device in Device.objects.all():
+            auto_create_ping(model=Device.__name__.lower(),
+                            app_label=Device._meta.app_label,
+                            object_id=str(device.pk),
+                            created=True)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('check', '0002_check_unique_together'),
+        ('config', '0026_hardware_id_not_unique')
+    ]
+
+    operations = [
+        migrations.RunPython(create_device_ping,
+                             reverse_code=migrations.RunPython.noop),
+    ]


openwisp_monitoring/monitoring/migrations/0005_create_ping.py
```diff
@@ -0,0 +1,25 @@
...
+from openwisp_controller.config.models import Device
+
+
+def create_device_ping(apps, schema_editor):
+    for device in Device.objects.all():
+        auto_create_ping(model=Device.__name__.lower(),
+                         app_label=Device._meta.app_label,
+                         object_id=str(device.pk),
+                         created=True
+        )
```
nemesisdesign, line 13: indentation here looks odd, can you move this parenthesis at the end of the previous line?

openwisp_monitoring/monitoring/migrations/0005_create_ping.py
```diff
@@ -0,0 +1,25 @@
+from django.db import migrations
+from openwisp_monitoring.check.tasks import auto_create_ping
+
+from openwisp_controller.config.models import Device
```
nemesisdesign, line 4: don't do this, use the migration framework to get the Device model, it's explained here: https://docs.djangoproject.com/en/3.0/topics/migrations/#data-migrations

See `apps.get_model`, there's an example with a related comment which explains why you have to use that instead of the concrete model.

TheOneAboveAllTitan: Thanks for pointing towards right resources. :smile:

openwisp_monitoring/monitoring/migrations/0005_create_ping.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations
```
nemesisdesign, line 1: this migration should go in the check app and should be executed only if `openwisp_monitoring.check.settings.AUTO_PING` is `True`.

openwisp_monitoring/check/migrations/0003_create_ping.py
```diff
@@ -0,0 +1,26 @@
...
+from openwisp_monitoring.check.settings import AUTO_PING
+from openwisp_monitoring.check.tasks import auto_create_ping
+
+
+def create_device_ping(apps, schema_editor):
+    if AUTO_PING:
+        Device = apps.get_model('config', 'Device')
+        for device in Device.objects.all():
+            auto_create_ping(model=Device.__name__.lower(),
+                            app_label=Device._meta.app_label,
```
nemesisdesign, line 11: indentation is off by one space, but let's not worry about this anymore since it will be handled by the black formatter soon.

Issue #14, opened 2020-04-04
React to metric collection and checksum signals


nemesisdesign:
## 1. Listen to the health status change signal

Whenever the status moves into critical, a cache key with the device UUID is created, eg:

`device-<pk>-react-to-updates` which should be set to `1`

If the health status moves to any other status, the key is set to `0' 

## 2. Emit a new signal when device metrics are received

Now add the code to emit a signal when device metrics are collected (`DeviceMetricView`, before a succesful response is returned).

## 3. React to the signals

Add a receiver for this new signal and for the device configuration checksum signal described here: https://github.com/openwisp/django-netjsonconfig/issues/174

The receiver connected to those signals should check the cache key described above and if  the value is 1, we shall queue the celery task to be run in the background, if it's 0 we skip.

The celery task shall find out if there's any check related to the device, if there is any it shall trigger the checks to be run in the background.

If there's no related check, move the status to OK.

The goal of this issue is to make the monitoring system faster in detecting recoveries because checks are run periodically (eg: 5 minutes). With this new solution we'll be able to find out recoveries as soon as they happen.

## 4. Behaviour configurable

Add a setting which controls this behaviour, defaulting to `True`.
If set to `False` these reactions will be disabled.

The name of the setting can be: `OPENWISP_MONITORING_DEVICE_REACTIONS`.

Do not forget to add a mention to this setting in the README.

nepython:
Working on it :smiley:

PR #20, opened 2020-04-10
[feature] Listen to DeviceConnection.is_working changes #4

### LINKED ISSUE #4: [feature] Listen to DeviceConnection.is_working changes
To be done after: https://github.com/openwisp/openwisp-controller/issues/174

This is an idea/draft, the whole flow should be revisited and improved.

Listen to the `DeviceConnection.is_working` change signal and perform the following operations:

- if the device connection has switched from working to not working:
  - set the device monitoring status to `PROBLEM`
  - create a related notification, include the information about why the connection is not working
  - if the monitoring status was set to `OK` (before we set it to problem) we should find out if there's any check existing for this device, if yes, trigger the checks (will be run in the background by celery)
- if the device connection has switched from not working to working
  - create a related notification
  - if there's any check related to this device, trigger the checks, if there's no available check, set the monitoring status to OK

nepython:
Closes #4 
@nemesisdesign, this PR is not completely ready yet, I just wanted to confirm is this what you had in mind for the issue or any changes are required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ before_install:
 
 install:
   - pip install -e .
+  # To be removed after new release of openwisp-controller
+  - pip install --upgrade https://github.com/openwisp/openwisp-controller/tarball/master
   # Until release of django-netjsonconfig 0.12.0
   - pip install --upgrade https://github.com/openwisp/django-netjsonconfig/tarball/master
   # temporary: remove when openwisp-utils 0.5.0 is released
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -4,6 +4,7 @@ from django.db.models.signals import post_delete
 from django.utils.translation import ugettext_lazy as _
 from django_netjsonconfig.signals import checksum_requested
 
+from ..monitoring.utils import notify_users
 from . import settings as app_settings
 from .signals import device_metrics_received, health_status_changed
 from .utils import get_device_recovery_cache_key, manage_short_retention_policy
@@ -16,6 +17,7 @@ class DeviceMonitoringConfig(AppConfig):
 
     def ready(self):
         manage_short_retention_policy()
+        self.connect_is_working_changed()
         self.connect_device_post_delete()
         self.device_recovery_detection()
 
@@ -69,7 +71,46 @@ class DeviceMonitoringConfig(AppConfig):
 
     @classmethod
     def trigger_device_recovery_checks(cls, instance, **kwargs):
-        from .tasks import trigger_device_recovery
+        from .tasks import trigger_device_checks
 
         if cache.get(get_device_recovery_cache_key(device=instance)):
-            trigger_device_recovery.delay(pk=instance.pk)
+            trigger_device_checks.delay(pk=instance.pk)
+
+    @classmethod
+    def connect_is_working_changed(self):
+        from openwisp_controller.connection.models import DeviceConnection
+        from openwisp_controller.connection.signals import is_working_changed
+
+        is_working_changed.connect(
+            self.is_working_changed_receiver,
+            sender=DeviceConnection,
+            dispatch_uid='is_working_changed_monitoring',
+        )
+
+    @classmethod
+    def is_working_changed_receiver(cls, instance, is_working, **kwargs):
+        from ..check.models import Check
+        from .tasks import trigger_device_checks
+
+        device = instance.device
+        device_monitoring = device.monitoring
+        initial_status = device_monitoring.status
+        status = 'ok' if is_working else 'problem'
+        if not is_working:
+            # Create a related notification explaining why it's not working
+            desc = instance.failure_reason
+            opts = dict(
+                actor=device, level='warning', verb='not working', description=desc
+            )
+            if initial_status == 'ok':
+                trigger_device_checks.delay(pk=device.pk)
+        else:
+            # create a notification that device is working
+            opts = dict(actor=device, level='info', verb='connected successfully')
+            # if checks exist trigger them else, set status as 'ok'
+            if Check.objects.filter(object_id=instance.device.pk).exists():
+                trigger_device_checks.delay(pk=device.pk)
+            else:
+                device_monitoring.update_status(status)
+        opts['data'] = {'email_subject': f'[{status.upper()}] {device}'}
+        notify_users(opts)
--- a/openwisp_monitoring/device/tests/test_models.py
+++ b/openwisp_monitoring/device/tests/test_models.py
@@ -1,9 +1,13 @@
 import json
+from unittest.mock import patch
 
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
+from swapper import load_model
 
+from openwisp_controller.connection.models import Credentials, DeviceConnection
 from openwisp_utils.tests import catch_signal
 
+from ...check.models import Check
 from ...monitoring.utils import get_db
 from .. import settings as app_settings
 from ..models import DeviceData, DeviceMonitoring
@@ -320,3 +324,54 @@ class TestDeviceMonitoring(BaseTestCase):
         self.assertEqual(dm.status, 'problem')
         load.check_threshold(20)
         self.assertEqual(dm.status, 'ok')
+
+    def test_device_connection_change(self):
+        admin = self._create_admin()
+        Notification = load_model('notifications', 'Notification')
+        d = self._create_device()
+        dm = d.monitoring
+        dm.status = 'unknown'
+        dm.save()
+        c = Credentials.objects.create()
+        dc = DeviceConnection.objects.create(credentials=c, device=d)
+        dc.is_working = True
+        dc.save()
+        self.assertEqual(dm.status, 'ok')
+        n = Notification.objects.get(level='info')
+        self.assertEqual(n.verb, 'connected successfully')
+        self.assertEqual(n.actor, d)
+        dc.is_working = False
+        dc.save()
+        # self.assertEqual(dm.status, 'problem')
+        n = Notification.objects.get(level='warning')
+        self.assertEqual(n.verb, 'not working')
+        self.assertEqual(n.actor, d)
+        self.assertEqual(n.recipient, admin)
+
+    @patch('openwisp_monitoring.check.models.Check.perform_check')
+    def test_is_working_false_true(self, mocked_call):
+        d = self._create_device()
+        dm = d.monitoring
+        dm.status = 'unknown'
+        dm.save()
+        ping_path = 'openwisp_monitoring.check.classes.Ping'
+        Check.objects.create(name='Check', content_object=d, check=ping_path)
+        c = Credentials.objects.create()
+        dc = DeviceConnection.objects.create(credentials=c, device=d)
+        dc.is_working = True
+        dc.save()
+        mocked_call.assert_called_once()
+
+    @patch('openwisp_monitoring.check.models.Check.perform_check')
+    def test_is_working_true_false(self, mocked_call):
+        d = self._create_device()
+        dm = d.monitoring
+        dm.status = 'ok'
+        dm.save()
+        ping_path = 'openwisp_monitoring.check.classes.Ping'
+        Check.objects.create(name='Check', content_object=d, check=ping_path)
+        c = Credentials.objects.create()
+        dc = DeviceConnection.objects.create(credentials=c, device=d)
+        dc.is_working = False
+        dc.save()
+        mocked_call.assert_called_once()
--- a/openwisp_monitoring/monitoring/utils.py
+++ b/openwisp_monitoring/monitoring/utils.py
@@ -1,7 +1,10 @@
 import logging
 from datetime import datetime
 
+from django.contrib.auth import get_user_model
+from django.db.models import Q
 from influxdb import client
+from swapper import load_model
 
 from . import settings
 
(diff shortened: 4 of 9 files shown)


nepython:
@nemesisdesign, this PR is ready for review. I am not very sure of notifications though as I have not priorily used django-notifications so implemented it as best I could understand from it's documentation. Also, were checks supposed to be triggered as they have been in this PR (took inspiration from `Threshold`) :smile:

.travis.yml
```diff
@@ -41,6 +41,8 @@ before_install:
 install:
   - pip install $DJANGO
   - pip install -e .
+  # To be removed after new release of openwisp-controller
+  - pip install --upgrade https://github.com/nepython/openwisp-controller/tarball/master
```
TheOneAboveAllTitan, line 45: Use OpenWISP's repository if possible.

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+        from openwisp_controller.connection.signals import is_working_changed
+        is_working_changed.connect(self.health_status_changed,
+                                   sender=DeviceConnection,
+                                   dispatch_uid='change_health_status')
+
+    @classmethod
+    def health_status_changed(cls, instance, is_working, **kwargs):
+        from ..check.models import Check
+        d = instance.device
+        dm = d.monitoring
```
TheOneAboveAllTitan, line 28: More descriptive identifiers will make it more readable.

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+        initial_status = dm.status
+        status = 'ok' if is_working else 'problem'
```
TheOneAboveAllTitan, line 30: It is generally not recommended to write conditional blocks in one lines since it affects code coverage reports.

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+                    Check.objects.get(object_id=d.pk)
+                    management.call_command('run_checks')
+                except Check.DoesNotExist:
+                    pass
+        else:
+            # create a notification that device is working
+            opts = dict(actor=d, level='info', verb='connected successfully')
+            try:
+                # if checks exist trigger them else, set status as 'ok'
+                Check.objects.get(object_id=instance.device.pk)
```
TheOneAboveAllTitan, line 48: This will run checks on all devices instead of only this one. Please correct me if I am missing something.

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+            User.objects.select_related('notificationuser')
+            .order_by('date_joined')
+            .filter(where)
+        )
+        Notification = load_model('notifications', 'Notification')
+        for user in qs:
+            n = Notification(**opts)
+            n.recipient = user
+            n.full_clean()
+            n.save()
```
TheOneAboveAllTitan, line 71: Did you considered using `notify` signal as mentioned in the [django-notifications documentation](https://github.com/django-notifications/django-notifications#generating-notifications)?

openwisp_monitoring/device/apps.py
nepython: Here, it stays unaffected. Still, I wasn't aware of it, thanks for sharing. Can you please share the source, I would like to know why so.

.travis.yml
nepython: I know, will do after a PR gets merged.

openwisp_monitoring/device/apps.py
nepython: Yep, it shouldn't be the case but I have kept it for now. The reason being, I have created a new task for #14 and will use the same here later on. I wanted to avoid duplication so did away with them here.

openwisp_monitoring/device/apps.py
nepython: Yes, spent some good time understanding and deciding upon it. The reason I have done it in this manner is as I wanted to maintain consistency with what has been done already. Anyhow, this reminds me that it would be best if we can have such a function in the notifications app maybe notification module is going to have an independent existence for best :smile:

openwisp_monitoring/device/apps.py
nepython: I am not sure about this, it appears all right to me

nepython:
@TheOneAboveAllTitan, I think with this all your doubts should be cleared. If they aren't feel free to ask them on gitter. Also, thanks for going through this PR :smile:

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
     def ready(self):
         manage_short_retention_policy()
+        from openwisp_controller.connection.models import DeviceConnection
+        from openwisp_controller.connection.signals import is_working_changed
+        is_working_changed.connect(self.health_status_changed,
+                                   sender=DeviceConnection,
+                                   dispatch_uid='change_health_status')
+
+    @classmethod
+    def health_status_changed(cls, instance, is_working, **kwargs):
```
nemesisdesign, line 25: call it `is_working_changed_receiver`

openwisp_monitoring/device/apps.py
nemesisdesign: do not use abbreviated variable names in non test code please

openwisp_monitoring/device/apps.py
nemesisdesign: this is ok

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+                    management.call_command('run_checks')
+                except Check.DoesNotExist:
+                    pass
+        else:
+            # create a notification that device is working
+            opts = dict(actor=d, level='info', verb='connected successfully')
+            try:
+                # if checks exist trigger them else, set status as 'ok'
+                Check.objects.get(object_id=instance.device.pk)
+                management.call_command('run_checks')
```
nemesisdesign, line 49: this will call for all devices

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
 
     def ready(self):
         manage_short_retention_policy()
+        from openwisp_controller.connection.models import DeviceConnection
```
nemesisdesign, line 18: I'd move this code in its own method and make sure the feature can be disabled with a setting

openwisp_monitoring/device/apps.py
```diff
@@ -11,3 +15,57 @@ class DeviceMonitoringConfig(AppConfig):
...
+        if not is_working:
+            dm.update_status(status)
+            # Create a related notification explaining why it's not working
+            desc = instance.failure_reason
+            opts = dict(actor=d, level='warning', verb='not working', description=desc)
+            if initial_status == 'ok':
+                try:
+                    # if checks exist trigger them
+                    Check.objects.get(object_id=d.pk)
+                    management.call_command('run_checks')
```
nemesisdesign, line 40: will execute all checks, we must execute only checks of the device

nepython: @nemesisdesign, I will update this after PR`#22` as there I have created a `find_related_checks` task for the same purpose. :smile:

openwisp_monitoring/device/apps.py
nepython: @nemesisdesign, I would have added an `if` condition prior to it, something like this
```
if getattr(settings, <setting-name>, True):
    # Rest of the code ... 
```
You have pointed it out correctly, this feature should have a setting attribute. Should I add it, something with name `OPENWISP_MONITORING_LISTEN_DEVICE_CONNECTION` ?

nepython:
@nemesisdesign, I have iterated this PR after #22 got merged and it's ready for review :smile:

openwisp_monitoring/device/apps.py
```diff
@@ -73,3 +77,63 @@ def trigger_device_recovery_checks(cls, instance, **kwargs):
...
+
+    @classmethod
+    def connect_is_working_changed(self):
+        from openwisp_controller.connection.models import DeviceConnection
+        from openwisp_controller.connection.signals import is_working_changed
+
+        is_working_changed.connect(
+            self.is_working_changed_receiver,
+            sender=DeviceConnection,
+            dispatch_uid='change_health_status',
```
nemesisdesign, line 89: I think this should be:

```python
dispatch_uid='is_working_changed_monitoring',
```

openwisp_monitoring/device/apps.py
```diff
@@ -73,3 +77,63 @@ def trigger_device_recovery_checks(cls, instance, **kwargs):
...
+                actor=device, level='warning', verb='not working', description=desc
+            )
+            if initial_status == 'ok':
+                trigger_device_recovery.delay(pk=device.pk)
+        else:
+            # create a notification that device is working
+            opts = dict(actor=device, level='info', verb='connected successfully')
+            try:
+                # if checks exist trigger them else, set status as 'ok'
+                Check.objects.get(object_id=instance.device.pk)
```
nemesisdesign, line 115: if you only need to check whether an element is present, use exists() which is more efficient because designed for this purpose: https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.exists

openwisp_monitoring/device/apps.py
```diff
@@ -73,3 +77,63 @@ def trigger_device_recovery_checks(cls, instance, **kwargs):
...
+                trigger_device_recovery.delay(pk=device.pk)
+            except Check.DoesNotExist:
+                device_monitoring.update_status(status)
+        opts['data'] = {'email_subject': f'[{status.upper()}] {device}'}
+        cls._notify_users(opts)
+
+    @classmethod
+    def _notify_users(cls, opts):
```
nemesisdesign, line 123: this code looks duplicated from `Metric._notify_users`. **Do not duplicate code**.
Create a function which handles both cases and call that in these two places.
Duplicating code is the best way to maintenance hell. Refrain yourself from doing it.

openwisp_monitoring/device/apps.py
```diff
@@ -73,3 +77,63 @@ def trigger_device_recovery_checks(cls, instance, **kwargs):
...
+    def is_working_changed_receiver(cls, instance, is_working, **kwargs):
+        from ..check.models import Check
+        from .tasks import trigger_device_recovery
+
+        device = instance.device
+        device_monitoring = device.monitoring
+        initial_status = device_monitoring.status
+        status = 'ok' if is_working else 'problem'
+        if not is_working:
+            device_monitoring.update_status(status)
```
nemesisdesign, line 102: The code in `trigger_device_recovery` also changes the status so we may be changing the status twice in a matter of milliseconds. 
I would delegate the health status change entirely to that task, it doesn't matter if we flag it as problem or critical, we'll fine tune this in the future as everything becomes clear.

I would rename the task to `trigger_device_checks` and just call that, eg:

```python
trigger_device_recovery.delay(pk=device.pk)
# ... prepare notification data ... 
if not is_working:
   # ... prepare notification data
else:
   # ...
notify_users()
```

nepython: @nemesisdesign, in the issue description it was mentioned to trigger checks only on specific instances like:
If device is not working & previous status was working only then trigger checks
While in your suggestion above related `checks` are triggered for all cases so what is required?

openwisp_monitoring/device/apps.py
nepython: I have tried to avoid duplication for most of the code except for creating the `opts`, it was deferring slightly for both though it has a workaround I have left these lines in the function `_notify_users` also to maintain backward compatibality for tests in the `notifications` app

openwisp_monitoring/device/models.py
```diff
@@ -68,9 +68,9 @@ def data_user_friendly(self):
                 remove.append(interface)
                 continue
             # human readable mode
-            interface['wireless']['mode'] = interface['wireless']['mode'].replace(
-                '_', ' '
-            )
+            # interface['wireless']['mode'] = interface['wireless']['mode'].replace(
+            #     '_', ' '
+            # )
```
PabloCastellano, line 73: Why is it commented? If it's not useful then let's remove it

openwisp_monitoring/monitoring/utils.py
```diff
@@ -60,3 +63,28 @@ def create_database():
...
+    # perform query
+    User = get_user_model()
+    qs = (
+        User.objects.select_related('notificationuser')
+        .order_by('date_joined')
+        .filter(where)
+    )
+    Notification = load_model('notifications', 'Notification')
+    for user in qs:
+        n = Notification(**opts)
```
PabloCastellano, line 87: For simplicity I would to `n = Notification(recipient=user, **opts)`

openwisp_monitoring/device/models.py
nepython: I was testing another template which didn't have `wireless` in `Devicedata` but it is useful

openwisp_monitoring/device/apps.py
```diff
@@ -69,7 +71,46 @@ def manage_device_recovery_cache_key(cls, instance, status, **kwargs):
...
+        is_working_changed.connect(
+            self.is_working_changed_receiver,
+            sender=DeviceConnection,
+            dispatch_uid='is_working_changed_monitoring',
+        )
+
+    @classmethod
+    def is_working_changed_receiver(cls, instance, is_working, **kwargs):
+        from ..check.models import Check
+        from .tasks import trigger_device_checks
```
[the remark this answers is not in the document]
nepython: That can't be done, because Django loads apps first followed by all the models in the app.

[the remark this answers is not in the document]
PabloCastellano: Okay I suspected that, ignore it then :)

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -320,3 +324,54 @@ def test_unknown_problem_ok(self):
...
+        dc = DeviceConnection.objects.create(credentials=c, device=d)
+        dc.is_working = True
+        dc.save()
+        self.assertEqual(dm.status, 'ok')
+        n = Notification.objects.get(level='info')
+        self.assertEqual(n.verb, 'connected successfully')
+        self.assertEqual(n.actor, d)
+        dc.is_working = False
+        dc.save()
+        # self.assertEqual(dm.status, 'problem')
```
nemesisdesign, line 345: you forgot one comment here

Issue #7, opened 2020-03-24
[bug] test_wifi_hostapd is failing


nepython:
**test_wifi_hostapd** is currently failing almost everytime on my end and occasionally on @nemesisdesign's end. 
It might be a local problem and might vary from PC to PC as can be concluded from above.
The full stack trace of the failing test is:
```
Traceback (most recent call last):
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/monitoring/tests/test_graphs.py", line 242, in test_wifi_hostapd
    self.assertEqual(data['traces'][0][1][-10:], [0, 2, 2, 2, 2, 2, 2, 2, 2, 4])
AssertionError: Lists differ: [0, 2, 2, 2, 2, 2, 1, 2, 1, 4] != [0, 2, 2, 2, 2, 2, 2, 2, 2, 4]

First differing element 6:
1
2

- [0, 2, 2, 2, 2, 2, 1, 2, 1, 4]
?                    ---   ^

+ [0, 2, 2, 2, 2, 2, 2, 2, 2, 4]
?                       ^ +++
```

**Few additional details**
* OS : Ubuntu 18.04 LTS
* I am running this in a virtual environment with all dependencies installed.
* The list being generated is not the same in every run and this should not be happening.
* This is the only failing test currently in this module.

nepython:
**UPDATE**
This test passes on most ocassions on Travis but it does fail sometimes https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/673729543

PR #22, opened 2020-04-13
[feature] Detect recoveries spontaneously #14

### LINKED ISSUE #14: React to metric collection and checksum signals

nepython:
Closes #14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ before_install:
 
 install:
   - pip install -e .
+  # Until release of django-netjsonconfig 0.12.0
+  - pip install --upgrade https://github.com/openwisp/django-netjsonconfig/tarball/master
   # temporary: remove when openwisp-utils 0.5.0 is released
   - pip install -U https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils[qa]
 
--- a/README.rst
+++ b/README.rst
@@ -302,6 +302,37 @@ If the devices are connecting to your OpenWISP instance using a shared layer2
 network, hence the OpenWSP server can reach the devices using the ``last_ip``
 field, you can set this to ``False``.
 
+``OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+The setting allows to detect recoveries as soon as they happen thus making
+the monitoring system faster.
+
+Signals
+-------
+
+``device_metrics_received``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Path**: ``openwisp_monitoring.device.signals.device_metrics_received``
+
+**Arguments**:
+
+- ``instance``: instance of ``Device`` whose metrics have been received
+- ``request``: the HTTP request object
+
+This signal is emitted when device metrics are received via the ``DeviceMetric`` view.
+
+The signal is emitted just before a successful response is returned,
+it is not sent if the response was not successful.
+
+
 Installing for development
 --------------------------
 
--- a/openwisp_monitoring/check/tasks.py
+++ b/openwisp_monitoring/check/tasks.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import, unicode_literals
-
 import json
 
 from celery import shared_task
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -20,6 +20,7 @@ from ... import settings as monitoring_settings
 from ...monitoring.models import Graph, Metric
 from ..models import DeviceData
 from ..schema import schema
+from ..signals import device_metrics_received
 
 
 class DevicePermission(BasePermission):
@@ -118,6 +119,9 @@ class DeviceMetricView(GenericAPIView):
             self._write(request, self.instance.pk)
         except ValidationError as e:
             return Response(e.message_dict, status=status.HTTP_400_BAD_REQUEST)
+        device_metrics_received.send(
+            sender=self.model, instance=self.instance, request=request
+        )
         return Response(None)
 
     def _init_previous_data(self):
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -1,8 +1,12 @@
 from django.apps import AppConfig
+from django.core.cache import cache
 from django.db.models.signals import post_delete
 from django.utils.translation import ugettext_lazy as _
+from django_netjsonconfig.signals import checksum_requested
 
-from .utils import manage_short_retention_policy
+from . import settings as app_settings
+from .signals import device_metrics_received, health_status_changed
+from .utils import get_device_recovery_cache_key, manage_short_retention_policy
 
 
 class DeviceMonitoringConfig(AppConfig):
@@ -13,6 +17,7 @@ class DeviceMonitoringConfig(AppConfig):
     def ready(self):
         manage_short_retention_policy()
         self.connect_device_post_delete()
+        self.device_recovery_detection()
 
     def connect_device_post_delete(self):
         from openwisp_controller.config.models import Device
@@ -23,6 +28,25 @@ class DeviceMonitoringConfig(AppConfig):
             dispatch_uid='device_post_delete_receiver',
         )
 
+    def device_recovery_detection(self):
+        from openwisp_controller.config.models import Device
+        from .models import DeviceData, DeviceMonitoring
+
+        if app_settings.DEVICE_RECOVERY_DETECTION:
+            health_status_changed.connect(
+                self.manage_device_recovery_cache_key, sender=DeviceMonitoring
+            )
+            checksum_requested.connect(
+                self.trigger_device_recovery_checks,
+                sender=Device,
+                dispatch_uid='checksum_requested',
+            )
+            device_metrics_received.connect(
+                self.trigger_device_recovery_checks,
+                sender=DeviceData,
+                dispatch_uid='received_device_metrics',
+            )
+
     @classmethod
     def device_post_delete_receiver(cls, instance, **kwargs):
         from ..device.models import DeviceData
@@ -30,3 +54,22 @@ class DeviceMonitoringConfig(AppConfig):
         instance.__class__ = DeviceData
         instance.checks.all().delete()
         instance.metrics.all().delete()
+
+    @classmethod
+    def manage_device_recovery_cache_key(cls, instance, status, **kwargs):
+        """
+        It sets the ``cache_key`` as 1 when device ``health_status`` goes to ``critical``
+        and deletes the ``cache_key`` when device recovers from ``critical`` state
+        """
+        cache_key = get_device_recovery_cache_key(device=instance.device)
+        if status == 'critical':
+            cache.set(cache_key, 1, timeout=None)
+        else:
+            cache.delete(cache_key)
+
+    @classmethod
+    def trigger_device_recovery_checks(cls, instance, **kwargs):
+        from .tasks import trigger_device_recovery
+
+        if cache.get(get_device_recovery_cache_key(device=instance)):
+            trigger_device_recovery.delay(pk=instance.pk)
--- a/openwisp_monitoring/device/tests/test_api.py
+++ b/openwisp_monitoring/device/tests/test_api.py
@@ -1,10 +1,12 @@
-import json
 from unittest.mock import patch
 
 from django.contrib.auth import get_user_model
 from django.urls import reverse
 
+from openwisp_utils.tests import catch_signal
+
 from ... import settings as monitoring_settings
+from ...device.signals import device_metrics_received
 from ...monitoring.models import Graph, Metric
 from ..models import DeviceData
 from . import DeviceMonitoringTestCase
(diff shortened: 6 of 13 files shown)


nepython:
- [x] Listen to the health status change signal and set a cache key
- [x] Emit a new signal (device_metrics_received) when device metrics are received
- [x] React to the signals `config_download_requested`, `checksum_requested`, `device_metrics_received`
- [x] Add a setting (I have named it `OPENWISP_MONITORING_SPONTANEOUS_RECOVERY_DETECTION` as it seemed appropriate considering the goal of the issue)
- [x] Mentioned it in Docs
- [ ] If possible, give better name to tests and other functions :sweat_smile:

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,38 @@
...
+        checksum_requested.connect(
+            self.check_cache_key,
+            sender=Device,
+            dispatch_uid='checksum_requested'
+        )
+        config_download_requested.connect(
+            self.check_cache_key,
+            sender=Device,
+            dispatch_uid='config_download_requested'
+        )
```
nemesisdesign, line 26: changed my mind, let's avoid `config_download_requested`, `checksum_requested` is enough (it always precedes a config download)

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,38 @@
...
+from django_netjsonconfig.signals import checksum_requested, config_download_requested
 
 
 class CheckConfig(AppConfig):
     name = 'openwisp_monitoring.check'
     label = 'check'
     verbose_name = _('Network Monitoring Checks')
+
+    def ready(self):
+        from openwisp_controller.config.models import Device
```
nemesisdesign, line 14: since this code is related to a Device, I'd move this to `openwisp_monitoring.device`

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,38 @@
...
+        device_metrics_received.connect(
+            self.check_cache_key,
+            sender=DeviceData,
+            dispatch_uid='received_device_metrics'
+        )
+
+    @classmethod
+    def check_cache_key(cls, instance, **kwargs):
+        from .tasks import find_related_checks
+        if getattr(settings, 'OPENWISP_MONITORING_SPONTANEOUS_RECOVERY_DETECTION', True):
```
nemesisdesign, line 36: I'd use `OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION`, it makes more and more sense to move this to the device subapp.

openwisp_monitoring/device/api/views.py
```diff
@@ -75,6 +76,7 @@ def get(self, request, pk):
             response = HttpResponse(self._get_csv(data), content_type='text/csv')
             response['Content-Disposition'] = 'attachment; filename=data.csv'
             return response
+        device_metrics_received.send(sender=self.model, instance=self.instance)
```
nemesisdesign, line 79: for consistency, include the request object as a parameter.

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,38 @@
...
 class CheckConfig(AppConfig):
     name = 'openwisp_monitoring.check'
     label = 'check'
     verbose_name = _('Network Monitoring Checks')
+
+    def ready(self):
+        from openwisp_controller.config.models import Device
+        from ..device.models import DeviceData
+        from ..device.signals import device_metrics_received
+        checksum_requested.connect(
```
nemesisdesign, line 17: if the feature is disabled via setting, you should not connect to these signals.

openwisp_monitoring/check/tasks.py
```diff
@@ -54,3 +53,21 @@ def auto_create_ping(model, app_label, object_id, created):
...
+
+
+@shared_task
+def find_related_checks(object_id):
+    """
+    Retrieves all related checks to the passed ``device``
+    and calls the ``perform_check`` task from each of them.
+    If no check exists changes the status to ``OK``.
+    """
+    iterator = Check.objects.filter(active=True,
```
nemesisdesign, line 65: to get the checks related to a device, we may use the reverse generic relation I mentioned above.

.travis.yml
```diff
@@ -43,6 +43,8 @@ install:
   - pip install -e .
 
 script:
+  # Until release of django-netjsonconfig 12.0
```
nemesisdesign, line 46: `0.12.0`

openwisp_monitoring/check/tests/test_ping.py
```diff
@@ -207,3 +209,37 @@ def test_auto_graph_disabled(self, *args):
         self.assertEqual(Graph.objects.count(), 0)
         check.perform_check()
         self.assertEqual(Graph.objects.count(), 0)
+
```
nemesisdesign, line 212: remember: In `test_ping.py` we have to put only tests strictly related to ping checks

openwisp_monitoring/check/apps.py
nepython: Humble request please update the issue too :sweat_smile:

nemesisdesign: Done

nepython:
@nemesisdesign, I have implemented all other suggestions. `Reverse generic relation` might take a while.

openwisp_monitoring/check/tasks.py
```diff
@@ -54,3 +53,21 @@ def auto_create_ping(model, app_label, object_id, created):
...
+                                    content_type__model='device',
+                                    object_id=object_id) \
+                            .only('id') \
+                            .values('id')
+    if not list(iterator):
+        DeviceMonitoring.objects.get(device_id=object_id).update_status('ok')
+    for check in iterator:
+        perform_check.delay(check['id'])
```
nemesisdesign, line 73: This could have be rewritten as:

```python
device = DeviceData.objects.get(pk=pk)
checks = device.checks.filter(active=True).only('id').values('id')
for check in checks:
    perform_check.delay(check['id'])
# if there's no available check for this device, we'll flag it as OK directly
else:
    device.monitoring.update_status('ok')
```

Anyway, I don't think this is the right place to put this code.
I would rename this as `trigger_device_recovery` and move it to `openwisp_monitoring.device` since most of the changes are being done there.

openwisp_monitoring/device/apps.py
```diff
@@ -10,4 +14,33 @@ class DeviceMonitoringConfig(AppConfig):
...
+            )
+
+    @classmethod
+    def set_cache_key(cls, instance, status, **kwargs):
+        if getattr(settings, 'OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION', True):
+            is_critical = 1 if status == 'critical' else 0
+            cache.set(f'device-{instance.device.pk}-react-to-updates', is_critical, 250)
+
+    @classmethod
+    def check_cache_key(cls, instance, **kwargs):
```
nemesisdesign, line 43: the aim of this method is not to check the cache key, but to trigger the recovery of the device.

Hence, you can rename this method `trigger_device_recovery_checks`

openwisp_monitoring/device/apps.py
```diff
@@ -10,4 +14,33 @@ class DeviceMonitoringConfig(AppConfig):
     verbose_name = _('Device Monitoring')
 
     def ready(self):
+        from openwisp_controller.config.models import Device
+        from .models import DeviceData, DeviceMonitoring
         manage_short_retention_policy()
+        health_status_changed.connect(
```
nemesisdesign, line 20: this should also be done only if device recovery detection is enabled

openwisp_monitoring/device/apps.py
```diff
@@ -10,4 +14,33 @@ class DeviceMonitoringConfig(AppConfig):
...
+                dispatch_uid='checksum_requested'
+            )
+            device_metrics_received.connect(
+                self.check_cache_key,
+                sender=DeviceData,
+                dispatch_uid='received_device_metrics'
+            )
+
+    @classmethod
+    def set_cache_key(cls, instance, status, **kwargs):
```
nemesisdesign, line 37: I'd use `manage_device_recovery_cache_key` and add a docstring to this method explains why it sets or remove this cache key

openwisp_monitoring/device/apps.py
```diff
@@ -10,4 +14,33 @@ class DeviceMonitoringConfig(AppConfig):
...
+        if getattr(settings, 'OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION', True):
+            is_critical = 1 if status == 'critical' else 0
+            cache.set(f'device-{instance.device.pk}-react-to-updates', is_critical, 250)
+
+    @classmethod
+    def check_cache_key(cls, instance, **kwargs):
+        from ..check.tasks import find_related_checks
+        if cache.get(f'device-{instance.pk}-react-to-updates'):
```
nemesisdesign, line 45: please create a helper function to generate the cache key because it's being repeated twice and we have to avoid repetition

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -433,6 +437,31 @@ def test_get_device_metrics_400_bad_timezone(self):
...
+        request = response.renderer_context['request']
+        handler.assert_called_once_with(
+            instance=dd,
+            request=request,
+            sender=DeviceData,
+            signal=device_metrics_received,
+        )
+
+    # Tests device status is set to ok if no related checks present
+    def test_status_set_ok(self):
```
nemesisdesign, line 457: why is this test here? `test_api` should only test the API.
I'm also not 100% sure what's the aim of this test.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -216,6 +218,36 @@ def test_status_changed(self):
             signal=health_status_changed,
         )
 
+    def test_cache_key_not_set(self):
+        dm, ping, load, process_count = self._create_env()
+        setattr(settings,
+                'OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION',
+                False)
+        query = f'device-{dm.device.pk}-react-to-updates'
```
nemesisdesign, line 226: use the helper method I asked also here.
Also, is not a query but a cache key.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -216,6 +218,36 @@ def test_status_changed(self):
             signal=health_status_changed,
         )
 
+    def test_cache_key_not_set(self):
+        dm, ping, load, process_count = self._create_env()
+        setattr(settings,
```
nemesisdesign, line 223: decorate the method with:

```python
@mock.patch('openwisp_monitoring.device.settings.DEVICE_RECOVERY_SETTINGS', False)
```

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -216,6 +218,36 @@ def test_status_changed(self):
...
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'critical')
+        self.assertEqual(cache.get(query), None)
+        setattr(settings,
+                'OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION',
+                True)
+
+    def test_cache_key_set(self):
+        dm, ping, load, process_count = self._create_env()
+        query = f'device-{dm.device.pk}-react-to-updates'
```
nemesisdesign, line 237: same here

openwisp_monitoring/device/tests/test_api.py
nepython: It was stated in the issue
> If there's no related check, move the status to OK.

The above test was included to verify the same but yes, I have messed up the location of almost all functions which might make it difficult to relate the same

nemesisdesign: ok, no problem, now I see more clearly, last night I was pretty tired from an entire day of review and I got confused regarding why this test was here. You may want to move all this functionality to a new test suite called `test_recovery.TestRecovery`.

nepython: Done

README.rst
```diff
@@ -282,6 +282,36 @@ you can use the following configuration:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Path**: ``openwisp_monitoring.device.signals.device_metrics_received``
+
+**Arguments**:
+
+- ``instance``: instance of ``Device`` whose metrics have been received
+- ``request``: the HTTP request object
+
+This signal is emitted when device metrics are received via the ``DeviceMetric`` views.
```
nemesisdesign, line 310: views > view

openwisp_monitoring/device/apps.py
```diff
@@ -30,3 +51,26 @@ def device_post_delete_receiver(cls, instance, **kwargs):
         instance.__class__ = DeviceData
         instance.checks.all().delete()
         instance.metrics.all().delete()
+
+    @classmethod
+    def _generate_cache_key(cls, device):
```
nemesisdesign, line 56: you may use a simple function and store it here: https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/utils.py

I think it could be better since you'll be using that in tests as well. It makes sense as it's not something specifically related to the AppConfig class.

I think it can be named simply `get_device_recovery_cache_key'

openwisp_monitoring/device/apps.py
```diff
@@ -30,3 +51,26 @@ def device_post_delete_receiver(cls, instance, **kwargs):
...
+        return f'device-{device.pk}-react-to-updates'
+
+    @classmethod
+    def manage_device_recovery_cache_key(cls, instance, status, **kwargs):
+        """
+        WIP
+        """
+        cache_key = cls._generate_cache_key(device=instance.device)
+        if status == 'critical':
+            if app_settings.DEVICE_RECOVERY_DETECTION:
```
nemesisdesign, line 66: maybe this is not necessary here?

nepython: The line is not required in the module as such but was being required while unit testing. The reason being the setting `DEVICE_RECOVERY_DETECTION` is `True` by default and hence the function `manage_device_recovery_cache_key` gets connected to `health_status_changed` signal when the app is loaded right before unit tests start. I was thinking of some better way to avoid the above and finally did it as below:
1. Disconnect the signal in the test.
2. Reload the app to verify that function doesn't get connected to signal when setting is patched as `False`.
3. After verifying that cache key has not been set, reload the app with setting as `True` (Default),

You may check `test_device_recovery_cache_key_not_set` in `test_recovery.TestRecovery` for the code. It might seem a bit lengthy but checks the condition in a better manner than above workaround :smiley:

openwisp_monitoring/device/api/views.py
```diff
@@ -75,6 +76,9 @@ def get(self, request, pk):
             response = HttpResponse(self._get_csv(data), content_type='text/csv')
             response['Content-Disposition'] = 'attachment; filename=data.csv'
             return response
+        device_metrics_received.send(
```
nemesisdesign, line 79: The metrics are not received in the GET, but in the POST: GET is used by the web UI to retrieve the data for the graphs, POST is used by the devices to send their metrics.
I didn't catch this earlier.

Please move the signal to the `post` method and update the tests accordingly.

nepython:
@nemesisdesign, I have moved few helper function used in `test_api.TestDeviceApi` to `__init.py__` as they were needed in `test_recovery.TestRecovery`. I didn't wanted to duplicate it and am not very sure what's the standard practice in such a scenario. Also, one of the travis build is failing owing to #7 :sweat_smile:

PR #23, opened 2020-04-13
[monitoring] Wait and retry if influxdb is not ready #21

### LINKED ISSUE #21: [monitoring] Wait and retry if influxdb is not ready
If influxdb is starting up, this may fail: https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/apps.py#L15

We should catch that failure, log a warning, wait something like 3 seconds and try again and repeat this procedure 5 times before giving up.

Once we implement this, we can remove this: https://github.com/openwisp/openwisp-monitoring/blob/master/.travis.yml#L36

nepython:
Closes #21

### DIFF
--- a/openwisp_monitoring/monitoring/apps.py
+++ b/openwisp_monitoring/monitoring/apps.py
@@ -1,6 +1,9 @@
+from time import sleep
+
 from django.apps import AppConfig
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
+from requests.exceptions import ConnectionError
 
 from .utils import create_database
 
@@ -11,6 +14,23 @@ class MonitoringConfig(AppConfig):
     verbose_name = _('Network Monitoring')
 
     def ready(self):
-        # create influxdb database if doesn't exist yet
-        create_database()
+        self.create_database()
         setattr(settings, 'OPENWISP_ADMIN_SHOW_USERLINKS_BLOCK', True)
+
+    def create_database(self):
+        # create influxdb database if doesn't exist yet
+        for attempt_number in range(5):
+            try:
+                create_database()
+                break
+            except ConnectionError as e:
+                self.warn_and_delay(attempt_number + 1)
+                if attempt_number == 4:
+                    raise e
+
+    def warn_and_delay(self, attempt_number):
+        print(
+            'Got error while connecting to timeseries DB. '
+            f'Retrying again in 3 seconds. Attempt {attempt_number} out of 5'
+        )
+        sleep(3)
--- a/openwisp_monitoring/monitoring/tests/test_db_creation.py
+++ b/openwisp_monitoring/monitoring/tests/test_db_creation.py
@@ -1,0 +1,38 @@
+import io
+from contextlib import redirect_stdout
+from unittest.mock import patch
+
+from django.apps import apps
+from django.test import TestCase
+from requests.exceptions import ConnectionError
+
+from ..utils import get_db
+
+
+def mock_create_database(self, db):
+    raise ConnectionError
+
+
+class TestDatabase(TestCase):
+    @patch('openwisp_monitoring.monitoring.settings.INFLUXDB_DATABASE', 'test_db')
+    @patch('openwisp_monitoring.monitoring.apps.MonitoringConfig.warn_and_delay')
+    @patch('influxdb.client.InfluxDBClient.create_database', mock_create_database)
+    def test_check_retry(self, mock):
+        try:
+            apps.get_app_config('monitoring').create_database()
+        except ConnectionError:
+            pass
+        get_db().drop_database('test_db')
+        self.assertEqual(mock.call_count, 5)
+
+    @patch('openwisp_monitoring.monitoring.apps.sleep', return_value=None)
+    def test_warn_and_delay(self, mock):
+        f = io.StringIO()
+        with redirect_stdout(f):
+            apps.get_app_config('monitoring').warn_and_delay(0)
+        self.assertEqual(
+            f.getvalue(),
+            'Got error while connecting to timeseries DB. '
+            'Retrying again in 3 seconds. Attempt 0 out of 5\n',
+        )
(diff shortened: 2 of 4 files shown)


openwisp_monitoring/monitoring/utils.py
```diff
@@ -58,5 +60,16 @@ def create_database():
     databases = [database['name'] for database in items]
     # if database does not exists, create it
     if settings.INFLUXDB_DATABASE not in databases:
-        db.create_database(settings.INFLUXDB_DATABASE)
-        logger.info('Created inlfuxdb database {0}'.format(settings.INFLUXDB_DATABASE))
+        for n in range(5):
```
nemesisdesign, line 63: don't do this here. It should not be the job of this function to retry, this is a lower level function.
Do it in the ready() method of the app, that's where this function will be called.

openwisp_monitoring/monitoring/utils.py
```diff
@@ -58,5 +60,16 @@ def create_database():
...
+                db.create_database(settings.INFLUXDB_DATABASE)
+                break
+            except InfluxDBClientError:
+                warn_and_delay(n)
+        logger.info(f'Created influxdb database {settings.INFLUXDB_DATABASE}'
+
+
+def warn_and_delay(n):
+    print(f'Caught an error, retrying ({n}/5 tries)')
+    print(f'Database: {settings.INFLUXDB_DATABASE} not created')
```
nemesisdesign, line 74: remove this line

openwisp_monitoring/monitoring/utils.py
```diff
@@ -58,5 +60,16 @@ def create_database():
...
+            try:
+                db.create_database(settings.INFLUXDB_DATABASE)
+                break
+            except InfluxDBClientError:
+                warn_and_delay(n)
+        logger.info(f'Created influxdb database {settings.INFLUXDB_DATABASE}'
+
+
+def warn_and_delay(n):
+    print(f'Caught an error, retrying ({n}/5 tries)')
```
nemesisdesign, line 73: Use `Got error while connecting to timeseries DB. Retrying again in 3 seconds. Attempt {n} out of 5.`

openwisp_monitoring/monitoring/tests/test_db_creation.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TestDatabase(TestCase):
+    ORIGINAL_DB = app_settings.INFLUXDB_DATABASE
+    TEST_DB = 'test_db'
+
+    @patch('openwisp_monitoring.monitoring.utils.warn_and_delay')
+    @patch('influxdb.client.InfluxDBClient.create_database', mock_create_database)
+    def test_check_retry(self, mock):
+        setattr(app_settings, 'INFLUXDB_DATABASE', self.TEST_DB)
```
nemesisdesign, line 21: can you try mocking this as well?

`@patch('openwisp_monitoring.monitoring.app_settings.INFLUXDB_DATABASE', 'test_db')`

nepython:
@nemesisdesign done

openwisp_monitoring/monitoring/apps.py
```diff
@@ -12,5 +15,15 @@ class MonitoringConfig(AppConfig):
...
+            try:
+                create_database()
+                break
+            except InfluxDBClientError:
+                self.warn_and_delay(n)
         setattr(settings, 'OPENWISP_ADMIN_SHOW_USERLINKS_BLOCK', True)
+
+    def warn_and_delay(self, n):
+        print('Got error while connecting to timeseries DB. '
+              f'Retrying again in 3 seconds. Attempt {n} out of 5')
```
nemesisdesign, line 28:
```python
>>> list(range(5))
[0, 1, 2, 3, 4]
```


At the first attempt this will be returned as output:

`Retrying again in 3 seconds. Attempt 0 out of 5`

Add 1 to n to solve this. Please also rename `n` to `attempt_number`

openwisp_monitoring/monitoring/apps.py
```diff
@@ -12,5 +15,15 @@ class MonitoringConfig(AppConfig):
 
     def ready(self):
         # create influxdb database if doesn't exist yet
-        create_database()
+        for n in range(5):
```
nemesisdesign, line 18: Please rename `n` to `attempt_number`

openwisp_monitoring/monitoring/apps.py
```diff
@@ -12,5 +15,15 @@ class MonitoringConfig(AppConfig):
...
+            try:
+                create_database()
+                break
+            except InfluxDBClientError:
+                self.warn_and_delay(n)
```
nemesisdesign, line 23: Let's move this block of code to a dedicated method and then call it with `self.create_database()`.
This maintains consistency with other openwisp modules.

nepython:
@nemesisdesign, I can't reproduce the error. I had used `InfluxDBClientError` as `InfluxDB` docs mentioned it to be the default exception being raised in case database is not created but there's some misunderstanding on my part, I assume. I think the exception raised might be similar to this [travis-build](https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/672183803#L507) (Correct me in case I am wrong). If that's the case we can implement a script similar to this.
```shell
#!/bin/bash
docker run -d --name influxdb -p 8086:8086 influxdb
command=docker exec -it influxdb influx -execute "CREATE DATABASE openwisp2"
log="Got error while connecting to timeseries DB. Retrying again in 3 seconds."
for attempt_number in 1 2 3 4 5
do
  if ($command); then
    break
  fi
  sleep 3 && echo log && echo "Attempt $attempt_number out of 5"
done
```
and to hide the stderr
```c
>>> ./start-influxdb 2> /dev/null
```
If there is some other exception, can you please share how to reproduce it / `Exception`

nepython:
PR is ready for review :smile:

PR #40, opened 2020-04-17
[notifications] Added links for related fields #5

### LINKED ISSUE #5: [notifications] Link related object
Find a way to link the related object.

TheOneAboveAllTitan:
Closes #5

### DIFF
--- a/openwisp_monitoring/notifications/admin.py
+++ b/openwisp_monitoring/notifications/admin.py
@@ -1,4 +1,6 @@
 from django.contrib import admin
+from django.urls import NoReverseMatch, reverse
+from django.utils.html import format_html
 from django.utils.translation import ugettext_lazy as _
 
 from openwisp_users.admin import UserAdmin
@@ -10,24 +12,41 @@ from .models import Notification, NotificationUser
 @admin.register(Notification)
 class NotificationAdmin(admin.ModelAdmin):
     raw_id_fields = ('recipient',)
+    readonly_fields = [
+        'action_object_link',
+        'actor_object_link',
+        'target_object_link',
+        'related_object',
+    ]
     list_display = ('description', 'read', 'level', 'timesince')
     list_filter = (
         'level',
         'unread',
     )
     fieldsets = (
-        (None, {'fields': ('timestamp', 'level', 'description', 'emailed',)}),
+        (
+            None,
+            {
+                'fields': (
+                    'timestamp',
+                    'level',
+                    'description',
+                    'related_object',
+                    'emailed',
+                )
+            },
+        ),
         (
             _('Advanced options'),
             {
                 'classes': ('collapse',),
                 'fields': (
                     'actor_content_type',
-                    'actor_object_id',
+                    'actor_object_link',
                     'action_object_content_type',
-                    'action_object_object_id',
+                    'action_object_link',
                     'target_content_type',
-                    'target_object_id',
+                    'target_object_link',
                     'data',
                 ),
             },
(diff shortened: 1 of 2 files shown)


openwisp_monitoring/notifications/admin.py
```diff
@@ -23,11 +24,11 @@ class NotificationAdmin(admin.ModelAdmin):
                 'classes': ('collapse',),
                 'fields': (
                     'actor_content_type',
-                    'actor_object_id',
+                    'actor_object_',
```
nemesisdesign, line 27: instead of the ending underscore you may use `_link`, eg: `actor_object_link`.

TheOneAboveAllTitan: Using ending underscore helped me to render it as `Actor Object` without the need of adding any other code. Nevertheless I am shifting the code logic to `admin.py` and will implement it in `base.BaseNotificationAdmin` of Notification module, since it is more of an admin related feature.

openwisp_monitoring/notifications/admin.py
```diff
@@ -57,11 +77,87 @@ def mark_as_read(self, request, queryset):
...
+            return '-'
+
+    target_object_link.short_description = _('Target Object')
+
+    def related_object(self, obj):
+        target_object = self.target_object_link(obj)
+        if isinstance(target_object, SafeText):
+            url = self.target_object_link(obj).split('"')[1]
+            return format_html(
+                '<a href="{url}" id="related-object-url">{content_type} | {name}</a>',
```
nemesisdesign, line 146: This is the anchor text I see:

`config | device | Cap-Dev-01`

But this would be more user friendly:

`device: Cap-Dev-01`

You should be able to get the device model name using `content_type.model` or something similar.

openwisp_monitoring/notifications/admin.py
```diff
@@ -57,11 +77,87 @@ def mark_as_read(self, request, queryset):
...
+            return obj.target_object_id
+        except AttributeError:
+            return '-'
+
+    target_object_link.short_description = _('Target Object')
+
+    def related_object(self, obj):
+        target_object = self.target_object_link(obj)
+        if isinstance(target_object, SafeText):
+            url = self.target_object_link(obj).split('"')[1]
```
nemesisdesign, line 144: it works, but I suggest you to not do these kind of things, because it does not cost anything to add a flag in this method which returns the URL directly, without further splitting. These kind of shortcuts can produce problems when future maintainers may change this part of the code. Restrain yourself from doing this. I'll leave a suggestion in the next comment (see the `html` argument of the funciton I will suggest)

openwisp_monitoring/notifications/admin.py
```diff
@@ -57,11 +77,87 @@ def mark_as_read(self, request, queryset):
...
+            )
+            return format_html(
+                '<a href="{url}" id="target-object-url">{id}</a>',
+                url=url,
+                id=obj.target_object_id,
+            )
+        except NoReverseMatch:
+            return obj.target_object_id
+        except AttributeError:
+            return '-'
```
nemesisdesign, line 137: the previous 3 methods have identical structure so it's good to generalize its code and avoid repetition.

Try starting from here (I did not test this code):

```python
def _get_link(self, obj, field, html=True):
    content_type = getattr(obj, f'{field}_content_type')
    object_id = getattr(obj, f'{field}_object_id')
    try:
        url = reverse(
            f'admin:{content_type.app_label}_{content_type.model}_change'
            args=[object_id],
        )
        if not html:
            return url
        return format_html(
            f'<a href="{url}" id="{field}-object-url">{object_id}</a>'
        )
    except NoReverseMatch:
        return object_id
    except AttributeError:
        return '-'
```

:eyes:  Notice the html argument which will allow you to get the URL without the HTML so you don't need to split in `related_object`.

You can then call this method in all the other methods which produce an HTML link. 

Get used to avoid repetition (DRY: Do not repeat yourself, it's a concept also expressed in the book I should have suggested to you).

openwisp_monitoring/notifications/admin.py
TheOneAboveAllTitan: I will keep this in mind.

openwisp_monitoring/notifications/admin.py
TheOneAboveAllTitan: Thanks for suggestion. I see how much it improves quality of code. I try to follow DRY and other principles in future. Should I apply similar logic in tests as well?

nemesisdesign: Regarding tests, it can be also useful but it's less black and white and more a matter of judgement. In general, if I have the same test logic repeated in more than 2 tests, I may try to generalize it in a test method and reuse it to test more cases faster. I just have done that here: https://github.com/openwisp/openwisp-monitoring/pull/48/files#diff-0d2e8c67f37123458a0ec4cbbcc3b9d6L97-R137

nemesisdesign:
Great, thanks :+1:

PR #42, opened 2020-04-21
[monitoring] Ping when Device IP is not present #24

### LINKED ISSUE #24: [monitoring] Ping when Device IP is not present
Due to this case https://github.com/openwisp/openwisp-controller/pull/188#issuecomment-613076203, we need to make the Ping class able to discern that.

If `_get_ip` returns None, we should:

- check whether the health status of the device is `ok` or (`problem` AND if ping is listed in `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS`)
- if none of the previous cases is true, keep behaviour as is now (just return)
- otherwise call:

```python
self.store_result({
    'reachable': 0,
    'loss': 100,
})
return
```

Which will cause the health stauts to become critical.

NoumbissiValere:
Fixed the issue as describe in its comment
and also added test cases to test this
enhancement

Fixes #24

### DIFF
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -8,6 +8,8 @@ from jsonschema.exceptions import ValidationError as SchemaError
 from openwisp_controller.config.models import Device
 
 from ... import settings as monitoring_settings
+from ...device import settings as device_settings
+from ...device.models import DeviceMonitoring
 from ...monitoring.models import Graph, Metric, Threshold
 from ..exceptions import OperationalError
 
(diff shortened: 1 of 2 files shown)


NoumbissiValere:
@nemesisdesign Please this is ready for review

openwisp_monitoring/check/classes/ping.py
```diff
@@ -77,6 +79,14 @@ def check(self, store=True):
         timeout = self._get_param('timeout')
         ip = self._get_ip()
         if not ip:
+            device_monitored = DeviceMonitoring.objects.get(
```
nemesisdesign, line 82: I think this could be simplified to:

```python
self.related_object.monitoring
```

openwisp_monitoring/check/classes/ping.py
```diff
@@ -77,6 +79,14 @@ def check(self, store=True):
         timeout = self._get_param('timeout')
         ip = self._get_ip()
         if not ip:
+            device_monitored = DeviceMonitoring.objects.get(
+                device__pk=self.related_object.pk
+            )
+            if (
+                device_monitored.status in ['ok', 'problem']
+                and device_settings.CRITICAL_DEVICE_METRICS[0]['key'] == 'ping'
```
nemesisdesign, line 87: if there are multiple metrics listed in this setting and ping is not the first one, this will fail.

The Metric model has a method to find out if a metric is critical, use that one next time.

PR #48, opened 2020-04-23
[check] Handled case in which IP is not present. 

nemesisdesign:
Continuation of #42. Related to #24.

### DIFF
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -8,6 +8,8 @@ from jsonschema.exceptions import ValidationError as SchemaError
 from openwisp_controller.config.models import Device
 
 from ... import settings as monitoring_settings
+from ...device import settings as device_settings
+from ...device.models import DeviceMonitoring
 from ...monitoring.models import Graph, Metric, Threshold
 from ..exceptions import OperationalError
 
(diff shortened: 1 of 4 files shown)


openwisp_monitoring/check/classes/ping.py
```diff
@@ -78,16 +77,17 @@ def check(self, store=True):
...
-                self.store_result({'reachable': 0, 'loss': 100})
-            return
+            monitoring = self.related_object.monitoring
+            # device not known yet, ignore
+            if monitoring.status == 'unknown':
+                return
+            # device is known, simulate down
+            result = {'reachable': 0, 'loss': 100}
+            if store:
+                self.store_result(result)
```
NoumbissiValere, line 89: @nemesisdesign what  if `ping` is not the metric we are using and `store=True`? this will set the device as critical whereas this should have not been the case

nemesisdesign: Not sure what you mean with exactly "what if ping is not the metric we are using", because if the ping class is being used, we're surely dealing with the ping metric. Could you clarify please?

Initially I asked to check some preconditions before performing the check, but then while testing I understood that we should avoid the check only if the device is in unknown status, because if the device is known by the system, users will expect to see the data in the graphs to be present but if we avoid performing the check at all the graph will be empty and they'll be confused.

Let me know if this is explanation is clear and if not don't hesitate to ask further questions and if you still doubt this is correct let me know as I also do mistakes (may as well find more problems in the coming weeks, hopefully not :blush:).

NoumbissiValere: Yeah that's true. it makes sense to reasoned that since we are dealing with the `Ping` class, the check should obviously be a `ping` and if i get things right, `ping` is the only check we support right now. thanks for the clarification. code looks great :smile:

nemesisdesign: yes it's the only check for now. I hope we'll implement SNMP soon which will allow us to monitor also other type of devices, not only OpenWRT.

PR #49, opened 2020-04-23
[notifications] Swapped in openwsip_notifications

### LINKED ISSUE #57: [monitoring][notifications] Decouple notification module
- [x] Extract code for notification module and create a separate module for it. 
- [ ] Adapt monitoring module to use `openwisp-notifications`

pandafy:
Swapped out openwisp_monitoring.notifications with openwisp_notifications
Added data migration for backward compatibility

~Depends on https://github.com/openwisp/openwisp-notifications/pull/7~
Depends on on https://github.com/openwisp/openwisp-notifications/pull/16 
Fixes #57


requirements.txt
```diff
@@ -5,3 +5,5 @@ django-apptemplates>=1.4,<2.0
 django-celery-email>=3.0.0,<3.1
 djangorestframework>=3.11,<3.12
 swapper
+#TODO : Remove after release of openwisp-notifications
+https://github.com/openwisp/openwisp-notifications/tarball/master
```
nemesisdesign, line 9: as far as I know this doesn't work with setuptools, usually here we put just the version `openwisp-notification~=0.1.0`, and we install from github in the .travis.yml config

openwisp_monitoring/notifications/migrations/0005_auto_20200421_2201.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 3.0.4 on 2020-04-21 20:01
```
nemesisdesign, line 1: watch out for the name of the file, we don't allow the default auto-generated names, openwisp-qa-check will complain

pandafy:
I have squashed all commits and rebased on current master. 

I tested the data migration script as follows:
- Using git checked out the commit before these changes, ran migrations, and created few notifications from the Django shell. 
- Then checked out this commit, ran migrations, and queried all notifications from Django shell using models of `openwisp-notifications`

This PR now depends on https://github.com/openwisp/openwisp-notifications/pull/16

Issue #47, opened 2020-04-21
[enhancement] Refactor the query parameter of Chart


nepython:
This is an **idea/draft** and needs more refining, few details also need to be added.

**The Need**
The query parameter of a chart (now called Graph) is editable via the django admin. This poses several issues: from UX (it's hard to deal with it) to security (the query could be manipulated to get data to which the user does not have access to).

This needs to be refactored so that the user can choose among a predefined set of queries from a list.

The list should be customizable using a django setting, so that users can implement their own queries if needed. Add documentation about this setting explaining how to customize it.

**Details missing** :
What all queries should be added in the list by default?
What would be the best location to define the base list from in the `monitoring` app. Currently, I am thinking of `utils.py`

nemesisdesign:
> Details missing :
> What all queries should be added in the list by default?
> What would be the best location to define the base list from in the monitoring app. Currently, I am thinking of utils.py

All the graphs currently supported: bandiwdth, wifi clients, uptime, packet loss, rtt, histograms.

I will work on improving the definition of this issue during this week. There's a chance I may be doing part of it to help you understand, and also to help to clean the mess I have created in the first place :blush:

PR #51, opened 2020-04-26
[enhancement] Refactor the query parameter of Chart #47

nepython:
This is a draft and not yet ready. Adding `help_text`, more `query` in the list and better documentation remains. I am opening this PR for discussion (was not sure if this is how it was supposed to be done :sweat_smile: ) and would iterate accordingly.

**PS**: I realized now that the example that I have provided in the README is one of the forbidden queries so will modify it.


nemesisdesign:
See #63.

PR #61, opened 2020-04-30
[monitoring] Correct uptime and local_time #32

### LINKED ISSUE #32: [monitoring] Correct uptime and local_time
To be done after https://github.com/openwisp/openwisp-monitoring/issues/31.

`uptime` and `local_time` in `DeviceData.data` show the values as of the time when the measurement of the data was last stored.

So when viewing what time the device has, if the measurement was taken 4 minutes ago, we'll see the time of 4 minutes ago, but this is confusing for the user. 

If the health status of the device is not critical, we should auto-correct those 2 values to take in consideration the time passed since the last measurement was stored.

It would also be handy to have the information of when the measurement was taken available in the `.data` and `.data_human_friendly`.

NoumbissiValere:
Fixed the issue by updating data_user_friendly
to keep record of uptime_seconds. This was
used in the newly added js file to update the
uptime every minute while waitinf for the
crontab to finish its job. Also, the local_time
was update using the js native Date function
every minute while waiting for the crontab to
finish its job

Fixes #32


NoumbissiValere:
@nemesisdesign I think this is ready for review

NoumbissiValere:
Ok. Thanks for the feedback. I will correct it

NoumbissiValere:
@nemesisdesign  I have updated the PR with some few changes as to what was demanded. after looking at the code and workings, i concluded that it will be good to leave the `.data` unaltered with `measurement timestamp` as it's not part of the data which was send the `influxdb`. thus ensuring its integrity, such that users get back only what they stored to the db. reason why i instead added an extra property which gets the `measurement timestamp`  and it's called by `data_user_friendly` to update the `local_time` and `uptime`.

Please if my assumption and conclusions are wrong, correct me so i could correct the PR. Thanks

openwisp_monitoring/device/models.py
```diff
@@ -78,22 +80,35 @@ def data_user_friendly(self):
...
         if self.__data:
             return self.__data
-        q = (
-            "SELECT data FROM {0}.{1} WHERE pk = '{2}' "
-            "ORDER BY time DESC LIMIT 1".format(SHORT_RP, self.__key, self.pk)
-        )
-        points = list(query(q).get_points())
+        points = self._get_db_points()
         if not points:
             return None
```
nemesisdesign, line 99: why don't you set `measured_at` here?

I'd also rename this attribute to 'data_timestamp'  (is it a timestamp right?).

That way we will avoid two queries to get the same information.

openwisp_monitoring/device/models.py
```diff
@@ -45,16 +45,18 @@ def __init__(self, *args, **kwargs):
     def data_user_friendly(self):
         if not self.data:
             return None
+        measured_at = datetime.strptime(self.measured_at[0:19], '%Y-%m-%dT%H:%M:%S')
```
nemesisdesign, line 48: what does `self.measured_at` contain? Why is it being sliced?

NoumbissiValere: @nemesisdesign `influxdb` stores time like this `2016-06-13T17:43:50.1004002Z` where the value after the dot are nanoseconds. so i am cutting off the nanosecond to focus just on the time.

NoumbissiValere:
@nemesisdesign please can review this when you have time?

PR #64, opened 2020-05-01
[monitoring] Allow to store IP addresses of interfaces in DeviceData.data #30

### LINKED ISSUE #30: [monitoring] Allow to store IP addresses of interfaces in DeviceData.data
The device status information (accessible via `DeviceData.data`) stores information about the current status of the device (uptime, memory usage, load averages).

When the device has interfaces (eg LAN or WAN) set in DHCP client mode, there's no way to easily know which address the device got from the DHCP server.

This aims to fix that.

Incoming payoad example:
```
{
    'type': 'DeviceMonitoring',
    'interfaces': [
        {
            'name': 'br-lan',
            'addresses': [
                {
                    'proto': 'static',
                    'family': 'ipv4',
                    'address': '192.168.1.1',
                    'mask': 24
                },
                {
                    'proto': 'dhcp',
                    'family': 'ipv4',
                    'address': '10.0.0.4',
                    'mask': 24
                }
            ]
        }
    }
}
```

The schema in `openwisp_monitoring.device.schema` will have to be updated to include the addresses schema: http://netjson.org/rfc.html#addresses1
The gateway option can be ignored.

If we get a non-wireless interface without any address, we will not store it for now (wireless interfaces are stored anyway, they contain other useful info related to the wireless settings).

Once the data is saved and validated correctly (a few tests will need to ensure bogus addresses will be rejected) we will have to ensure the ip addresses of each interface are shown in the device status tab:
https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/templates/admin/config/device/change_form.html#L66-L124

Will require https://github.com/openwisp/lua-monitoring/issues/1 to be complete.

nepython:
~~This is still WIP. The following need to be done:-~~
* [x] show the addresses in the device status tab: https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/templates/admin/config/device/change_form.html
* [x] add tests to ensure bogus addresses will be rejected
* [x] Creating a PR in `monitoring-lua` as mentioned in the issue

**PS**: The JSON type for `family` was provided as `integer` [here](http://netjson.org/rfc.html#addresses1) but it appeared visibly as `string` to me

Closes #30

### DIFF
--- a/openwisp_monitoring/device/models.py
+++ b/openwisp_monitoring/device/models.py
@@ -68,9 +68,10 @@ class DeviceData(Device):
                 remove.append(interface)
                 continue
             # human readable mode
-            interface['wireless']['mode'] = interface['wireless']['mode'].replace(
-                '_', ' '
-            )
+            if 'wireless' in interface and 'mode' in interface['wireless']:
+                interface['wireless']['mode'] = interface['wireless']['mode'].replace(
+                    '_', ' '
+                )
             # convert to GHz
             if 'wireless' in interface and 'frequency' in interface['wireless']:
                 interface['wireless']['frequency'] /= 1000
--- a/openwisp_monitoring/device/schema.py
+++ b/openwisp_monitoring/device/schema.py
@@ -145,6 +145,27 @@ schema = {
                             "tx_carrier_errors": {"type": "integer"},
                         },
                     },
+                    "addresses": {
+                        "type": "array",
+                        "title": "Addresses",
+                        "uniqueItems": True,
+                        "additionalItems": True,
+                        "items": {
+                            "additionalProperties": True,
+                            "title": "Address",
+                            "type": "object",
+                            "required": ["proto", "family", "address", "mask"],
+                            "properties": {
+                                "proto": {"type": "string"},
+                                "family": {"type": "string"},
+                                "address": {
+                                    "type": "string",
+                                    "anyOf": [{"format": "ipv4"}, {"format": "ipv6"}],
+                                },
+                                "mask": {"type": "integer"},
+                            },
+                        },
+                    },
                 },
             },
         },
--- a/openwisp_monitoring/device/static/monitoring/css/monitoring.css
+++ b/openwisp_monitoring/device/static/monitoring/css/monitoring.css
@@ -158,3 +158,11 @@ td.field-health_status,
 #ow-graph-container b{
     background-color: #000 !important;
 }
+
+table.device-addresses {
+  margin-top: 20px;
+}
+
+table.device-addresses td.dhcp {
+  background-color: lightgray;
+}
--- a/openwisp_monitoring/device/templates/admin/config/device/change_form.html
+++ b/openwisp_monitoring/device/templates/admin/config/device/change_form.html
@@ -141,6 +141,28 @@
                   {% endif %}
                   {% endif %}
               </div>
+              {% if interface.addresses %}
+              <table class="device-addresses">
+                  <thead>
+                      <tr>
+                          <th>{% trans 'Address' %}</th>
+                          <th>{% trans 'Mask' %}</th> 
+                          <th>{% trans 'DHCP client ?' %}</th>
+                      </tr>
+                  </thead>
+                  {% for address in interface.addresses %}
+                  <tr>
+                      <td>{{ address.address }}</td>
+                      <td>{{ address.mask }}</td>
+                      {% if address.proto == "static" %}
+                          <td class="static">{% trans 'No' %}</td>
+                      {% else %}
+                          <td class="dhcp">{% trans 'Yes' %}</td>
+                      {% endif %}
+                  </tr>
+                  {% endfor %}
+              </table>
+              {% endif %}
               {% endfor %}
               {% endif %}
             </fieldset>
--- a/openwisp_monitoring/device/tests/test_models.py
+++ b/openwisp_monitoring/device/tests/test_models.py
@@ -1,3 +1,4 @@
+import copy
 import json
 
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
@@ -102,6 +103,32 @@ class BaseTestCase(DeviceMonitoringTestCase):
                     "ssid": "testnet",
                     "tx_power": 6,
                 },
+                "addresses": [
+                    {
+                        "proto": "static",
+                        "family": "ipv4",
+                        "address": "192.168.1.1",
+                        "mask": 8,
+                    },
+                    {
+                        "proto": "dhcp",
+                        "family": "ipv4",
+                        "address": "10.0.0.4",
+                        "mask": 24,
+                    },
+                    {
+                        "proto": "static",
+                        "family": "ipv6",
+                        "address": "2001:3238:dfe1:ec63::fefb",
+                        "mask": 12,
+                    },
+                    {
+                        "proto": "dhcp",
+                        "family": "ipv6",
+                        "address": "fd46:9038:f983::1",
+                        "mask": 22,
+                    },
+                ],
             },
         ],
         "arp_table": [
@@ -222,6 +249,19 @@ class TestDeviceData(BaseTestCase):
             pass
         else:
             self.fail('metric was not deleted')
+
+    def test_bad_address_fail(self):
+        dd = self._create_device_data()
+        data = copy.deepcopy(self._sample_data)
+        data['interfaces'][1]['addresses'][0]['address'] = '123'
+        try:
+            dd.data = data
+            dd.validate_data()
+        except ValidationError as e:
+            self.assertIn('Invalid data in', e.message)
+            self.assertIn("'123\' is not a \'ipv4\'", e.message)
+        else:
+            self.fail('ValidationError not raised')
 
 
 class TestDeviceMonitoring(BaseTestCase):


openwisp_monitoring/device/models.py
```diff
@@ -74,6 +74,11 @@ def data_user_friendly(self):
             # convert to GHz
             if 'wireless' in interface and 'frequency' in interface['wireless']:
                 interface['wireless']['frequency'] /= 1000
+            # TODO: Refactor this to a suitable form in which address can be retrieved
+            if 'addresses' in interface:
+                data['addresses'] = {}
+                for count, address in enumerate(interface['addresses']):
+                    data['addresses']['address' + str(count)] = address['address']
```
nemesisdesign, line 81: unless I'm missing something, I would keep this data structure as is and just display it as a table in the device status tab.

nepython:
@nemesisdesign, I tried out few styles for the table (I think, I am pretty bad at CSS :sweat_smile: but I am improving). Can you please tell me is this what you have in mind.
DeviceData_table
I thought it would be appropriate to highlight the ones with dhcp client.

openwisp_monitoring/device/static/monitoring/css/monitoring.css
```diff
@@ -47,3 +47,25 @@ td.field-health_status,
 .health-ok{ color: green }
 .health-problem{ color: orange }
 .health-critical{ color: red }
+
+table {
```
nemesisdesign, line 51: it's not a good idea to use `table` without referring it to a class or id, because it will be applied to any table, although this css is loaded only in a few pages so it won't affect the rest of the application, but do not get used to bad habits.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -120,6 +120,29 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   {% endif %}
                   {% endif %}
               </div>
+              <div class="random">
+                {% if interface.addresses %}
+                <h2>{% trans 'ADDRESSES' %}:</h2>
+                  <table style="width:100%">
+                    <tr>
+                      <th>Address</th>
```
nemesisdesign, line 128: Should be `{% translate 'Address' %}`, same applies to the other strings.

openwisp_monitoring/device/static/monitoring/css/monitoring.css
nepython: I was going to change it as I noticed with it the device change list displays terribly :laughing:, expected after all it is also a `table`

nepython:
@nemesisdesign, the PR is not yet ready. I just read your `gitter` advice to use https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/lib/openwisp/net.lua inorder to retrieve the addresses, missed to read it somehow. I tested it by introducing the data via django shell :sweat_smile:. Let me try this and update the PR accordingly

openwisp_monitoring/device/schema.py
```diff
@@ -142,6 +142,23 @@
...
+                        "uniqueItems": True,
+                        "additionalItems": True,
+                        "items": {
+                            "additionalProperties": True,
+                            "title": "Address",
+                            "type": "object",
+                            "properties": {
+                                "proto": {"type": "string"},
+                                "family": {"type": "string"},
+                                "address": {"type": "string", "format": "email"},
```
PabloCastellano, line 157: Is it a mistake?

```
                    "ip_address": {
                        "type": "string",
                        "anyOf": [{"format": "ipv4"}, {"format": "ipv6"}],
                    },
```

should work better

nepython: Well, I checked up all formats and nothing was validating so I was figuring it out then. Now it validates, just read the jsonschema docs again and found out `format_checker` isn't enabled by default, changed it and it validates all formats perfectly. 
Thanks, for the suggestion. Any reason why you prefer `anyOf` over `oneOf`, I haven't used **jsonschema** much, hence curious :smiley:

nepython:
@nemesisdesign, it is ready for review. I am working on getting the addresses from the interface currently, it would be great if you can pass few instructions to get started with [this](https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/lib/openwisp/net.lua) and I have updated the styling slightly as in the screenshot :point_up:.

nepython:
@nemesisdesign, rebased the PR and tested it by using the script **netjson-monitoring-wan.lua** as config, it works perfectly :heart_eyes:. I will open a PR soon in `monitoring-lua` as mentioned in the issue #30

openwisp_monitoring/device/schema.py
PabloCastellano: For this case the result is the same so it doesn't matter. Probably anyOf is faster because of short-circuit evaluation.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -223,6 +238,19 @@ def test_device_deleted(self):
         else:
             self.fail('metric was not deleted')
 
+    def test_bad_address_fail(self):
```
PabloCastellano, line 241: Can you add another test for ipv6?

openwisp_monitoring/device/schema.py
```diff
@@ -145,6 +145,26 @@
...
+                        "type": "array",
+                        "title": "Addresses",
+                        "uniqueItems": True,
+                        "additionalItems": True,
+                        "items": {
+                            "additionalProperties": True,
+                            "title": "Address",
+                            "type": "object",
+                            "properties": {
+                                "proto": {"type": "string"},
```
PabloCastellano, line 158: Possible values for `proto` so far are: `static`, `dhcp`, `dhcpv6`. @nemesisdesign do you think we should validate that proto contains any of these values?

nemesisdesign: netjson has only `static` and `dhcp`, for understanding if it's ipv4 or ipv6 we use `family`. Yes I think it's useful to enforce this as well, we can use `enum` or anything similar

PabloCastellano: So we need to send dhcp from the lua script even if ubus returns dhcpv6, right?

nepython: Same query :smile:

openwisp_monitoring/device/tests/test_models.py
nepython: @PabloCastellano, is it required? In this test, I have put the `ip-address` as some bogus string and the output message mentions `ipv4` specifically as it is the first of the `format` in the two possibilities (I am unsure though, I might be wrong). Do you mean the tests should also validate that the `ip-address` have the format as `family`. If yes, I don't think it can be done with **json-schema** so I will have to try something else.

nepython:
@nemesisdesign, I had rebased previously when #63 got merged. Updated the style, now it is same as `arp-table` besides two minor changes, I kept a `margin-top` for the table as there wasn't any space between `form-row` and the `table` and highlighted the ones having `dhcp` client :blush:

openwisp_monitoring/device/schema.py
nemesisdesign: What do you mean with ubus returns dhcpv6 exactly?

I think the answer to the question is yes, proto is `dhcp` or `static`, family is `ipv4` or `ipv6`.

PabloCastellano: @nemesisdesign I meant that `ubus call network.interface dump` returns `dhcp` and `dhcpv6` as possible values in the `proto` field so we cannot just send the value from ubus

openwisp_monitoring/device/tests/test_models.py
PabloCastellano: I think we don't need it anymore :+1:

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -141,6 +141,31 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   {% endif %}
                   {% endif %}
               </div>
+              <div class = "device_addresses">
+                {% if interface.addresses %}
+                  <h2>{% trans 'ADDRESSES' %}:</h2>
```
nemesisdesign, line 146: there's no need to write it uppercase, that's CSS's job.

The fact that the table contain IP addresses is self explanatory so this heading is not needed. You can remove it.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -141,6 +141,31 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   {% endif %}
                   {% endif %}
               </div>
+              <div class = "device_addresses">
```
nemesisdesign, line 144: the css class should use the dash for consistency.
The div is not needed here, you can elminate the div and move the class to the table element

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -141,6 +141,31 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
...
+                {% if interface.addresses %}
+                  <h2>{% trans 'ADDRESSES' %}:</h2>
+                    <table>
+                      <thead>
+                        <tr>
```
nemesisdesign, line 149: be consistent with indentation, 4 space indentation is being used in this file, so follow it please. Adopt the same methodology on any file you work on please.

PR #72, opened 2020-05-07
[check] OPENWISP_MONITORING_MANAGEMENT_IP_ONLY issues #59

### LINKED ISSUE #59: OPENWISP_MONITORING_MANAGEMENT_IP_ONLY issues
Move to `check` app since it will be only used by active checks.

Set to `False` by default in `tests/openwisp2/settings.py` to allow testing during development (developers use their local network to test and get confused when they don't see the ping check working as expected); this will need some tests to be changed because they expect the value to default to `True`.

NoumbissiValere:
Fixed the issue by moving the MANAGEMENT_IP_ONLY in device
settings to check module settings. and updating the test
cases

Fixes #59

### DIFF
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -8,8 +8,8 @@ from jsonschema.exceptions import ValidationError as SchemaError
 from openwisp_controller.config.models import Device
 
 from ... import settings as monitoring_settings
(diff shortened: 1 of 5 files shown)


NoumbissiValere:
@nemesisdesign i think this issue have been addressed. Please can you review? thanks

nemesisdesign:
Thanks! :+1:

PR #74, opened 2020-05-08
[monitoring] Show associated wifi clients #60

### LINKED ISSUE #60: [monitoring] Show associated wifi clients
We already receive the information of the connected wifi clients, which is used to generate the graph which shows how many wifi clients connect over time.

It would be very useful to also display the list of clients for each interface.

We already show some status information about the wifi interface, eg: `INTERFACE STATUS: WLAN0`. We could add a new section within this table which shows the list of clients, each row could have columns like ma address, wds, ht, vht, wmm, wps.

NoumbissiValere:
Fixed the issue by updating change_form template

Fixes #60


openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -137,6 +137,30 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                       <div class="readonly">
                           {{ interface.wireless.clients|length }}
                       </div>
+                      <div class="readonly">
+                        <table>
+                            <caption>{% trans 'clients' %}</caption>
+                            <tr>
+                                <th>{% trans 'MAC address' %}</th>
+                                <th>wds?</th>
```
nemesisdesign, line 145: in this case I'd remove the `?` and use uppercase since they're all acronyms.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -137,6 +137,30 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
...
+                                <th>ht?</th>
+                                <th>vht?</th>
+                                <th>wmm?</th>
+                                <th>wps?</th>
+                            </tr>
+                            {% for client in interface.wireless.clients %}
+                            <br />
```
nemesisdesign, line 152: break elements cannot be put between table rows

NoumbissiValere:
Thanks for the review. Will update the PR today.

NoumbissiValere:
@PabloCastellano and @nemesisdesign  this is ready for review.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -60,7 +60,7 @@ class BaseTestCase(DeviceMonitoringTestCase):
                 },
                 "wireless": {
                     "channel": 1,
-                    "clients": {},
+                    "clients": [],
```
PabloCastellano, line 63: Please don't leave it empty. Instead add real data so that the testing is more useful for what you have implemented. You can just copy-paste what Federico posted here -> https://github.com/openwisp/openwisp-monitoring/issues/60#issuecomment-625617960

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -94,7 +94,7 @@ class BaseTestCase(DeviceMonitoringTestCase):
...
```
PabloCastellano, line 97: Also add data here, you can use some variants from the above

openwisp_monitoring/device/schema.py
```diff
@@ -145,6 +145,42 @@
...
+                                        "auth": {"type": "boolean"},
+                                        "authorized": {"type": "boolean"},
+                                        "ht": {"type": "boolean"},
+                                        "vht": {"type": "boolean"},
+                                        "wds": {"type": "boolean"},
+                                        "wmm": {"type": "boolean"},
+                                        "wps": {"type": "boolean"},
+                                        "mac": {"type": "string"},
+                                        "mfp": {"type": "boolean"},
+                                        "preauth": {"type": "boolean"},
```
PabloCastellano, line 178: Please add also:

* `additionalProperties` false
* `additionalItems` -> change to false
* `required: [aid, assoc, auth... all of them]`

openwisp_monitoring/device/schema.py
```diff
@@ -145,6 +145,42 @@
...
+                                "additionalItems": True,
+                                "items": {
+                                    "type": "object",
+                                    "title": "Client",
+                                    "properties": {
+                                        "aid": {"type": "integer"},
+                                        "assoc": {"type": "boolean"},
+                                        "auth": {"type": "boolean"},
+                                        "authorized": {"type": "boolean"},
+                                        "ht": {"type": "boolean"},
```
PabloCastellano, line 171: According to https://github.com/openwisp/openwisp-monitoring/issues/60#issuecomment-625617960 there are two missing properties yet: rrm and signature.

openwisp_monitoring/device/schema.py
```diff
@@ -145,6 +145,42 @@
                             "tx_carrier_errors": {"type": "integer"},
                         },
                     },
+                    "wireless": {
+                        "type": "object",
```
PabloCastellano, line 149: Please add the following:

* additionalItems False
* additionalProperties false
* required: [channel, country... all of them]

NoumbissiValere:
ok thanks. will do that

NoumbissiValere:
@PabloCastellano could you please have a look again? :smile:

openwisp_monitoring/device/schema.py
```diff
@@ -145,6 +145,74 @@
...
+                                        "ht": {"type": "boolean"},
+                                        "vht": {"type": "boolean"},
+                                        "wds": {"type": "boolean"},
+                                        "wmm": {"type": "boolean"},
+                                        "wps": {"type": "boolean"},
+                                        "mac": {"type": "string"},
+                                        "mfp": {"type": "boolean"},
+                                        "preauth": {"type": "boolean"},
+                                        "signature": {"type": "string"},
+                                        "rrm": {"type": "array"},
```
PabloCastellano, line 210: If you define rrm as an array, you also need to define the type of items.

Please read this guide, it will help you a lot understading jsonschema: https://json-schema.org/learn/getting-started-step-by-step.html

For rrm I would also add maxItems and minItems, both set to 5. Check out https://json-schema.org/understanding-json-schema/reference/array.html for reference

openwisp_monitoring/device/schema.py
PabloCastellano: You have not added additionalProperties and additionalItems to the correct places.

`additionalProperties` has to be in the same level as `properties` and similarly `additionalItems` for `items`.

I think reading carefully the rest of the schema and following this guide will help you understand it better :+1:

https://json-schema.org/learn/getting-started-step-by-step.html

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -120,19 +122,10 @@ def _data(self):
...
-
-    _garbage_clients = {
-        'type': 'DeviceMonitoring',
-        'interfaces': [
-            {'name': 'garbage1', 'wireless': {'clients': {}}},
-            {'name': 'garbage2', 'wireless': {'clients': [{'what?': 'mac missing'}]}},
-            {'name': 'garbage3', 'wireless': {}},
-            {'name': 'garbage4'},
-        ],
-    }
```
PabloCastellano, line 138: Why did you remove it?

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -218,14 +218,16 @@ def test_200_multiple_measurements(self):
         # expected upload wlan1
         self.assertEqual(data['traces'][1][1][-1], 1.5)
 
-    def test_garbage_clients(self):
```
PabloCastellano, line 221: Why did you remove this test?

NoumbissiValere: Because we've made all the properties required and have equally defined their schemas, meaning there will never be an instance where the system will allowed wireless interface with garbage_client since the schema validation will fail

NoumbissiValere:
@PabloCastellano on a second thought, going through the code today, i realized it is not good to set the properties of `clients` required because a wireless interface will not always have `clients`. i think we need to let them be and just define their schema. 
Please explain to me why it should be required if i am making a mistake. thanks

PabloCastellano:
@NoumbissiValere You are right! `clients` shouldn't be in the required list :+1:

NoumbissiValere:
@PabloCastellano I have updated the PR

NoumbissiValere:
> LGTM
> 
> Some minor comments:
> 
> 1. The following fields are not shown in the device page. I guess we can do it later since that fields are very specific
> 
> * aid
> * assoc
> * auth
> * authorized
> * mfp
> * preauth
> * rrm
> * signature
> 
> 1. The following change is needed or it will crash (this was also fixed by @nepython) in his pending PR:
> 
> ```
> diff --git a/openwisp_monitoring/device/models.py b/openwisp_monitoring/device/models.py
> index 9b282f9..5060f17 100644
> --- a/openwisp_monitoring/device/models.py
> +++ b/openwisp_monitoring/device/models.py
> @@ -68,9 +68,10 @@ class DeviceData(Device):
>                  remove.append(interface)
>                  continue
>              # human readable mode
> -            interface['wireless']['mode'] = interface['wireless']['mode'].replace(
> -                '_', ' '
> -            )
> +            if 'wireless' in interface and 'mode' in interface['wireless']:
> +                interface['wireless']['mode'] = interface['wireless']['mode'].replace(
> +                    '_', ' '
> +                )
>              # convert to GHz
>              if 'wireless' in interface and 'frequency' in interface['wireless']:
>                  interface['wireless']['frequency'] /= 1000
> ```

Since @nepython has already handle this,  I don't need to handle it right?

NoumbissiValere:
Yes @PabloCastellano that's true. I didn't include those fields because in the issue description those fields were not required. Reason while I added just fields which were demanded to the table. Should I include this one to the table as well?

PR #75, opened 2020-05-09
[tests] Pushing up coverage #71

### LINKED ISSUE #71: [tests] Ensure test coverage is above 98%.

nepython:
Closes #71

### DIFF
--- a/openwisp_monitoring/device/settings.py
+++ b/openwisp_monitoring/device/settings.py
@@ -1,38 +1,56 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
+
+def get_critical_device_metrics():
+    critical_metrics = getattr(
+        settings,
+        'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS',
+        [{'key': 'ping', 'field_name': 'reachable'}],
+    )
+    for item in critical_metrics:  # pragma: no-cover
+        try:
+            assert 'key' in item
+            assert 'field_name' in item
+        except AssertionError:
+            raise ImproperlyConfigured(
+                'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS must contain the following keys: key, field_name'
+            )
+    return critical_metrics
+
+
+def get_health_status_labels():
+    labels = getattr(
+        settings,
+        'OPENWISP_MONITORING_HEALTH_STATUS_LABELS',
+        {
+            'unknown': 'unknown',
+            'ok': 'ok',
+            'problem': 'problem',
+            'critical': 'critical',
+        },
+    )
+    try:
+        assert 'unknown' in labels
+        assert 'ok' in labels
+        assert 'problem' in labels
+        assert 'critical' in labels
+    except AssertionError:  # pragma: no-cover
+        raise ImproperlyConfigured(
+            'OPENWISP_MONITORING_HEALTH_STATUS_LABELS must contain the following '
+            'keys: unknown, ok, problem, critical'
+        )
+    return labels
+
+
 SHORT_RETENTION_POLICY = getattr(
     settings, 'OPENWISP_MONITORING_SHORT_RETENTION_POLICY', '24h0m0s'
 )
-CRITICAL_DEVICE_METRICS = getattr(
-    settings,
-    'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS',
-    [{'key': 'ping', 'field_name': 'reachable'}],
-)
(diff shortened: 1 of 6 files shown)


openwisp_monitoring/device/tests/test_settings.py
```diff
@@ -0,0 +1,31 @@
+from unittest.mock import patch
+
+from django.core.exceptions import ImproperlyConfigured
+
+from . import DeviceMonitoringTestCase
+
+
+# TODO: The tests are validating correctly but coverage
+# remains unaffected by them, figure out why?
```
nemesisdesign, line 9: use `ipdb` to inspect the code and brute force it until you find out what is happening

nepython: > Use the brute-force, young padawan.

Learning from the great Jedi :blush: 

I used `ipdb` to inspect and found out the code was raising the error `ImproperlyConfigured` but `ipdb` did't show that any line of `settings.py` got executed while running the test.

After some thoughts I realized `settings.py` was read as a string and then being executed line by line by **exec()** in my previous commit, so `settings.py` was never imported or executed :sweat_smile:. After that, I tried with other options like `os.system()`, `subprocess()` but neither worked so finally did a bad workaround (I don't think it's a nice way of doing it but lines are getting covered), please check it out

openwisp_monitoring/device/settings.py
```diff
@@ -1,38 +1,55 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
+
+def get_or_create_critical_device_metrics():
```
PabloCastellano, line 5: Name it just `get_critical_device_metrics`, you can also remove the comment in the following line because it doesn't provide much value

openwisp_monitoring/device/settings.py
```diff
@@ -1,38 +1,55 @@
...
+    )
+    for item in critical_metrics:  # pragma: no-cover
+        if not all(['key' in item, 'field_name' in item]):
+            raise ImproperlyConfigured(
+                'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS contains invalid items'
+            )
+    return critical_metrics
+
+
+def get_or_create_health_status_labels():
```
PabloCastellano, line 20: Same: `get_health_status_labels`

openwisp_monitoring/device/settings.py
```diff
@@ -1,38 +1,55 @@
...
+
+def get_or_create_critical_device_metrics():
+    # defines ``critical_device_metrics`` if they don't exist and validates them
+    critical_metrics = getattr(
+        settings,
+        'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS',
+        [{'key': 'ping', 'field_name': 'reachable'}],
+    )
+    for item in critical_metrics:  # pragma: no-cover
+        if not all(['key' in item, 'field_name' in item]):
```
PabloCastellano, line 13: Can you use the `assert` approach as it was used before for `OPENWISP_MONITORING_HEALTH_STATUS_LABELS`?

openwisp_monitoring/device/settings.py
```diff
@@ -1,38 +1,55 @@
...
+            raise ImproperlyConfigured(
+                'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS contains invalid items'
```
PabloCastellano, line 15: "OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS must contain the following keys: key, field_name" sounds better

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -35,6 +35,23 @@ def _post_data(self, id, key, data):
...
+                'uptime': 8003,
+                'hostname': '08-00-27-04-8E-1C',
+            },
+            'resources': {
+                'memory': {
+                    'total': 249774080,
+                    'shared': 90112,
+                    'free': 224497664,
+                    'cached': 6774784,
+                    'available': 203497472,
```
PabloCastellano, line 49: available and shared are not expected in the current schema. Same for general->hostname, can you update them in one different PR?

nepython: I copied the actual device data, funny that schema doesn't have `shared` because that is being displayed in `Device Status` :sweat_smile:

PabloCastellano: Yeah the current schema is somewhat flexible and remember e.g. that we were not forcing format checking but as it gets more stable and we are confident with it, IMO we should get stricter and reject metrics that don't follow the schema

PR #76, opened 2020-05-13
[feature] Collect and display memory, CPU and flash usage of devices #43

### LINKED ISSUE #43: [feature] Collect and display memory (RAM), CPU and flash (disk space) usage of devices
This is an idea/draft, the issue can undergo changes/revisions when revisited later. It can be broken down to:-
## 1. Collect Device Info:
* Collect memory (RAM), CPU and flash (disk space) usage of devices via the API. Refer [here](https://groups.google.com/d/msg/openwisp/tinsyzJt3BE/6CQGdd_GBAAJ).
* Convert the collected data into [NETJSON device monitoring specification](http://netjson.org/rfc.html#resources2) compliant format.
## 2. Create relevant charts and metrics:
* Automatically create different metrics and charts (one for memory usage, one for CPU usage, one for flash space usage) whenever this data is received and display the same in `Device Status`.
## 3. Add settings for this feature:
* Automatically create a related alert setting for each metric which creates an alert 
* The setting should set the device monitoring status to `problem` when the total value of each metric goes above X, where X is configurable for each metric with a different setting and defaults to: **95%** for `RAM`, **90%** for `CPU` and **80%** for `flash` usage.
## 4. Document the changes
* Prepare new test-cases and update the docs about the setting and add illustrative pictures of the above added charts.

nepython:
* [x] Collect Device Info (will open a PR soon in `lua-monitoring`)
* [x] Create relevant charts and metrics:
* [x] Added threshold and a setting with the `default` as mentioned in the issue (I need to test it)
* [x] Update the tests

Closes #43


openwisp_monitoring/device/api/views.py
```diff
@@ -209,7 +233,10 @@ def _calculate_increment(self, ifname, stat, value):
         # get previous counters
         data = self._previous_data
         try:
-            previous_counter = data['interfaces_dict'][ifname]['statistics'][stat]
+            if ifname is None:
```
PabloCastellano, line 236: Are you sure this is correct? if ifname is None, then `data['interfaces_dict'][ifname]['statistics'][stat]` will fail

openwisp_monitoring/device/settings.py
```diff
@@ -54,3 +54,9 @@ def get_health_status_labels():
 DEVICE_RECOVERY_DETECTION = getattr(
     settings, 'OPENWISP_MONITORING_DEVICE_RECOVERY_DETECTION', True
 )
+
+DEVICE_RESOURCES_THRESHOLDS = getattr(
+    settings,
+    'OPENWISP_MONITORING_DEVICE_RESOURCES_THRESHOLDS',
+    {'memory': 0.95, 'load': 0.90, 'disk_usage': 80},
```
PabloCastellano, line 61: If memory is 0.95 and load is 0.90, shouldn't disk_usage be 0.8?

openwisp_monitoring/device/api/views.py
```diff
@@ -200,7 +202,29 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        for resource, value in data.get('resources', {}).items():
```
PabloCastellano, line 205: This for loop looks weird.
Instead of iterating over data.get('resources') IMHO it makes more sense to directly iterate over self.resources like

```
for resource in self.resources:
    value = data.get('resources', {}).get(resource)
```

or directly (even better):

```
if 'memory' in data.get('resources', {}):
   ...
if 'load' in data.get('resources', {}):
   ...
if 'disk_usage' in data.get('resources', {}):
   ...
```

Why are you using data.get('resources') instead of data['resources']? Is there any case when resources will not be available?

openwisp_monitoring/device/api/views.py
```diff
@@ -200,7 +202,29 @@ def _write(self, request, pk):
...
+            if resource not in self.resources:
+                continue
+            if resource == 'load':
+                self.resources_metric_write(
+                    pk=pk, ct=ct, resource=resource, key='load', value=value[1]
```
PabloCastellano, line 210: You can use key=resource

openwisp_monitoring/device/api/views.py
nepython: It should be exact opposite, made some last minute changes yesterday. Now I get why those 3 tests where failing lately :sweat_smile:

openwisp_monitoring/device/settings.py
nepython: Very well, the format in which I was passing value to the metric was like this. I will change it, to be more consistent.

openwisp_monitoring/device/api/views.py
nepython: I used it for two reason one was to maintain consistency and the other was to avoid garbage clients or blank data from failing the API but better would be to use an `if` it doesn't make much sense to use a `for` as I had done so updated that :smile:

openwisp_monitoring/device/settings.py
```diff
@@ -58,3 +58,9 @@ def get_health_status_labels():
 MAC_VENDOR_DETECTION = getattr(
     settings, 'OPENWISP_MONITORING_MAC_VENDOR_DETECTION', True
 )
+
+DEVICE_RESOURCES_THRESHOLDS = getattr(
+    settings,
+    'OPENWISP_MONITORING_DEVICE_RESOURCES_THRESHOLDS',
+    {'memory': 95, 'load': 90, 'disk_usage': 80},
```
PabloCastellano, line 65: Sorry I wasn't clear enough. IMO it's better having the values as 0.95, 0.9, 0.8

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,20 @@ def _write(self, request, pk):
...
+                )
+            if 'disk_usage' in data['resources']:
+                usage = 0
+                for disk in data['resources']['disk_usage']:
+                    usage += disk['percent_used']
+                self.resources_metric_write(pk, ct, key='disk_usage', value=usage)
+            if 'memory' in data['resources']:
+                value = data['resources']['memory']
+                used = 1 - (value['free'] - value['buffered']) / value['total']
+                self.resources_metric_write(pk, ct, key='memory', value=100 * used)
```
PabloCastellano, line 217: Better do the full calculus in one line and then just `self.resources_metric_write(pk, ct, key='memory', value=used)`

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,20 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        if 'resources' in data:
+            if 'load' in data['resources']:
+                self.resources_metric_write(
+                    pk, ct, key='load', value=100 * data['resources']['load'][1]
+                )
```
PabloCastellano, line 208: We should save the three values in `data['resources']['load']` with keys `load_1`, `load_5` and `load_15`. There is no need to save 100 * value

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,20 @@ def _write(self, request, pk):
...
+            if 'disk_usage' in data['resources']:
+                usage = 0
+                for disk in data['resources']['disk_usage']:
```
PabloCastellano, line 211: @nemesisdesign What do you think about this? Should we save the value aggregated or one value for each filesystem?

In any case, saving the sum of `disk['percent_used']` doesn't make much sense, at least `disk['percent_used'] / num_elements`

openwisp_monitoring/device/api/views.py
PabloCastellano: @nemesisdesign Same here, what value do you think we should save?

openwisp_monitoring/device/settings.py
nepython: It can't be, they need to be integers as I realized a bit late that threshold can only be an integer.

openwisp_monitoring/device/api/views.py
nepython: @PabloCastellano, I need to save 3 metrics with different keys and values what I think would be best to avoid redundancy is store the keys and values in a dictionary and then use a for metric_write. Let me know, if this seems fine.

openwisp_monitoring/device/api/views.py
nepython: There is a need as I mentioned above threshold value can only be an integer, hence `*100`. @PabloCastellano , saving the three values doesn't matter I think (I might be wrong) since the device status gets updated every 5 mins and hence I had used that value only specifically. My attempt was to store only necessary info on influxdb.

PabloCastellano: It doesn't matter if the data is collected every 5min. You may want to check the history of cpu load. Don't be afraid of saving useful data :-)

openwisp_monitoring/device/settings.py
PabloCastellano: I see now, then OK, leave it like that and we can discuss about changing it later

openwisp_monitoring/device/api/views.py
PabloCastellano: Why don't you just save data['resources']['load'] (the full list, without accessing the indexes)?

nepython: `InfluxDB` doesn't allow it, I tried it.

openwisp_monitoring/device/api/views.py
```diff
@@ -248,5 +265,31 @@ def _create_clients_graph(self, metric):
         graph.full_clean()
         graph.save()
 
+    def _create_resources_graph(self, metric, resource):
+        graph = Graph(metric=metric, configuration=resource)
+        graph.full_clean()
+        graph.save()
+
+    def resources_metric_write(self, pk, ct, key, value):
```
PabloCastellano, line 273: Please use more descriptive parameter names for pk and ct

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,22 @@ def _write(self, request, pk):
...
+        if 'resources' in data:
+            resources = {}
+            if 'load' in data['resources']:
+                resources['load'] = 100 * data['resources']['load'][1]
+            if 'disk_usage' in data['resources']:
+                used_space, total_size = 0, 0
+                for disk in data['resources']['disk_usage']:
+                    used_space += disk['used_space']
+                    total_size += disk['size']
+                resources['disk_usage'] = 100 * used_space / total_size
```
PabloCastellano, line 213: I think that by doing this we are wasting the data sent by the monitoring script here.

Imagine you are an operator wanting to know the same information that you would get by ssh'ing into the device and just get an average. Some mount points are full and you don't even notice (!)

Please just save all the data sent from the monitoring script and don't early optimize that much. We can change it later. I propose storing load_1, load_5 and load_15 and just enable by now automatically the load_5 graph. WDYT?

openwisp_monitoring/device/api/views.py
```diff
@@ -248,5 +265,31 @@ def _create_clients_graph(self, metric):
...
+        graph.save()
+
+    def resources_metric_write(self, pk, ct, key, value):
+        if key not in monitoring_settings.AUTO_GRAPHS:
+            return
+        metric, created = Metric._get_or_create(
+            object_id=pk,
+            content_type=ct,
+            key=key,
+            field_name=key,
```
PabloCastellano, line 280: Using field_name=key=key is not very useful when storing the disk usage data. Here is my proposal, let me know what you think:

* name: "mountpoint /" (using key "mounted_on")
* field_name: "used_space" and  "size" (save these 2 metrics)
* key "/dev/sda1" (using key "name")

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,22 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        if 'resources' in data:
+            resources = {}
+            if 'load' in data['resources']:
+                resources['load'] = 100 * data['resources']['load'][1]
```
PabloCastellano, line 207: Why multiplying by 100? This would be confusing for the enduser and Influxdb allows float values.

openwisp_monitoring/device/api/views.py
```diff
@@ -248,5 +265,31 @@ def _create_clients_graph(self, metric):
         graph.full_clean()
         graph.save()
 
+    def _create_resources_graph(self, metric, resource):
+        graph = Graph(metric=metric, configuration=resource)
```
PabloCastellano, line 269: Let's add `if resource not in monitoring_settings.AUTO_GRAPHS` similarly to `_create_clients_graph()`

openwisp_monitoring/device/api/views.py
nepython: Threshold value can only be an integer. From what I know, load is a value less than 1 (in case it's not the case please correct me), hence had to do it

openwisp_monitoring/device/api/views.py
nepython: Good idea for `disk_usage`, I can use the `field_name` say `used_memory` (and similarly `shared` and `buffered` when adding them) for `memory`. Any suggestions, what can I name the `field_name` for the `metric` saving `load` data.

PabloCastellano: Yeah, load_1, load_5 and load_15 look good to me

openwisp_monitoring/device/api/views.py
PabloCastellano: If the reason of multiplying the value * 100 is that Threshold only allows integers then I think it makes more sense to change Threshold.

For the moment we can just create the Metric and skip the Threshold for this case.

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,22 @@ def _write(self, request, pk):
...
+            if 'disk_usage' in data['resources']:
+                used_space, total_size = 0, 0
+                for disk in data['resources']['disk_usage']:
+                    used_space += disk['used_space']
+                    total_size += disk['size']
+                resources['disk_usage'] = 100 * used_space / total_size
+            if 'memory' in data['resources']:
+                memory = data['resources']['memory']
+                used = 1 - (memory['free'] - memory['buffered']) / memory['total']
+                resources['memory'] = 100 * used
```
PabloCastellano, line 217: Same here, save all data, and then we can calculate averages and so on from the influxdb query

PabloCastellano:
@nepython Regarding documenting the changes, I have opened an issue to track it -> https://github.com/openwisp/openwisp-monitoring/issues/81 Don't worry about documentation in this PR

openwisp_monitoring/device/api/views.py
nepython: @PabloCastellano, regarding warning the `operator` I had a suggestion. How's this, if disk_usage of a `mounted_space` exceeds a `threshold` say `80%` we notify the user displaying the `mounted_space` name and `used` percentage. In my opinion it won't be a good idea to display all the `mounted points` on the `graph` because they might just be too many. It's 7 in my case and will vary as per device (also `charts` don't support dynamic values currently, if I am not wrong).

nemesisdesign: I would say let's save the raw data but for the moment focus on displaying only the total like you're doing now.

In the future when we want to start displaying more data, we'll have the data already.

> also charts don't support dynamic values currently, if I am not wrong

Yes it's supported with the usage of `top_fields`, if you specify `top_fields` and set it to 10, it will do a query to get the 10 keys with most results and will then build the final query to use those fields.

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,38 @@ def _write(self, request, pk):
...
                 self._create_clients_graph(metric)
+        if 'resources' in data:
+            resources = {}
+            extra_values = {}
+            if 'load' in data['resources']:
+                extra_values['load'] = {
+                    'load_1': 100 * data['resources']['load'][0],
+                    'load_15': 100 * data['resources']['load'][2],
+                }
+                resources['load'] = 100 * data['resources']['load'][1]
```
nemesisdesign, line 212: @PabloCastellano reminded me that we need to know the number of processors in order to calculate the CPU usage, read here: https://blog.appsignal.com/2018/03/28/understanding-system-load-and-load-averages.html

Can you add this information to the payload that it's sent?

On OpenWRT you can get the number of processors with:

```
cat /proc/cpuinfo | grep -c processor
```

(@PabloCastellano `numprocs` is not available by default on OpenWRT)

openwisp_monitoring/device/api/views.py
nemesisdesign: I would say let's save the raw data but for the moment focus on displaying only the total like you're doing now.

In the future when we want to start displaying more data, we'll have the data already.

> also charts don't support dynamic values currently, if I am not wrong

Yes it's supported with the usage of `top_fields`, if you specify `top_fields` and set it to 10, it will do a query to get the 10 keys with most results and will then build the final query to use those fields.

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,38 @@ def _write(self, request, pk):
...
+            if 'disk_usage' in data['resources']:
+                used_space, total_size = 0, 0
+                for disk in data['resources']['disk_usage']:
```
PabloCastellano, line 215: Remember to change the key names: https://github.com/openwisp/lua-monitoring/pull/11/files#r425648187 and save **all** received raw data.

Also notice that you will be receiving `data['resources']['disk']['used_percent']` so you don't need to divide used_space by total_size. Also don't modify the value *100, just save the raw data at the moment.

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,38 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        if 'resources' in data:
```
PabloCastellano, line 204: Can you do `if 'resources' not in data: return` here? So we avoid one extra level of indentation

openwisp_monitoring/device/api/views.py
PabloCastellano, line 212: Please do the following changes:

Save the key as load_1, and for extra values use load_5 and load_15. Remember to not multiply *100. We will have float Thresholds soon :+1:

openwisp_monitoring/device/api/views.py
```diff
@@ -248,5 +281,40 @@ def _create_clients_graph(self, metric):
...
+            return key
+
+    def _create_resources_graph(self, metric, resource):
+        if resource not in monitoring_settings.AUTO_GRAPHS:
+            return
+        graph = Graph(metric=metric, configuration=resource)
+        graph.full_clean()
+        graph.save()
+
+    def resources_metric_write(
```
PabloCastellano, line 298: This function is becoming a monster :grin:. I propose removing it and `_get_field_name` too because they are complicating the readability too much only trying to make the code less repetitive.

Once removed, duplicate the code that gets_or_create the metric, write it and create the graph if needed.

E.g.

```

if 'resources' in data:
	 if 'load' in data['resources']:
		...calculate value and extra_values...
		... Metric.get_or_create() ...
		... metric.write ...
		... if created: create graphs and thresholds ...
	if 'disk_usage' in data['resources']:
                ... same ...
	if 'memory' in data['resources']:
                ... same ...
```

Then we can refactor once we find similarities in the algorithm

openwisp_monitoring/device/settings.py
PabloCastellano, line 65: Let's use these as default thresholds: `free_memory`, `load_1` and `free_space` keys.

openwisp_monitoring/settings.py
```diff
@@ -3,5 +3,14 @@
...
     settings,
     'OPENWISP_MONITORING_AUTO_GRAPHS',
-    ('traffic', 'wifi_clients', 'uptime', 'packet_loss', 'rtt',),
+    (
+        'traffic',
+        'wifi_clients',
+        'uptime',
+        'packet_loss',
+        'rtt',
+        'memory',
```
PabloCastellano, line 12: Also update the names here

openwisp_monitoring/device/settings.py
nepython: A threshold value needs to be a percentage here, if I am not wrong. The reason being `free_memory` and `free_space` will vary from device to device we can't set a default threshold as per one device, right? I can use `load_1` as a threshold because it's value will lie between 0 and 1 always.

nemesisdesign: are you sure it's comprised between 0 and 1 only? I knew it can be more than 1, especially on multi core devices. A load of 2 with 2 cores is the same of 1 with 1 core.
But couldn't also happen to have a load of 1.1 with 1 core @PabloCastellano? If I'm not wrong this can happen and indicates an overload.

nepython: I never observed it as my device (VM) is single core :sweat_smile:.
Should I use ```load = load_5/(no.of cores)```, that should be consistent, I guess.

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,62 @@ def _write(self, request, pk):
...
+            }
+            metric, created = Metric._get_or_create(
+                object_id=pk,
+                content_type=ct,
+                key='load',
+                field_name='load',
+                name='Resources: Load',
+            )
+            metric.write(
+                data['resources']['load'][1] / data['resources']['cpus'],
```
PabloCastellano, line 220: Don't worry about the percentage. Just save the raw data and we will make the transformations later for the user.

Just write `data['resources']['load'][0]`

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,62 @@ def _write(self, request, pk):
...
+                extra_values=extra_values,
+            )
+            if created:
+                self._create_resources_graph(metric, resource='load')
+                self._create_resources_threshold(metric, resource='load')
+        if 'disk' in data['resources']:
+            used_bytes, size_bytes = 0, 0
+            for disk in data['resources']['disk']:
+                used_bytes += disk['used_bytes']
+                size_bytes += disk['size_bytes']
```
PabloCastellano, line 230:
```
used_percent = 0
for disk in data['resources']['disk']:
    used_percent += disk['used_percent']
used_percent = used_percent / len(data['resources']['disk']) / 100
```

openwisp_monitoring/device/api/views.py
```diff
@@ -200,6 +201,62 @@ def _write(self, request, pk):
...
+            metric, created = Metric._get_or_create(
+                object_id=pk,
+                content_type=ct,
+                key='disk',
+                field_name='disk',
```
PabloCastellano, line 235: field_name should be `used_disk`

openwisp_monitoring/device/api/views.py
nepython: Saving the number of cores, isn't helpful I guess since it's a constant

openwisp_monitoring/device/api/views.py
nepython: @PabloCastellano, the `df` command currently outputs this on my **VM**
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/root               258016     18792    233984   7% /
tmpfs                   121960        76    121884   0% /tmp
/dev/sda1                16112      3872     11916  25% /boot
tmpfs                      512         0       512   0% /dev
```
Now if you observe the used percent represents percentage usage of that individual filesystem and does not represent the `used_percent` of the whole system, am I missing something here?

PabloCastellano: You are right, that approach is better. Nevermind!

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,64 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        if 'resources' not in data:
+            return
+        if 'load' in data['resources']:
```
nemesisdesign, line 210: ok, this method is becoming huge, let's improve this part (later on we'll likely improve the part above too).

I would do something like the following:

```python
if 'resources' not in data:
    return
self._write_cpu(data)
self._write_disk(data)
self._write_memory(data)
```

and move the relevant logic to each of those methods.

openwisp_monitoring/device/api/views.py
```diff
@@ -251,5 +311,18 @@ def _create_clients_graph(self, metric):
...
+    def _create_resources_graph(self, metric, resource):
+        if resource not in monitoring_settings.AUTO_GRAPHS:
+            return
+        graph = Graph(metric=metric, configuration=resource)
+        graph.full_clean()
+        graph.save()
+
+    def _create_resources_threshold(self, metric, resource):
+        value = DEVICE_RESOURCES_THRESHOLDS[resource]
+        t = Threshold(metric=metric, operator='>', value=value, seconds=0)
```
nemesisdesign, line 323: ok, I think we need a way to specify the default value passed to `seconds`.

I realize now that `seconds` is not a good name because it may not convey what it does, that field indicates the sensitivity of the threshold and is used to avoid constantly sending alerts.

While I was writing this I thought that rather than spending time on this, we'd better investigate in a general solution which I tried to describe in #106.

So nothing to do here now, I just added this comment to explain how and why #106 appeared.

nepython:
>@nepython looks like you overwrote my last changes, one of them was casting the load values always to float.

>IMHO we should avoid git push -f unless we have a good reason.

Well, I am trying to avoid it lately but couldn't avoid in this case because there were some conflicts wiith the master so had to rebase and resolve the conflicts. Thanks for the lost patch :smiley:

PabloCastellano:
@nepython There is no situation where you can't avoid it. Indeed you lost some changes and this is what should be avoid first Next time just merge both branches and diff between them if there are many changes.

openwisp_monitoring/monitoring/charts.py
```diff
@@ -161,7 +161,10 @@
         'order': 260,
         'query': {
             'influxdb': (
-                "SELECT 100 * MEAN(load_1) AS CPU_load FROM {key} WHERE "
+                "SELECT 100 * MEAN(load_1) / MEAN(cpus) AS Average_1m, "
+                "100 * MEAN(load_5) / MEAN(cpus) AS Average_5m, "
+                "100 * MEAN(load_15) / MEAN(cpus) AS Average_15m "
+                "FROM {key} WHERE "
```
nemesisdesign, line 167: why this change? I think we should have only 1 line in the chart as it was before.

openwisp_monitoring/monitoring/charts.py
```diff
@@ -161,7 +161,10 @@
         'order': 260,
         'query': {
             'influxdb': (
-                "SELECT 100 * MEAN(load_1) AS CPU_load FROM {key} WHERE "
+                "SELECT 100 * MEAN(load_1) / MEAN(cpus) AS Average_1m, "
```
nemesisdesign, line 164: let's avoid performing this division each time, we don't need to store the number of CPUs in the timeseries, because as you noted @nepython, this never changes.
I'm not sure if @PabloCastellano told you to do this, but I advise against doing this because it has no value, it only adds operational costs.

openwisp_monitoring/monitoring/charts.py
```diff
@@ -179,7 +182,7 @@
         'order': 270,
         'query': {
             'influxdb': (
-                "SELECT 100 * MEAN(used_disk) AS disk_usage FROM {key} WHERE "
+                "SELECT 100 * MEAN(used_disk) / MEAN(size_bytes) AS disk_usage FROM {key} WHERE "
```
nemesisdesign, line 185: I think this was also better before, I advise avoiding this division.

nepython: @nemesisdesign, the division was taking previously too. `used_bytes/size_bytes` was getting stored in the `influxdb measurement` earlier (As, we need a percentage for the threshold). Now `size_bytes` and `used_bytes` both are getting stored as individual fields. This might be useful in case we need the usage in `bytes` and not `percentage` in the future but on more thoughts, I think in can be easily changed when the need arises (for now it will just consume space in influxdb :sweat_smile:). Hence, changing it back to how it was before.

openwisp_monitoring/monitoring/charts.py
PabloCastellano: @nemesisdesign Yes, that was my idea because IMO influxdb should always have ther raw values but if you prefer to do it otherwise I'm fine with that

openwisp_monitoring/monitoring/charts.py
```diff
@@ -131,6 +131,60 @@
...
+        'title': _('Memory Usage'),
+        'description': _('Percentage of memory (RAM) being used.'),
+        'summary_labels': [_('Memory Usage')],
+        'unit': '%',
+        'colors': [DEFAULT_COLORS[4]],
+        'order': 250,
+        'query': {
+            'influxdb': (
+                "SELECT 100 * (1 - ((MEDIAN(free_memory) + MEDIAN(buffered_memory)) "
+                "/ MEDIAN(total_memory))) AS memory_usage "
```
nemesisdesign, line 145: @nepython @PabloCastellano if there's a way to simplify this query, I would do it as well, with the goal of removing load on influxdb when retrieving the value from the DB.

nepython: It can be done by doing storing the percentage itself in influxdb and retrieving it like I had been doing previously. Also, in that case, only one value will get stored in influxdb. @nemesisdesign @PabloCastellano, should I do it that way then?

PabloCastellano: What about returning `available_memory` instead?

nepython: That might be better as the data stored might be usable somewhere else too later on (percentage might not be usable) but this reminds me did we resolve the threshold default value? I can't recall :sweat_smile:. The threshold would compare with the field `available_memory` which will vary from device to device so we can use a default threshold value only with percentage I think. The other work around might be setting some percentage of total memory as the threshold value to compare with.

nemesisdesign: correct, for the problem of the threshold, we need to keep it in percentage.
Let's keep the raw data, but also go back to store the processed data, which will be faster.

Further optimizations can be done only after trying it in the real world, but storage issues can be dealt with more easily than RAM/CPU usage to do the calculations.

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,74 @@ def _write(self, request, pk):
                 metric.write(client['mac'])
             if created:
                 self._create_clients_graph(metric)
+        if 'resources' not in data:
+            return
+        self._write_cpu(data, pk, ct)
```
PabloCastellano, line 210: Improvement: In these internal methods we don't need to full `data` so we can just pass `data['resources']` as `resources`

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,74 @@ def _write(self, request, pk):
...
+        self._write_disk(data, pk, ct)
+        self._write_memory(data, pk, ct)
+
+    def _write_cpu(self, data, primary_key, content_type):
+        if 'load' in data['resources']:
```
PabloCastellano, line 215: Improvement: same logic as we did already:

```
if 'load' not in data['resources']
    return
...
```

openwisp_monitoring/monitoring/charts.py
PabloCastellano: Last time I checked thresholds worked fine with percentages (0-1).

We are already storing the percentage as the *primary* value:

```
            metric.write(used_bytes / size_bytes)
```

nepython: @PabloCastellano , that's for `Disk` and it's correct. We were talking about `Memory` though. Also, I had one change to recommend, shall we store,
```python
threshold = cpus*DEVICE_RESOURCES_THRESHOLDS['load']
```
Since, currently it's correct threshold only for single core devices. This might be needed as we are **not** storing metric value for load as `load_0 / cpus`

openwisp_monitoring/device/api/views.py
nepython: An extra test was required after this change to maintain the coverage.

PabloCastellano: Well, tbh I think we can assume that if 'resources' is present, then 'load', 'memory' and 'disk' are too. WDYT?

openwisp_monitoring/monitoring/charts.py
PabloCastellano: You are right but same applies for memory:

```
used_memory = 1 - (memory['free'] + memory['buffered']) / memory['total']
metric.write(used_memory, extra_values=extra_values)
```

used_memory is a float between 0 and 1

> threshold = cpus*DEVICE_RESOURCES_THRESHOLDS['load']

Instead of doing this, I would save the percentage in the *primary* key of the Metric. It could be named `cpu_usage` and save `load_1` among the rest of extra values. IMO this way it is clearer because you understand that the threshold is 90% of the usage (0.9). Seeing instead a threshold of 1,8 because it's a 2-cores device is confusing.

openwisp_monitoring/device/api/views.py
nepython: Yes, it's safe unless the user modifies the `lua` script (unlikely, I think). So, shall I remove the lines verifying that `load`, `memory`, `disk` are present in `resources`?

PabloCastellano: Yes please, let's take the opportunity to reduce the complexity :+1:

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -37,13 +37,34 @@ def _data(self):
             'type': 'DeviceMonitoring',
             'general': {'local_time': 1589026500, 'uptime': 8003},
             'resources': {
+                'cpus': 1,
                 'memory': {
                     'total': 249774080,
+                    'shared': 86016,
                     'free': 224497664,
                     'cached': 6774784,
+                    'available': 223397664,
```
nemesisdesign, line 46: add these two properties to the schema

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -37,13 +37,34 @@ def _data(self):
...
+                        'size_bytes': 258016,
+                    },
+                    {
+                        'used_bytes': 3872,
+                        'available_bytes': 11916,
+                        'filesystem': '/dev/sda1',
+                        'mount_point': '/boot',
+                        'used_percent': 25,
+                        'size_bytes': 16112,
+                    },
```
nemesisdesign, line 66: we need a schema for this as well

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -242,6 +243,16 @@ def test_garbage_clients(self):
         )
         self.assertEqual(r.status_code, 400)
 
+    def test_no_resources(self):
+        o = self._create_org()
+        d = self._create_device(organization=o)
+        data = deepcopy(self._data())
```
nemesisdesign, line 249: deepcopy is not needed here because:

```python
def _data(self):
        return {
```

`return` always creates a new data structure.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -267,10 +278,11 @@ def test_get_device_metrics_200(self):
         r = self.client.get(self._url(d.pk.hex, d.key))
         self.assertEqual(r.status_code, 200)
         self.assertIsInstance(r.data['graphs'], list)
-        self.assertEqual(len(r.data['graphs']), 4)
+        self.assertEqual(len(r.data['graphs']), 7)
         self.assertIn('x', r.data)
         graphs = r.data['graphs']
         for graph in graphs:
+            # print(graph)
```
nemesisdesign, line 285: remove this

openwisp_monitoring/settings.py
```diff
@@ -3,5 +3,14 @@
...
+        'load',
```
nemesisdesign, line 13: what do you guys think of renaming this and other occurrences of load (when using internally as metric name, chart name, metric key) to `cpu`? since we have `memory` and `disk`, `load` feels a bit out of place.

I just had to rename the data from influxdb and while looking for it I couldn't find it:

```
> show measurements
name: measurements
name
----
br_lan
br_wan
device_data
disk
eth0
eth0_1
eth0_2
load
memory
ping
tap0
wlan0
wlan1
```
After a while I realized I was looking for `cpu` but it's named `load` from load aveage, but I believe the value we're storing and showing in the "CPU Load" chart doesn't have much to do with load averages anymore and it would be better to refer to it as `cpu`.

@PabloCastellano @nepython

PabloCastellano: I agree. Indeed I opened an issue to review measurements names https://github.com/openwisp/openwisp-monitoring/issues/111

nepython:
All requested changes have been addressed :smile:.
I don't have any idea what's causing build failure in the latest commit since all tests are passing locally :/

openwisp_monitoring/device/schema.py
```diff
@@ -35,7 +35,20 @@
...
                         "free": {"type": "integer"},
                         "buffered": {"type": "integer"},
-                        "cache": {"type": "integer"},
+                        "cached": {"type": "integer"},
+                        "shared": {"type": "integer"},
+                        "available": {"type": "integer"},
+                    },
+                },
+                "disk": {
+                    "type": "array",
```
nemesisdesign, line 44: this schema is wrong. Please see the definition of other arrays and check the jsonschema spec.

Please flag the fields required to do the calculation of the value we store as required.

Please also write a test which sends bogus disk data to ensure the API rejects it.

nepython: > Please flag the fields required to do the calculation of the value we store as required.

I didn't get this.

nemesisdesign: > > Please flag the fields required to do the calculation of the value we store as required.
> 
> I didn't get this.

To flag as required all the fields that are needed for the calculation of the used disk percentage which we graph, otherwise we would be allowing to send data which will not work or may even cause exceptions.

I see that you flagged them all required so that's ok for me.

nemesisdesign: Define the fields of the disk section that you use to do the calculations as required, you did it.

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,72 @@ def _write(self, request, pk):
...
+
+    def _write_memory(self, resources, primary_key, content_type):
+        memory = resources['memory']
+        extra_values = {
+            'total_memory': memory['total'],
+            'free_memory': memory['free'],
+            'buffered_memory': memory['buffered'],
+            'shared_memory': memory['shared'],
+            'available_memory': memory['available'],
+            'cached_memory': memory['cached'],
```
nemesisdesign, line 260: pay attention: you are taking for granted that all the keys of `memory` that you are accessing are present, but this is not reflected in the schema, you should define that these keys are required (except `available`, which on some systems is not present).

Write a test with bogus data for this as well please.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -390,6 +390,49 @@ def test_mac_vendor_info_empty(self):
         dd.save_data()
         self.assertEqual(dd.data['neighbors'][0]['vendor'], '')
 
+    def test_bad_disk_fail(self):
+        dd = self._create_device_data()
+        data = deepcopy(self._sample_data)
+        with self.subTest('Incorrect type'):
```
nemesisdesign, line 396: could you write another test with the following:

```python
del data['resources']['disk']
```

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,72 @@ def _write(self, request, pk):
...
+        metric.write(
+            float(resources['load'][0] / resources['cpus']), extra_values=extra_values,
+        )
+        if created:
+            self._create_resources_graph(metric, resource='cpu')
+            self._create_resources_threshold(metric, resource='cpu')
+
+    def _write_disk(self, resources, primary_key, content_type):
+        used_bytes, size_bytes, available_bytes = 0, 0, 0
+        for disk in resources['disk']:
```
nemesisdesign, line 236: here, I think you're not taking into account the case in which the `disk` key in `resources` is not present.
It may happen. If it does happen, it looks to me that the code will fail and the metric collection will fail.
I think you did check this before, was it @PabloCastellano that suggested you to not do it?

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,72 @@ def _write(self, request, pk):
...
                 self._create_clients_graph(metric)
+        if 'resources' not in data:
+            return
+        self._write_cpu(data['resources'], pk, ct)
+        self._write_disk(data['resources'], pk, ct)
+        self._write_memory(data['resources'], pk, ct)
+
+    def _write_cpu(self, resources, primary_key, content_type):
+        extra_values = {
+            'load_1': float(resources['load'][0]),
```
nemesisdesign, line 216: same here, we're assuming `load` is present, but `load` is not flagged as required in the schema.
So if the monitoring data doesn't include `load`, this will generate an exception and no metric will be collected.

We could make `load` mandatory, but I think not making it mandatory gives us flexibility.

openwisp_monitoring/device/schema.py
```diff
@@ -28,14 +28,43 @@
...
+                "cpus": {"type": "integer"},
                 "memory": {
                     "id": "memory",
                     "type": "object",
                     "properties": {
                         "total": {"type": "integer"},
                         "free": {"type": "integer"},
                         "buffered": {"type": "integer"},
-                        "cache": {"type": "integer"},
+                        "cached": {"type": "integer"},
```
nemesisdesign, line 39: here: define the required keys. Ensure that the key which are not required are accessed using `memory.get('<key>')` (which does not raise `KeyError`) rather than `memory['<key>']`

openwisp_monitoring/device/api/views.py
```diff
@@ -203,6 +205,72 @@ def _write(self, request, pk):
...
+            'cached_memory': memory['cached'],
+        }
+        metric, created = Metric._get_or_create(
+            object_id=primary_key,
+            content_type=content_type,
+            key='memory',
+            field_name='used_memory',
+            name='Resources: Memory',
+        )
+        used_memory = 1 - (memory['free'] + memory['buffered']) / memory['total']
```
nemesisdesign, line 269: as we discussed, here we need is to:

- consider `available` optional (we need a test for this, which tests the metric collection without available, because some systems don't have it and older openwrt versions don't have it)
- if `available` is present, use it instead of `free`, but only if it's greater than `free`

openwisp_monitoring/device/api/views.py
nepython: Yes sometime back https://github.com/openwisp/openwisp-monitoring/pull/76#discussion_r433954476, but I agree with his arguement. @nemesisdesign, the only case in which disk key might not be present is if some other `lua` script is used but then that holds for `memory` and `load` too, right? Shall I change the schema to make these 3 required always or simply revert back to checking if these keys are present and then proceed with the code.

openwisp_monitoring/device/schema.py
nepython: That's already done. I have made all fields required taking into account that all shall be needed for #116

nepython:
I think it would be better to rename `used_memory` to `percent_used` or something similar since currently querying influxdb shows as below
```python
[{'time': 1591464990, 'buffered_memory': 974848, 'cached_memory': 6774784, 'content_type': 'config.device', 'free_memory': 224497664, 'object_id': '51b63501-70b8-47bb-b494-141685863e4b', 'shared_memory': 86016, 'total_memory': 249774080, 'used_memory': 0.09729419481797308}]
```

As can be clearly observed, `used_memory` is the only float (it's percentage after all) and the name thus doesn't seem right to me.

PR #85, opened 2020-05-19
[enhancement] Allow Threshold values to be float

### LINKED ISSUE #82: Threshold should allow float values
At the moment the value field in the Threshold model is `models.IntegerField(help_text=_('threshold value'))`. We should allow float values too.

nepython:
Closes #82


openwisp_monitoring/monitoring/models.py
```diff
@@ -569,13 +569,15 @@ class Threshold(TimeStampedEditableModel):
...
-    value = models.IntegerField(help_text=_('threshold value'))
+    value = models.FloatField(help_text=_('threshold value'))
     seconds = models.PositiveIntegerField(
         default=0, validators=[MaxValueValidator(604800)], help_text=_(_SECONDS_HELP)
     )
 
     def _value_crossed(self, current_value):
         threshold_value = self.value
+        if type(current_value) == int:
+            current_value = float(current_value)
```
nemesisdesign, line 580: is this necessary?

```
>>> 1.0 > 1
False
>>> 1.1 > 1
True
>>> 1 > 1.0
False
>>> 1 == 1.0
True
```

Also, if the data is converted to float there's no need to do this.

nepython: I was thinking likewise, shall I then change this 
```python
method = '__gt__' if self.operator == '>' else '__lt__'
        return getattr(current_value, method)(threshold_value)

```

nemesisdesign: does it work without this addition? If it does, I don't think is necessary. Am I missing something?
Where did the need for this change arise?

nepython: @nemesisdesign, I added to take care of this,
```python
>>> getattr(1, '__gt__')(1.00)
NotImplemented
```
We want a `bool` value, if `NotImplemented` is returned it would raise a `ValueError`

nemesisdesign: I didn't know that!

Issue #28, opened 2020-04-14
[monitoring] Remove description field from metric


nemesisdesign:
Remove description field from metric, is not useful.

nepython:
@nemesisdesign, I think this can be tagged with a **GCI** label

nemesisdesign:
@nepython Ideally we want to do this sooner than GCI.

nemesisdesign:
Closed by #44

PR #91, opened 2020-05-21
[enhancement] Introduce the concept of metric configuration #44

### LINKED ISSUE #44: [enhancement] Introduce the concept of metric type/class/configuration
**Info about metric model**:
The Metric model allows to store different aspects of a metric, for example, the ping metric allows to store the `reachable` result (1 for reachable and 0 for unreachable), the packet loss and the round trip information (rtt_min, rtt_max and rtt_average).

**The Problem**:
Each aspect is stored with a different key in the same metric object of the timeseries DB, but the django model does not show this information and this makes some operations harder (eg: automatically creating charts which uses some or all the aspects of the metric).

**Possible Solution**: 
Introduce the concept of a metric type and have this kind of general information stored on the metric type, provided metric types are extensible and users can define their own types.
~~This could [help](https://github.com/bitlabstudio/django-influxdb-metrics)~~

nepython:
Closes #44, #28

**PS**: I have introduced `related_fields` which might be useful for metrics like `ping` and soon to be added `memory`, `load`, etc.


openwisp_monitoring/check/classes/ping.py
```diff
@@ -174,8 +174,10 @@ def _get_metric(self):
             name=check.name,
             object_id=obj_id,
             content_type=ct,
+            # TODO: Find why metric is not taking these values directly from config
             field_name='reachable',
             key=self.__class__.__name__.lower(),
```
nepython, line 179: I don't get why this is being needed. I tried `ping` configuration across tests and there it works perfectly.

nepython:
For documentation view #92

openwisp_monitoring/monitoring/metrics.py
```diff
@@ -0,0 +1,42 @@
...
+        'key': 'ping',
+        'field_name': 'reachable',
+        'related_fields': ['loss', 'rtt_min', 'rtt_max', 'rtt_avg'],
+    },
+    'traffic': {'label': _('Traffic'), 'key': '{key}', 'field_name': '{field}'},
+    'clients': {'label': _('Clients'), 'key': '{key}', 'field_name': 'clients'},
+}
+
+
+def deep_merge_dicts(dict1, dict2):
```
nemesisdesign, line 17: never duplicate code, follow the DRY principle.
We already added this to openwisp-utils. Look it up and use it please.

openwisp_monitoring/monitoring/migrations/0010_metric_configuration.py
```diff
@@ -0,0 +1,32 @@
...
+    dependencies = [
+        ('monitoring', '0009_allow_float_threshold'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='metric',
+            name='configuration',
+            field=models.CharField(
+                choices=[
```
nemesisdesign, line 17: do it like it's done here please:

https://github.com/openwisp/openwisp-monitoring/blob/c4defa4773f1fc11d5c99ff94dcf0983c9af3ba4/openwisp_monitoring/monitoring/migrations/0006_add_configuration.py#L7-L21

tests/openwisp2/sample_monitoring/migrations/0003_graph_configuration.py
```diff
@@ -0,0 +1,32 @@
...
+    dependencies = [
+        ('sample_monitoring', '0002_add_details_field'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='metric',
+            name='configuration',
+            field=models.CharField(
+                choices=[
```
nemesisdesign, line 17: same as the previous comment

openwisp_monitoring/monitoring/metrics.py
```diff
@@ -0,0 +1,42 @@
+from django.utils.translation import ugettext_lazy as _
+
+from . import settings as app_settings
+
+DEFAULT_METRICS = {
+    'ping': {
+        'label': _('Ping'),
+        'key': 'ping',
+        'field_name': 'reachable',
+        'related_fields': ['loss', 'rtt_min', 'rtt_max', 'rtt_avg'],
```
nemesisdesign, line 10: ok, now let's make this information useful.

I would change `Metric.write()` to enforce that `field_name` and `related_fields` are being used as they're configured here.

So that if by mistake a different string is used, the write will fail with an exception (eg: `ValueError`).

openwisp_monitoring/monitoring/metrics.py
nepython: I didn't knew that, shall I do the same for `charts.py`?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -39,12 +40,14 @@
 
 
 class AbstractMetric(TimeStampedEditableModel):
+    METRICS = get_metric_configuration()
     name = models.CharField(max_length=64)
     description = models.TextField(blank=True)
```
nemesisdesign, line 45: take advantage of this to remove this field and close https://github.com/openwisp/openwisp-monitoring/issues/28

README.rst
```diff
@@ -391,6 +391,40 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+    OPENWISP_MONITORING_METRICS = {
+        'clients': {
+            'label': _('Clients'),
+            'key': '{key}',
+            'field_name': 'wifi_clients',
+        },
+    }
+
+**NOTE**: In order to make use of ``{key}`` and ``{field}`` in your custom metric configuration,
+you need to define the name in the following format ``<key> <field>``.
```
PabloCastellano, line 423: I don't understand this sentence, can you reword it? Also, shouldn't it be `field_name` instead of `field`?

README.rst
```diff
@@ -391,6 +391,40 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+
+For example, the metric name ``Resources Memory`` will replace all instances of ``{key}`` with ``Resources``
```
PabloCastellano, line 425: I don't understand this sentence neither. Actually it is `Resources: Memory` with a colon

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -39,12 +40,13 @@
...
+    METRICS = get_metric_configuration()
     name = models.CharField(max_length=64)
-    description = models.TextField(blank=True)
-    key = models.SlugField(
-        max_length=64, blank=True, help_text=_('leave blank to determine automatically')
-    )
+    key = models.SlugField(max_length=64, blank=True)
     field_name = models.CharField(max_length=16, default='value')
+    configuration = models.CharField(
+        max_length=16, null=True, choices=get_metric_configuration_choices()
```
PabloCastellano, line 48: Not really sure about the chocies parameter. Everytime you add a new metric configuration will make Django generate a new migration. @nemesisdesign WDYT?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -64,9 +66,20 @@ def __str__(self):
...
 
+    # TODO: Improve this!
     def clean(self):
         if not self.key:
-            self.key = self.codename
+            try:
+                self.key = self.config_dict['key'].format(key=self.name.split()[0])
+            except KeyError:
+                self.key = self.codename
+        self.key = self._makekey(self.key)
```
PabloCastellano, line 76: What do you think about moving 

```
        if not self.key:
            try:
                self.key = self.config_dict['key'].format(key=self.name.split()[0])
            except KeyError:
                self.key = self.codename
```

inside _makekey too?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -96,6 +107,22 @@ def codename(self):
...
+            raise InvalidMetricConfigException(
+                f'Invalid metric configuration: "{self.configuration}"'
+            )
+
+    @property
+    def related_fields(self):
+        try:
+            return self.config_dict['related_fields']
+        except KeyError:
+            return None
```
PabloCastellano, line 124: IMO, it should return the empty list for consistency. You can simplify it with

```
return self.config_dict.get('related_fields', [])
```

openwisp_monitoring/monitoring/migrations/0011_remove_metric_description.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('monitoring', '0010_metric_configuration'),
+    ]
+
+    operations = [
+        migrations.RemoveField(model_name='metric', name='description',),
+    ]
```
PabloCastellano, line 14: There's no need to have two migrations. I would merge both into just one

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,6 +181,9 @@ def write(self, value, time=None, database=None, check=True, extra_values=None):
         """ write timeseries data """
         values = {self.field_name: value}
         if extra_values and isinstance(extra_values, dict):
+            for key in extra_values.keys():
+                if key not in self.related_fields:
```
PabloCastellano, line 185: I think this line will fail when self.related_fields is None because NoneType is not iterable

openwisp_monitoring/monitoring/tests/test_models.py
```diff
@@ -77,11 +91,18 @@ def test_write(self):
         self.assertEqual(measurement['value'], 2)
 
     def test_general_write(self):
-        m = self._create_general_metric(name='Sync test')
+        # TODO: Remove or modify this test
```
PabloCastellano, line 94: Why?

openwisp_monitoring/monitoring/tests/test_models.py
```diff
@@ -77,11 +91,18 @@ def test_write(self):
...
+        m = self._create_general_metric(name='Sync test', key='Sync test')
         m.write(1)
         measurement = list(query('select * from sync_test').get_points())[0]
         self.assertEqual(measurement['value'], 1)
 
+    def test_metric_write_wrong_related_fields(self):
+        m = self._create_general_metric(name='ping', configuration='ping')
+        extra_values = {'reachable': 0, 'rtt_avg': 0.51, 'rtt_max': 0.6, 'rtt_min': 0.4}
+        with self.assertRaises(ValueError):
```
PabloCastellano, line 103: It would assert also that  the exception message contains 'reachable not defined in metric configuration'

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,6 +181,9 @@ def write(self, value, time=None, database=None, check=True, extra_values=None):
...
+                    raise ValueError(f'{key} not defined in metric configuration')
```
PabloCastellano, line 186: I'd add double quotes around {key} to make it easier to understand when reading the error message that key is a configuration value

openwisp_monitoring/monitoring/tests/test_models.py
nepython: It needs to modified since the `query` used in the test is specific for `influxDB`. I don't recall though why I mentioned `remove`

openwisp_monitoring/monitoring/base/models.py
nepython: I think failure is what's expected as `related_fields` shouldn't be **None** for a metric configuration receiving `extra_values`.

openwisp_monitoring/monitoring/migrations/0011_remove_metric_description.py
nepython: Well, there are two separate issues here. I preferred a separate migration as it's easier when we migrate back for an case (just in case). Let me know if I should merge them both, I would prefer to keep it this way.

openwisp_monitoring/monitoring/base/models.py
PabloCastellano: Yes, but at the moment None is the returned value when the KeyError exception is risen

nepython: Should be taken care of, now :smiley:

openwisp_monitoring/monitoring/base/models.py
PabloCastellano: My concern is solved using the `register_chart` function :+1:

openwisp_monitoring/monitoring/migrations/0011_remove_metric_description.py
PabloCastellano: I still don't see the benefit. Also we don't migrate back often

nepython: Very well then, I shall merge both of them

README.rst
```diff
@@ -391,6 +391,58 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+
+    OPENWISP_MONITORING_METRICS = {
+        'resources': {
+            'label': _('Resources'),
+            'key': '{key}',
+            'field_name': '{field_name}',
+        },
+    }
+
+If the metric with ``resources`` configuration is named as **Resources Memory** then it's ``key``
```
PabloCastellano, line 442: it's -> its

nepython:
I was forced to introduce 
https://github.com/openwisp/openwisp-monitoring/pull/91/files#diff-1725f73787a09eb5f7e86f63c82290dfR184 as the default django manager was not calling the `full_clean` method. The alternative way might be to call `Metric._get_or_create` but I didn't think it is a good practice so didn't do so.

openwisp_monitoring/check/classes/ping.py
```diff
@@ -178,10 +178,10 @@ def _get_metric(self):
             name=check.name,
             object_id=obj_id,
             content_type=ct,
-            field_name='reachable',
-            key=self.__class__.__name__.lower(),
+            configuration=self.__class__.__name__.lower(),
         )
         metric, created = Metric.objects.get_or_create(**options)
+        metric.full_clean()
```
PabloCastellano, line 184: I think it's fine to call _get_or_create

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -79,8 +89,6 @@ def _get_or_create(cls, **kwargs):
         like ``get_or_create`` method of django model managers
         but with validation before creation
         """
-        if 'key' in kwargs:
-            kwargs['key'] = cls._makekey(kwargs['key'])
```
PabloCastellano, line 83: So if the key is not specified, will it be blank?

openwisp_monitoring/check/classes/ping.py
nepython: Very well then :smiley:

README.rst
```diff
@@ -391,6 +391,58 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+            'label': _('Resources'),
+            'key': '{key}',
+            'field_name': '{field_name}',
+        },
+    }
+
+If the metric with ``resources`` configuration is named as **Resources Memory** then its ``key``
+would be **Resources** and ``field_name`` shall be **Memory** respectively.
+If you name it as **Memory**, then both ``key`` and ``field_name`` shall have the same value **Memory**
+
```
PabloCastellano, line 445: I find it simpler and clearer if the user defines their own key and field_name values, instead of guessing them from the metric name. Did you take this approach for any special reason?

nepython: That option is always open, the user can always define `key` and `field_name` as per their wish. I have provided the option of not defining it and let it be automatically determined for the case of avoiding redundancy.
Use case, let's assume there are metrics which are similar in nature except for the `key`, `field_name` (like they can be of same type like resources metrics). In that case, it would be better to have a common configuration for them and let the `key` and `field_name` be automatically determined from the name (than creating 3-4 individual configurations).
I think that's the purpose of configuration to label the ones that are alike, right?
Also, `key` and `field_name` were optional field previously with `key` taking the value from name and `field_name` defaulting to `value` so I thought better not to change it.

PR #92, opened 2020-05-21
[docs] Document metrics stored in influxdb #81

### LINKED ISSUE #81: Document metrics stored in influxdb
We should collect in a documentation page all the metrics that we are storing in InfluxDB at the moment, including name, description, value type and keys.

nepython:
Blocked by #91, #76
Closes #81


nepython:
> I propose using an independent METRICS.rst file for this kind of documentation.

I too had been thinking of dividing `README.rst` since it's already having optimum length (I think so, unsure how it's defined in `OpenWISP`) and it's length might only increase in coming weeks.

README.rst
```diff
@@ -22,13 +25,13 @@ OpenWISP 2 monitoring module (Work in progress).
 Available Features
 ------------------
 
-* Displays ``Uptime``, ``Memory Status``, ``Load Status``, ``Addresses``, ``Neighbours`` of the **Device**
-* Displays ``Monitoring Graphs`` namely ``Uptime``, ``Packet Loss``, ``Round Trip Time``, ``Traffic`` in the form of lively charts
+* Displays `Uptime, Memory Status, Load Status, Addresses, Neighbours of the **Device** <docs/device_status.png>`_
```
PabloCastellano, line 28: \*\*Device\*\*  -> Device (the link shows the asterisks, not bold)

README.rst
```diff
@@ -22,13 +25,13 @@ OpenWISP 2 monitoring module (Work in progress).
...
+* Displays ``Monitoring Graphs`` namely ``Uptime``, ``Packet Loss``, ``Round Trip Time``, ``Traffic`` in the form of `lively charts <#default-metrics>`_
```
PabloCastellano, line 29: Uptime is already mentioned in the first bullet point

README.rst
```diff
@@ -391,6 +394,40 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+
+    OPENWISP_MONITORING_METRICS = {
+        'clients': {
+            'label': _('Clients'),
+            'key': '{key}',
+            'field_name': 'wifi_clients',
+        },
+    }
+
+**NOTE**: Inorder to make use of ``{key}`` and ``{field}`` in your custom metric configuration,
```
PabloCastellano, line 425: typo: Inorder -> In order

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|     key      |                   ``disk``                      |
++--------------+-------------------------------------------------+
+|  value type  |                   ``float``                     |
++--------------+-------------------------------------------------+
+|  field_name  |                 ``used_disk``                   |
++--------------+-------------------------------------------------+
+|configuration |                   ``disk``                      |
++--------------+-------------------------------------------------+
+|    charts    |                  [``load``]                     |
```
PabloCastellano, line 588: Is `load` correct here?

README.rst
nepython: oops

README.rst
nepython: That was referring to the **uptime** shown by Device status tab(data). While second one refers to the Uptime chart. I have added pics for both.

README.rst
```diff
@@ -13,6 +13,9 @@ OpenWISP 2 monitoring module (Work in progress).
 
 ------------
 
+.. figure:: docs/monitoring.gif
```
nemesisdesign, line 16: use absolute paths please, or it won't work on pypi

README.rst
```diff
@@ -22,13 +25,13 @@ OpenWISP 2 monitoring module (Work in progress).
 Available Features
 ------------------
 
-* Displays ``Uptime``, ``Memory Status``, ``Load Status``, ``Addresses``, ``Neighbours`` of the **Device**
-* Displays ``Monitoring Graphs`` namely ``Uptime``, ``Packet Loss``, ``Round Trip Time``, ``Traffic`` in the form of lively charts
+* Displays `Uptime, Memory Status, Load Status, Addresses, Neighbours of the Device <docs/device_status.png>`_
```
nemesisdesign, line 28: absolute paths

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|configuration |                  ``traffic``                    |
++--------------+-------------------------------------------------+
+|    charts    |                 [``traffic``]                   |
++--------------+-------------------------------------------------+
+
+.. figure:: docs/traffic.png
+  :align: center
+
+``Clients Metric``
```
nemesisdesign, line 516: WiFI Clients

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
++--------------+----------------------------------------------------------+
+|related_fields|[``total_memory``, ``buffered_memory``, ``shared_memory``]|
++--------------+----------------------------------------------------------+
+|    charts    |                 [``memory``]                             |
++--------------+----------------------------------------------------------+
+
+.. figure:: docs/memory.png
+  :align: center
+
+``Load Metric``
```
nemesisdesign, line 555: CPU Load

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
+|  field_name  |                   ``load_1``                    |
++--------------+-------------------------------------------------+
+|configuration |                   ``load``                      |
++--------------+-------------------------------------------------+
+|related_fields|           [``load_5``, ``load_15``]             |
++--------------+-------------------------------------------------+
+|    charts    |                 [``load``]                      |
++--------------+-------------------------------------------------+
+
+``Disk Metric``
```
nemesisdesign, line 574: Disk Usage

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
+|  value type  |                   ``int``                       |
++--------------+-------------------------------------------------+
+|  field_name  |                  ``clients``                    |
++--------------+-------------------------------------------------+
+|configuration |                  ``clients``                    |
++--------------+-------------------------------------------------+
+|    charts    |               [``wifi_clients``]                |
++--------------+-------------------------------------------------+
+
+``Memory Metric``
```
nemesisdesign, line 533: Memory Usage

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
+.. figure:: docs/uptime.png
+  :align: center
+
+.. figure:: docs/downtime.png
+  :align: center
+
+.. figure:: docs/rtt.png
+  :align: center
+
+``Traffic Metric``
```
nemesisdesign, line 496: Traffic

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
 The signal is emitted just before a successful response is returned,
 it is not sent if the response was not successful.
 
+Default Metrics
+---------------
+
+``Ping Metric``
```
nemesisdesign, line 468: Ping

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
+|  field_name  |                 ``reachable``                   |
++--------------+-------------------------------------------------+
+|configuration |                   ``ping``                      |
++--------------+-------------------------------------------------+
+|related_fields|[``loss``, ``rtt_min``, ``rtt_max``, ``rtt_avg``]|
++--------------+-------------------------------------------------+
+|    charts    |      [``uptime``, ``packet_loss``, ``rtt``]     |
++--------------+-------------------------------------------------+
+
+.. figure:: docs/uptime.png
```
nemesisdesign, line 487: abs path, apply to all others as well please

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|     key      |                   ``ping``                      |
++--------------+-------------------------------------------------+
+|  value type  |  ``int`` (reachable / loss), ``float`` (rtt)    |
++--------------+-------------------------------------------------+
+|  field_name  |                 ``reachable``                   |
++--------------+-------------------------------------------------+
+|configuration |                   ``ping``                      |
++--------------+-------------------------------------------------+
+|related_fields|[``loss``, ``rtt_min``, ``rtt_max``, ``rtt_avg``]|
```
nemesisdesign, line 482: I'd change this to:

```
``loss``, ``rtt_min``, ``rtt_max``, ``rtt_avg``
```
applies to other metrics as well

README.rst
```diff
@@ -425,6 +462,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|    charts    |      [``uptime``, ``packet_loss``, ``rtt``]     |
```
nemesisdesign, line 484: I'd change this to:

```
``uptime``, ``packet_loss``, ``rtt``
```

applies to other metrics as well

README.rst
PabloCastellano: Memory and Load Status is not very meaningful. Please reword, e.g. memory usage and CPU load and mention `disk usage` too

README.rst
PabloCastellano: For someone who reads it for the first time it's confusing, but we can reword it later

README.rst
nepython: Currently, the `Device Status` is not displaying `disk_usage`, it can be changed easily but do we need to change it (might be redundant since the graph is already doing it.)

nepython:
### Up for Discussion
I had a suggestion that shall we store the `gif` in a external location (gdrive or any other platform) and use the link. The reason being the size of gif is around `6 MB` and that would unnecessarily increase repo size along with it. Not a big issue but currently repo size is in KBs and it would be better to keep it that way, ideal when cloning/ forking in my opinion :smile:

nepython:
> Let's proceed to store the gif outside of the repo.

I queried on gitter some time back, where exactly :sweat_smile:
I prefer gdrive (don't know if OpenWISP has a Google account)

README.rst
```diff
@@ -425,6 +428,135 @@ view (only when using HTTP POST).
...
+|  value type  |                   ``float``                     |
++--------------+-------------------------------------------------+
+|  field_name  |                 ``used_disk``                   |
++--------------+-------------------------------------------------+
+|configuration |                   ``disk``                      |
++--------------+-------------------------------------------------+
+|    charts    |                   ``disk``                      |
++--------------+-------------------------------------------------+
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/master/docs/disk_usage.png
```
nemesisdesign, line 557: Does this work? https://drive.google.com/file/d/1BgxmXEkdEgCWlWASALZB2a3WeZad0wXw/view?usp=sharing

nepython: The link needs to be edited slightly but it works, added the commit.

README.rst
```diff
@@ -425,6 +428,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|     key      |                   ``ping``                      |
++--------------+-------------------------------------------------+
+|  value type  |  ``int`` (reachable / loss), ``float`` (rtt)    |
++--------------+-------------------------------------------------+
+|  field_name  |                 ``reachable``                   |
++--------------+-------------------------------------------------+
+|configuration |                   ``ping``                      |
++--------------+-------------------------------------------------+
+|related_fields| ``loss``, ``rtt_min``, ``rtt_max``, ``rtt_avg`` |
```
PabloCastellano, line 448: I'd merge `field_name` and `related_fields` into just `fields` in all tables, otherwise it sounds too technical and tied to the current implementation.

Keep in mind that the idea is that some newcomer can understand at first sight what are the metrics stored and what fields are available for them.

README.rst
```diff
@@ -425,6 +428,135 @@ view (only when using HTTP POST).
...
 it is not sent if the response was not successful.
 
+Default Metrics
+---------------
+
+``Ping``
+~~~~~~~~
+
++--------------+-------------------------------------------------+
+|     name     |              ``Ping <Device>``                  |
```
PabloCastellano, line 438: I think the name is not very useful to be shown here, however the description would be much more helpful.

README.rst
```diff
@@ -425,6 +428,135 @@ view (only when using HTTP POST).
...
++--------------+-------------------------------------------------+
+|     key      |                   ``ping``                      |
```
PabloCastellano, line 440: What about saying "Measurement" instead of just key? Key doesn't tell me anything at all. Measurement is more tied to influxdb but we can find a better name when we finish abstracting the metrics from the db

nepython: Well, I read that it is simply called `metric` in the case of `Prometheus` and several other timeseries db. I think giving this field the name `measurement/metric` might confuse new users unless they are acquainted with influxdb and how metric is acting as a middle agent here (less likely, I think). Suppose they are creating a new metric they will find the field `key` to fill and not `measurement` (increases confusion in my opinion but you might know better).

README.rst
nepython: Agreed, so I think it is safe to remove the `name`  from this and all other metric' s table as well hence will proceed with it.

README.rst
nepython: Makes sense, it is better for a newcomer but might add slight confusion for a contributor who might not find similarity between the model and :point_up: the table. How's this, merge `field_name` and `related_fields` calling it simply `fields` but adding a small note at the end of this section mentioning the same. It should work for both, I think :)

README.rst
PabloCastellano: IMO as a end-user or potential customer of openwisp-monitoring I expect the README to describe the metrics that I will get for free when using this module, rather than technical details of the Metric model, which is not useful to evaluate if I want to use the module or not and after all is going to change when we continue refactoring it

README.rst
PabloCastellano: Yes but we are probably going to change the current implementation, and the contributor can check out and study the code at any time.

README.rst
```diff
@@ -432,6 +435,117 @@ view (only when using HTTP POST).
...
+
+``CPU Load``
+~~~~~~~~~~~~
+
++--------------------+-------------------------------------------------+
+|  **measurement**:  |                   ``load``                      |
++--------------------+-------------------------------------------------+
+|  **value type**:   |                   ``float``                     |
++--------------------+-------------------------------------------------+
+|    **fields**:     |   ``cpu``, ``load_1``, ``load_5``, ``load_15``  |
```
PabloCastellano, line 524: cpu -> cpu_usage

README.rst
```diff
@@ -432,6 +435,117 @@ view (only when using HTTP POST).
...
+
+``Traffic``
+~~~~~~~~~~~
+
++--------------------+-------------------------------------------------+
+|  **measurement**:  |             ``<interface_name>``                |
++--------------------+-------------------------------------------------+
+|  **value type**:   |                    ``int``                      |
++--------------------+-------------------------------------------------+
+|    **fields**:     |          ``<stat>`` (rx_bytes / tx_bytes)       |
```
PabloCastellano, line 473: I think you can remove `<stat>`

nepython: I have separated the two traffic metrics info (which is the case currently). We can change this once these two metrics get merged.

PR #96, opened 2020-05-25
[fix] Sometimes mac_address key is not present in the dictionary

PabloCastellano:
An extract of what `ip neigh -json` returns in my device:

```
  ...
  {
    "dst": "fddc:c8c3:5b6c:0:8ca3:2665:625d:5921",
    "dev": "br-lan",
    "lladdr": "18:5e:0f:12:12:0e",
    "state": [
      "STALE"
    ]
  },
  {
    "dst": "2001:db80::1",
    "dev": "eth0.3",
    "state": [
      "FAILED"
    ]
  },
  ...
```

### DIFF
--- a/openwisp_monitoring/device/models.py
+++ b/openwisp_monitoring/device/models.py
@@ -151,8 +151,11 @@ class DeviceData(Device):
                 client['vendor'] = self._mac_lookup(client['mac'])
         for neighbor in self.data.get('neighbors', []):
             # Neighbors with state = FAILED have no mac_address
-            mac_address = neighbor.get('mac_address', '')
-            neighbor['vendor'] = self._mac_lookup(mac_address)
+            try:
+                mac_address = neighbor['mac_address']
+                neighbor['vendor'] = self._mac_lookup(mac_address)
+            except KeyError:
+                neighbor['vendor'] = ''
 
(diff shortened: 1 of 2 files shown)


openwisp_monitoring/device/models.py
```diff
@@ -150,12 +150,14 @@ def add_mac_vendor_info(self):
             for client in interface['wireless']['clients']:
                 client['vendor'] = self._mac_lookup(client['mac'])
         for neighbor in self.data.get('neighbors', []):
-            neighbor['vendor'] = self._mac_lookup(neighbor['mac_address'])
+            # Neighbors with state = FAILED have no mac_address
+            mac_address = neighbor.get('mac_address', '')
+            neighbor['vendor'] = self._mac_lookup(mac_address)
```
nemesisdesign, line 155: if the mac address is empty I would avoid the lookup altogether because it has some cost from a performance point of view, the table with all the mac address prefixes is large

PabloCastellano:
Made the requested changes.

I've found that running `dd.save_data()` after the mac vendor information has been added, results in a schema validation because we don't expect the `vendor` item. @nemesisdesign Do you think we should update the schema too even if we don't expect the vendor to be sent by the monitoring script?

PR #50, opened 2020-04-24
[enhancement] Implement Swappable models

nepython:
closes #45
This is still WIP and I have implemented it only across `monitoring` app. There were few queries like `proxy` model and `abstract` model not compatible with each other which would be needed to create `AbstactDeviceData`.
**NOTE**: In the `AbstractMetric` model, I was unable to define `full_clean` method as it would lead to recursive calls when models are being loaded and thus defined it in the `Metric` model.

PR #102, opened 2020-05-27
[feature] Swappable models (GSOC20)

nemesisdesign:
I'll be noting here the improvements to make on the work done in PR #50.

I'm merging your PR plus my improvements on a branch, let's keep working on that branch.

### ``sample_device_monitoring.DetailsModel``

This is not very useful. We're testing that it is possible to add additional models?
That's not the point, the point is to extend the behavior of the existing ones.

I would add a "details"  field on `DeviceMonitoring`, have it provide a default value and override the __str__ method to include it the content of this field in its output, and update the tests accordingly.

### ``sample_check.monitoring`` DetailsModel misunderstanding 

It should not be a concrete model, it should be a base abstract model used as a base for the other models, see here:

https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/tests/openwisp2/sample_firmware_upgrader/models.py#L12-L46

The point is to test extending the base app, adding fields, testing the differences, please do that, otherwise all this work is dead code, useless, we do not want to work on useless code right?

### ``sample_check.check`` missing details field

As above, the Check model of the sample app should have an additional field which is used somehow in the code.

### README

Update README regarding sample_device_monitoring and any other change that I introduced.
Look for my `TODO` notes in the README.

### Management commands

These shall be implemented in the sample apps too, see openwisp-network-topology/django-netjsongraph for an example, ask @atb00ker if you need help.

I leave the failing tests for you to fix.

Do not skip tests like you've been doing in #50, that's called taking shortcuts and it's bad.

### run-qa-checks

This must do the checks for the sample app as well, see openwisp-firmware-upgrader for reference.

### Clean up commit history

- Rebase on the current master
- Fix conflicts
- squash your commits from PR #50, keep my commit intact, add your changes afterwards


nepython:
@nemesisdesign, thanks for the pointers and taking time for reviewing / improving the PR.
I have updated everything as per your suggestions though please note the following:
- Few tests are still not running in the SAMPLE_APP (not skipping). 8 tests aren't running since sample_app shows 137(in which two have been added) tests got executed while total is 143. I am locating which ones and why so.
- I have updated the README but am not very sure if it's good for someone who hasn't used django earlier. Please review it and suggest any improvements if needed.
- Shall we skip one test `test_wifi_hostapd` in sample_app because it is a bit flaky and running it twice also doubles chances of failure (just a suggestion, you might know best)

nemesisdesign:
> @nemesisdesign, thanks for the pointers and taking time for reviewing / improving the PR.
> I have updated everything as per your suggestions though please note the following:
> 
> * [Solved] Few tests are still not running in the SAMPLE_APP (not skipping). 11 tests aren't running since sample_app shows 137(in which two have been added) tests got executed while total is 146. I am locating which ones and why so.

ok, we should be sure all the tests are run

>   --> notifications app's test weren't running. **Query**: Should notifications app's tests also be run again with sample_app?

Yes it's a good way to catch protential integration issues.

> * I have updated the `README` but am not very sure if it's good for someone who hasn't used django earlier. Please review it and suggest any improvements if needed.

Will check asap.

> * Shall we skip one test `test_wifi_hostapd` in sample_app because it is a bit flaky and running it twice also doubles chances of failure (just a suggestion, you might know best)

Explained in chat why we shouldn't skip it.

> * I didn't add a `details` field in `sample_check` since `check` app already contains `name`, `description` and has `__str__` method overiding the default. Adding `details` field might be redundant. Hence, I introduced a new field called `last_called` which stores the datetime of when was perform_check method last called. Added test for the same.

Sounds good, will review asap.

nepython:
> Yes it's a good way to catch protential integration issues.

Done :smile:

tests/openwisp2/sample_check/models.py
```diff
@@ -1,8 +1,21 @@
...
         swappable = swappable_setting('check', 'Check')
+
+    def perform_check(self, store=True):
+        """
+        initiates check instance and calls its check method
+        """
+        self.last_called = now()
+        self.full_clean()
+        self.save()
+        return self.check_instance.check(store=True)
```
nemesisdesign, line 21: I think this last line can be replaced by:

```python
return super().perform_check(store=store)
```

Which is more future proof.

tests/openwisp2/sample_monitoring/models.py
```diff
@@ -0,0 +1,28 @@
+from django.db import models
+from openwisp_monitoring.monitoring.base.models import (
+    AbstractGraph,
+    AbstractMetric,
+    AbstractThreshold,
+)
+
+
+class DetailsModel(models.Model):
+    details = models.CharField(max_length=64, blank=True, null=True)
```
nemesisdesign, line 10: good, but please make this field do something and add a test for each model

tests/openwisp2/sample_check/models.py
nepython: @nemesisdesign, can you please explain as to how exactly is it `Future Proof` so I can take note :smiley:

nemesisdesign: Found you.

I meant calling `super()` instead of duplicating `self.check_instance.check(store=True)` , because so if the parent method is refactored or changed this method here won't need to be changed.

PR #103, opened 2020-05-28
[enhancement] Abstract code communicating with the time series DB #65

### LINKED ISSUE #65: [enhancement] Abstract code communicating with the time series DB
Abstract the code that communicates with the timeseries DB so that new timeseries backends can be written to support other timeseries DBs.

* All the operations that are done with the timeseries DB are to be abstracted in a class, which we'll call timeseries DB backend.
* All queries and lines of code scattered around openwisp-monitoring which produce requests to influxdb have to be centralized in a class called timeseries DB backend.
* There must be a setting which allows configuring which backend to use, like django does with the SQL databases.

nepython:
* I have added few **TODO** comments as reminder notes when adding support for another timeseries db
* Updated the docs on how to select / declare a timeseries database in `settings.py`
* Introduced Deprecation warning to shift to the latest format of defining the timeseries db

Closes #65


openwisp_monitoring/monitoring/tests/test_db_creation.py
```diff
@@ -6,23 +6,25 @@
...
     @patch('influxdb.client.InfluxDBClient.create_database', mock_create_database)
     def test_check_retry(self, mock):
+        TIMESERIES_DB['NAME'] = 'test_db'
         try:
             apps.get_app_config('monitoring').create_database()
         except ConnectionError:
             pass
-        get_db().drop_database('test_db')
+        TimeseriesDB.get_db().drop_database('test_db')
+        TIMESERIES_DB['NAME'] = 'openwisp2'
```
PabloCastellano, line 27: Why is it necessary to change the name of the db twice if previously it wasn't?

Can you keep and adapt `@patch('openwisp_monitoring.monitoring.settings.INFLUXDB_DATABASE', 'test_db')` instead?

openwisp_monitoring/monitoring/models.py
```diff
@@ -500,9 +486,9 @@ def read(
             else:
                 data_query = self.get_query(**query_kwargs)
                 summary_query = self.get_query(summary=True, **query_kwargs)
-            points = list(query(data_query, epoch='s').get_points())
-            summary = list(query(summary_query, epoch='s').get_points())
-        except InfluxDBClientError as e:
+            points = list(TimeseriesDB.query(data_query, epoch='s').get_points())
+            summary = list(TimeseriesDB.query(summary_query, epoch='s').get_points())
+        except TIMESERIES_DB_EXCEPTION as e:
```
PabloCastellano, line 491: Please use CamelCase for the exception name and let's keep it less generic. What about TimeseriesDbClientError?

openwisp_monitoring/monitoring/exceptions.py
```diff
@@ -1,2 +1,14 @@
...
+
 class InvalidChartConfigException(Exception):
     pass
+
+
+database_exception = {
+    'influxdb': InfluxDBClientError,
+}
+
+TIMESERIES_DB_EXCEPTION = database_exception[TIMESERIES_DB['DATABASE']]
```
PabloCastellano, line 14: This is not the way to define Exceptions in python. This is just a string.

openwisp_monitoring/monitoring/utils.py
```diff
@@ -3,60 +3,93 @@
...
+        items = list(response.get_points('databases'))
+        databases = [database['name'] for database in items]
+        # if database does not exists, create it
+        if TIMESERIES_DB['NAME'] not in databases:
+            db.create_database(TIMESERIES_DB['NAME'])
+            logger.info(f'Created influxdb database {TIMESERIES_DB["NAME"]}')
+
+
+if TIMESERIES_DB['DATABASE'] == 'influxdb':
+    TimeseriesDB = InfluxDB()
```
PabloCastellano, line 95: I think it's better if we have a generic TimeseriesDB class with a `db_type` attribute. This way you don't need to check `TIMESERIES_DB['DATABASE']` every time and you can just call `TimeseriesDB.method()`, which will check the database type and send the request

tests/openwisp2/charts_query.py
```diff
@@ -0,0 +1,31 @@
+query = {
```
PabloCastellano, line 1: Why is this file inside tests/openwisp2?

tests/openwisp2/settings.py
```diff
@@ -178,13 +183,7 @@
...
         'top_fields': 2,
         'order': 999,
-        'query': {
-            'influxdb': (
-                "SELECT {fields|SUM|/ 1} FROM {key} "
-                "WHERE time >= '{time}' AND content_type = "
-                "'{content_type}' AND object_id = '{object_id}'"
-            )
-        },
+        'query': query[TIMESERIES_DB['DATABASE']]['histogram'],
```
PabloCastellano, line 186: As the above comment says, this is used ofr automated testing purposes. If we support different DBs we need to test them all, keep this is mind.

openwisp_monitoring/monitoring/exceptions.py
nepython: @PabloCastellano, I have changed it and I think it's better now but can you please enlighten me on why it's a string (I am not aware).

nepython:
> Why is this file inside tests/openwisp2?

It was earlier in **tests/openwisp2/settings.py**, I have just migrated the code to a new file and stored it in a dict which should make it easier to be retrieved. It seemed right to keep the dictionary object in a separate file and not in the settings file since it's primarily not a setting. If you find it right to move it back or store it somewhere else, please mention it and I shall do the same.
> As the above comment says, this is used ofr automated testing purposes. If we support different DBs we need to test them all, keep this is mind.

I was thinking of adding the queries of the new timeseries_db in the above mentioned `dict` so we only need to change the **TIMESERIES_DB['DATABASE']** and everything switches automatically, that was the idea. Besides that, I am unsure if you were trying to point me towards anything else. If yes,I didn't get it so can you please explain it a bit briefly.

README.rst
```diff
@@ -89,9 +89,12 @@ Follow the setup instructions of `openwisp-controller
     ]
 
     # Make sure you change them in production
-    INFLUXDB_USER = 'openwisp'
-    INFLUXDB_PASSWORD = 'openwisp'
-    INFLUXDB_DATABASE = 'openwisp2'
+    TIMESERIES_DB = {
+        'DATABASE': 'influxdb',
```
nemesisdesign, line 93: I think this should be a path to a class, which will have to be imported dinamically.
That way we allow contributors to write their customized backends more easily.
(please document how to do this, with a short section that mentions this).

I also think the dictionary key should be named `BACKEND`

README.rst
```diff
@@ -89,9 +89,12 @@ Follow the setup instructions of `openwisp-controller
     ]
 
     # Make sure you change them in production
-    INFLUXDB_USER = 'openwisp'
-    INFLUXDB_PASSWORD = 'openwisp'
-    INFLUXDB_DATABASE = 'openwisp2'
+    TIMESERIES_DB = {
```
nemesisdesign, line 92: I would name it `TIMESERIES_DATABASE` because as per the zen of python, explicit is better than implicit, you may shorten it to `TIMESERIES_DB` when importing, eg: `from ... import TIMESERIES_DATABASE as TIMESERIES_DB`, although unless it's used a lot in the code the explicit form will help readability

openwisp_monitoring/monitoring/utils.py
```diff
@@ -3,60 +3,96 @@
...
-    """ creates database if necessary """
-    db = get_db()
-    response = db.query('SHOW DATABASES')
-    items = list(response.get_points('databases'))
-    databases = [database['name'] for database in items]
-    # if database does not exists, create it
-    if settings.INFLUXDB_DATABASE not in databases:
-        db.create_database(settings.INFLUXDB_DATABASE)
-        logger.info(f'Created influxdb database {settings.INFLUXDB_DATABASE}')
+class TimeseriesDB:
```
nemesisdesign, line 11: we need to have one class for each database backend

tests/openwisp2/charts_query.py
```diff
@@ -0,0 +1,31 @@
...
+    'influxdb': {
```
nemesisdesign, line 2: I don't think moving this information in a separate file is useful at all

tests/openwisp2/settings.py
nemesisdesign: @PabloCastellano you're right, one of the measurable outcomes is to run the test with both timeseries DB

tests/openwisp2/settings.py
```diff
@@ -200,73 +199,47 @@
...
         'description': 'Default query for testing purposes',
         'unit': 'n.',
         'order': 999,
-        'query': {
-            'influxdb': (
-                "SELECT {field_name} FROM {key} WHERE time >= '{time}' AND "
-                "content_type = '{content_type}' AND object_id = '{object_id}'"
-            )
-        },
+        'query': query[TIMESERIES_DB['DATABASE']]['default'],
```
nemesisdesign, line 210: What's the advantage of storing the queries for the default charts in a separate file?
I think these changes are not useful, if I'm missing something let me know.

openwisp_monitoring/device/models.py
```diff
@@ -98,11 +99,13 @@ def data(self):
...
-            "ORDER BY time DESC LIMIT 1".format(SHORT_RP, self.__key, self.pk)
-        )
-        points = list(query(q).get_points())
+        if TIMESERIES_DB['DATABASE'] == 'influxdb':
+            q = (
+                "SELECT data FROM {0}.{1} WHERE pk = '{2}' "
+                "ORDER BY time DESC LIMIT 1".format(SHORT_RP, self.__key, self.pk)
+            )
+        # TODO: Might be different for the other database
+        points = list(TimeseriesDB().query(q).get_points())
```
nemesisdesign, line 108: the model doesn't have to know anything about influxdb or prometheus, it just have to call a method on the timeseries db backend to do what it needs to do, the differences for each timeseries DB are handled in its respective backend, which for the moment for us will be influxdb only.

openwisp_monitoring/monitoring/exceptions.py
PabloCastellano: OH you are rgiht, I didn't pay enough attention.

openwisp_monitoring/monitoring/utils.py
PabloCastellano: @nemesisdesign Makes sense.

@nepython forget my previous comment about having a db_type attribute in the class

openwisp_monitoring/db/backends/influxdb/queries.py
```diff
@@ -0,0 +1,68 @@
...
+default_chart_query = [
+    "SELECT {field_name} FROM {key} WHERE time >= '{time}'",
+    " AND content_type = '{content_type}' AND object_id = '{object_id}'",
+]
+
+device_data_query = (
+    "SELECT data FROM {0}.{1} WHERE pk = '{2}' " "ORDER BY time DESC LIMIT 1"
+)
+
+test_query = {
```
nemesisdesign, line 40: test code should not be mixed with the actual implementation

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -475,7 +455,7 @@ def _get_top_fields(self, number, time=DEFAULT_TIME, timezone=settings.TIME_ZONE
         q = self.get_query(
             query=q, summary=True, fields=['SUM(*)'], time=time, timezone=timezone
```
nemesisdesign, line 456: what about SUM?

tests/openwisp2/settings.py
```diff
@@ -16,6 +17,15 @@
     }
 }
 
+TIMESERIES_DATABASE = {
+    'BACKEND': 'openwisp_monitoring.db.backends.influxdb',
+    'USER': 'openwisp',
+    'PASSWORD': 'openwisp',
+    'NAME': 'openwisp2',
+    'HOST': 'localhost',
+    'PORT': '8086',
```
nemesisdesign, line 26: for backward compatibility, please make sure the application works also if this is not defined, supply the default values somewhere, but only if they're not defined (and log a warning using the logger, urging to migrate).
We'll then remove this once all instances are upgraded.

tests/openwisp2/settings.py
```diff
@@ -170,6 +176,8 @@
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 CELERY_EMAIL_BACKEND = EMAIL_BACKEND
 
+test_query = importlib.import_module(TIMESERIES_DATABASE["BACKEND"]).queries.test_query
```
nemesisdesign, line 179: I don't think we should be doing this.
In this section, some custom queries used in the tests are defined (we'll have to change this code to use register()), I think the definition of the test queries must stay here.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -171,30 +170,11 @@ def write(self, value, time=None, database=None, check=True, extra_values=None):
...
-            )
-        if conditions:
-            conditions = 'WHERE %s' % ' AND '.join(conditions)
-            q = '{0} {1}'.format(q, conditions)
-        if order:
-            q = '{0} ORDER BY {1}'.format(q, order)
-        if limit:
-            q = '{0} LIMIT {1}'.format(q, limit)
-        return list(query(q, epoch='s').get_points())
+        return TimeseriesDB().read(
```
nemesisdesign, line 175: I don't like that `TimeseriesDB` is being instantiated each time.

Either it needs to be instantiated once and used as a singleton or we should switch to using class methods only.

I think the best path would be to instantiate it only once, and perform some basic checks during instantiation (eg: does the DB exist and can we connect to it? If not fail).

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
...
+    def create_database(self, database=TIMESERIES_DB['NAME']):
+        """ creates database if necessary """
+        db = self.get_db()
+        response = db.query('SHOW DATABASES')
+        items = list(response.get_points('databases'))
+        databases = [database['name'] for database in items]
+        # if database does not exists, create it
+        if database not in databases:
+            db.create_database(database)
+            logger.info(f'Created influxdb database {database}')
```
PabloCastellano, line 21: What happens if we just create the database and it exists? Will it raise an exception? In this case I always prefer to “ask forgiveness not permission” and capture the exception. I feel it's also more pythonic

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def drop_database(self, database=TIMESERIES_DB['NAME']):
+        """ drops database if it exists """
+        db = self.get_db()
+        response = db.query('SHOW DATABASES')
+        items = list(response.get_points('databases'))
+        databases = [database['name'] for database in items]
+        if database in databases:
+            db.drop_database(database)
+            logger.info(f'Dropped influxdb database {database}')
```
PabloCastellano, line 31: Same - forgiveness, not permission

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,77 @@
...
+
+from influxdb import client
+
+from .. import TIMESERIES_DB
+from .creation import DatabaseCreation
+from .exception import DatabaseException
+
+
+class DatabaseClient(DatabaseCreation, DatabaseException):
+    def get_db(self):
```
PabloCastellano, line 11: This will return a new instance everytime we write(), read() or query(). Use the singleton pattern, something like:

```
if not self._db:
    self._db = client.InfluxDBClient(...)
return self._db
```

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
+import logging
+
+from .. import TIMESERIES_DB
+
+logger = logging.getLogger(__name__)
+
+
+class DatabaseCreation(object):
```
PabloCastellano, line 8: Not sure about this class. Have you taken inspiration from somewhere else?

openwisp_monitoring/monitoring/base/models.py
PabloCastellano: +1 to instantiating it only once

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -353,7 +333,7 @@ def unit(self):
     def query(self):
         query = self.config_dict['query']
         if query:
-            return query['influxdb']
+            return query
```
PabloCastellano, line 336: This function can be shortened like:

```
return self.config_dict['query'] or self._default_query
```

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -362,9 +342,9 @@ def top_fields(self):
 
     @property
     def _default_query(self):
-        q = "SELECT {field_name} FROM {key} WHERE time >= '{time}'"
+        q = default_chart_query[0]
         if self.metric.object_id:
-            q += " AND content_type = '{content_type}' AND object_id = '{object_id}'"
+            q += default_chart_query[1]
```
PabloCastellano, line 347: This doesn't look like enough abstraction and it's very tied to the influxdb query yet

tests/openwisp2/settings.py
```diff
@@ -178,13 +186,7 @@
...
         'top_fields': 2,
         'order': 999,
-        'query': {
-            'influxdb': (
-                "SELECT {fields|SUM|/ 1} FROM {key} "
-                "WHERE time >= '{time}' AND content_type = "
-                "'{content_type}' AND object_id = '{object_id}'"
-            )
-        },
+        'query': test_query['histogram'],
```
PabloCastellano, line 189: I would leave the test queries in this file

openwisp_monitoring/db/backends/influxdb/queries.py
nepython: This data structure needs to be retained but since queries of `influxdb` and other Timeseries db shall be different. I am unsure what might be the best place to store it so will put it in openwisp2/settings file currently.

openwisp_monitoring/monitoring/base/models.py
nepython: I checked few Timeseries dB query languages to realize that many things such as the basic operators, functions (includes sum) are common. Hence, didn't make any changes. Let me know if I should.

openwisp_monitoring/db/backends/influxdb/creation.py
nepython: Simply put, don't verify whether the database exists. In case influxdb raises an exception complaining db exists, capture and hide the exception, right? If so, I am not sure how does it help, does it speed up things significantly (unsure here). I will make the changes @PabloCastellano but it would be great if you can explain what's the clear advantage.

openwisp_monitoring/db/backends/influxdb/creation.py
nepython: From how Django does it https://github.com/django/django/blob/master/django/db/backends/sqlite3/creation.py

openwisp_monitoring/monitoring/base/models.py
nepython: Any suggestions :smiley:

openwisp_monitoring/db/backends/influxdb/creation.py
PabloCastellano: [This is a common Python coding style](https://docs.python.org/3.4/glossary.html#:~:text=Easier%20to%20ask%20for%20forgiveness,many%20try%20and%20except%20statements.).

In terms of speed, well, it will save one query (SHOW DATABASES) when the database does not exist and will remain the same when the database exists. The code is also much shorter and hence less prone to errors

openwisp_monitoring/db/backends/influxdb/queries.py
PabloCastellano: You can just move the test_query dictionary back to the settings file since that is the place where it will be useful

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
...
+    ORIGINAL_DB = TIMESERIES_DB['NAME']
+    TEST_DB = f'{ORIGINAL_DB}_test'
+
+    def create_database(self, database=TIMESERIES_DB['NAME']):
+        """ creates database if necessary """
+        db = self.get_db()
+        response = db.query('SHOW DATABASES')
```
PabloCastellano, line 15: FYI, there exists: [client.get_list_database()](https://influxdb-python.readthedocs.io/en/latest/api-documentation.html#influxdb.InfluxDBClient.get_list_database)

openwisp_monitoring/db/backends/influxdb/creation.py
PabloCastellano: If you want to follow the same pattern, you have to subclass `django.db.backends.base.creation.BaseDatabaseCreation`

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
...
+        items = list(response.get_points('databases'))
+        databases = [database['name'] for database in items]
+        if database in databases:
+            db.drop_database(database)
+            logger.info(f'Dropped influxdb database {database}')
+
+    def create_test_database(self):
+        TIMESERIES_DB['NAME'] = self.TEST_DB
```
PabloCastellano, line 34: I'm not convinced about this line. I think [this approach](https://github.com/django/django/blob/master/django/db/backends/sqlite3/creation.py#L24) is much better, it's clearer and doesn't modify settings without the user noticing

openwisp_monitoring/db/backends/influxdb/creation.py
```diff
@@ -0,0 +1,40 @@
...
+        self.create_database(self.TEST_DB)
+
+    def drop_test_database(self):
+        self.drop_database(self.TEST_DB)
+        TIMESERIES_DB['NAME'] = self.ORIGINAL_DB
+        self.get_db().query('DROP SERIES FROM /.*/')
```
PabloCastellano, line 40: This method does many different things instead of just "dropping the test database"

Why don't we just keep drop_database with db_name as parameter?

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -18,21 +17,13 @@
...
+        TimeseriesDB().create_test_database()
 
     @classmethod
     def tearDownClass(cls):
-        get_db().drop_database(cls.TEST_DB)
-        setattr(app_settings, 'INFLUXDB_DATABASE', cls.ORIGINAL_DB)
-
-    def tearDown(self):
-        query('DROP SERIES FROM /.*/')
+        TimeseriesDB().drop_test_database()
```
PabloCastellano, line 26: Why did you merge both tearDownClass and tearDown?

openwisp_monitoring/db/backends/influxdb/creation.py
nepython: from [influxdb-docs](https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#create-database)
>A successful CREATE DATABASE query returns an empty result. If you attempt to create a database that already exists, InfluxDB does nothing and does not return an error.

So, I can safely create database I think without `look before you leap`, `ask for forgiveness not permission` using either coding styles :smile:

openwisp_monitoring/db/backends/influxdb/creation.py
nepython: Same as above :smile:

openwisp_monitoring/db/backends/influxdb/queries.py
nepython: I think for now it is best to retain the original structure in openwisp2/settings since things shall change after `register/unregister chart` feature is added

nepython:
**Summary of latest commit**:
* instantiate only once
* Merged `creation.py` with `client.py` (primarily because there were only two methods)
* introduced `db_name` in `TimeseriesDB` (because as @PabloCastellano, rightly pointed it might not be right to change settings without user consent and I found instantiating only once posed a restriction of changing the db from anywhere in the code so inorder to overcome that short fall too.)
* Shifted back `test_query` to settings though it still exists as an independent structure, I think this can be improved after `chart register/ unregister` feature is introduced.
* introduced **Deprecation warning** to shift to the latest format of defining the `timeseries` db
* some other small improvements like introduced `__all__` in `db/__init__.py`

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,101 @@
...
+                TIMESERIES_DB['USER'],
+                TIMESERIES_DB['PASSWORD'],
+                self.db_name or TIMESERIES_DB['NAME'],
+            )
+            self.db_name = self._db._database
+        return self._db
+
+    def query(self, query, **kwargs):
+        db = self.get_db()
+        database = kwargs.get('database') or self.db_name or TIMESERIES_DB['NAME']
```
PabloCastellano, line 47: I see it too redundant, I like allowing the database parameter kwargs but `or self.db_name or TIMESERIES_DB['NAME']` seems excessive to me.

I think it's safe to remove the last `or TIMESERIES_DB['NAME']`, since self.db_name will always be defined at that point, right?

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,101 @@
...
+        self.get_db().write(
+            {'points': [point]},
+            {
+                'db': kwargs.get('database') or self.db_name or TIMESERIES_DB['NAME'],
+                'rp': kwargs.get('retention_policy'),
+            },
+        )
+
+    def read(self, key, fields, tags, **kwargs):
+        """ Method to be called via threading module. """
```
PabloCastellano, line 77: Can you explain what's up with the threading module?

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,101 @@
...
+        """ drops database if it exists """
+        db = self.get_db()
+        # influxdb does not raise an error if databese does not exist
+        db.drop_database(database)
+        logger.info(f'Dropped influxdb database "{database}"')
+
+    # TODO: Needs to be improved
+    def get_db(self):
+        """ Returns an ``InfluxDBClient`` instance """
+        if not self._db or self._db._database != self.db_name:
```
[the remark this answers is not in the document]
nepython: `self._db` was introduced to store InfluxDBClient as it's costly to make connection to InfluxDB every time as you correctly pointed it out. We might sometimes need a different InfluxDBClient instance (if we need to connect to a different db), thus we need to ignore existing instance and create a new connection in those cases.

Use case `self._db` is stored with name `openwisp2` when apps are being loaded but during tests we need might need a different InfluxDB client (different database) thus ignore existing instance and make a new connection.

nepython:
Need to add `create_retention_policy` in `DatabaseClient`. Refer to https://github.com/openwisp/openwisp-monitoring/pull/123/files#diff-4a3f1da11342c8c7815898ceaaac707fR65 :smile:

openwisp_monitoring/db/backends/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+logger = logging.getLogger(__name__)
+
+TIMESERIES_DB = getattr(settings, 'TIMESERIES_DATABASE', None)
+if not TIMESERIES_DB:
+    TIMESERIES_DB = {
+        'BACKEND': 'openwisp_monitoring.db.backends.influxdb',
+        'USER': 'openwisp',
+        'PASSWORD': 'openwisp',
+        'NAME': 'openwisp2',
+        'HOST': 'localhost',
```
nemesisdesign, line 17: please let's keep backward compatibility with the settings used before, this will facilitate the upgrade.

Eg:

```python
{
# ...
    'USER': getattr(settings, 'INFLUXDB_USER', 'openwisp')
# ...
}
```

openwisp_monitoring/db/backends/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        'PORT': '8086',
+    }
+    logger.warning(
+        'The previous method to define Timeseries Database has been deprecated. Please refer the docs\n'
```
nemesisdesign, line 21: I think it should be `Please refer to the docs:\n`

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,101 @@
...
+
+from influxdb import client
+
+from .. import TIMESERIES_DB
+from .exception import DatabaseException
+
+logger = logging.getLogger(__name__)
+
+
+class DatabaseClient(DatabaseException):
```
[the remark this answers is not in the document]
nemesisdesign: This is almost surely wrong, I've never seen an exception being subclassed to implement features.

Subclass `object` if there isn't anything better to subclass. Keep exceptions for errors and exceptional cases.

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, db_name=None):
+        self._db = None
+        self.db_name = db_name or TIMESERIES_DB['NAME']
+
+    def create_database(self):
+        """ creates database if necessary """
+        db = self.get_db()
+        # InfluxDB does not create a new database, neither raise an error if databese exists
+        db.create_database(self.db_name)
+        logger.info(f'Created InfluxDB database "{self.db_name}"')
```
nemesisdesign, line 22: I think the `debug` level should be used here.

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def drop_database(self):
+        """ drops database if it exists """
+        db = self.get_db()
+        # InfluxDB does not raise an error if databese does not exist
+        db.drop_database(self.db_name)
+        logger.info(f'Dropped InfluxDB database "{self.db_name}"')
```
nemesisdesign, line 29: I think the `debug` level should be used here.

openwisp_monitoring/db/backends/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+            f'openwisp_monitoring.db.backends.{b}' for b in builtin_backends
+        ]:
+            raise ImproperlyConfigured(
+                f"{backend_name} isn't an available database backend.\n"
+                "Try using 'openwisp_monitoring.db.backends.XXX', where XXX is one of:\n"
+                f"{builtin_backends}"
+            ) from e
+
+
+TimeseriesDB = load_backend(module='client').DatabaseClient()
```
nemesisdesign, line 62: `TimeseriesDB` should not be camel case because the convention is that camel case is used for classes.

This should be `timeseries_db`.

openwisp_monitoring/db/backends/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+TimeseriesDBQueries = load_backend(module='queries')
```
nemesisdesign, line 63: Why isn't this just a property of `timeseries_db`, eg: `timeseries_db.queries`?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -451,7 +428,7 @@ def _get_top_fields(self, number, time=DEFAULT_TIME, timezone=settings.TIME_ZONE
         q = self.get_query(
             query=q, summary=True, fields=['SUM(*)'], time=time, timezone=timezone
         )
-        res = list(query(q, epoch='s').get_points())
+        res = list(TimeseriesDB.query(q, epoch='s').get_points())
```
nemesisdesign, line 431: `epoch='s').get_points()` it's influxdb specific.

get_points is a method of the influxdb client library, and is being called from models, hence models is still aware of inflxudb specific code, so the abstraction is not there yet. Abstraction means that `models.py` must not know anything about influxdb It must only call one method of the timeseries DB class, it's this class that will know what to do with the call.

Eg:

```python
timeseries_db.get(q)
```

Applies to all other occurrences, please fix this.

openwisp_monitoring/monitoring/base/models.py
nemesisdesign: same logic applies, it's timeseries DB related code which must be delegated to the timeseries DB backend entirely, it should not be left lying around in here.

Let's find a way to do that.

nepython:
All major InfluxDB related code has been abstracted. Things that yet aren't abstracted:
- Few methods / fields are yet having InfluxDB specific names (eg. tags, fields, retention_policy, etc.). I have renamed most methods to neutral terms but these yet remain as I couldn't think of any neutral term for them yet.
- There might be two or three tests which are slightly InfluxDB specific (I couldn't shift them, maybe we can add subTests while adding the other timeseries. You can find TODO comments in such tests). I have shifted and added all other timeseries db related tests in another file.
- I have noted that the most important point in the abstraction layer might be the return value. The object returned by both the timeseries db should be identical (this might be though left best for now and we can optimize it when adding the other timeseries db).

openwisp_monitoring/db/backends/influxdb/client.py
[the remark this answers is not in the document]
nemesisdesign: @nepython can you use `cached_property` here? https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.functional.cached_property

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -306,7 +306,8 @@ def test_init(self):
         self.assertEqual(dd.data, self._sample_data)
 
     def test_retention_policy(self):
-        rp = get_db().get_list_retention_policies()
+        rp = timeseries_db.get_list_retention_policies()
+        # TODO: Test is InfuxDB specific currently
```
nemesisdesign, line 310: move all the influxdb specific tests to a new test file in `/openwisp_monitoring/db/backends/influxdb/tests.py`

openwisp_monitoring/monitoring/base/__init__.py
```diff
@@ -0,0 +1,9 @@
+GROUP_MAP = {
+    '1d': '10m',
+    '3d': '20m',
+    '7d': '1h',
+    '30d': '24h',
+    '365d': '24h',
+}
+
+DEFAULT_TIME = '7d'
```
nemesisdesign, line 9: wait a moment, this is not influxdb specific, I think it shouldn't me moved out

openwisp_monitoring/monitoring/base/models.py
nemesisdesign: looks better now

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def get_list_retention_policies(self):
+        return self.get_db().get_list_retention_policies()
+
+    def delete_metric_data(self, key=None, tags=None):
+        """
+        deletes a specific metric based on the key and tags
+        provided, you may also choose to delete all metrics
+        """
+        if key == 'all_metrics':
```
nemesisdesign, line 166: let's change this so that if both `key` and `tags` are `None` this part is executed, passing `'all_metrics'` is weird.

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -0,0 +1,286 @@
...
+            self.query('DROP SERIES FROM /.*/')
+        else:
+            self.get_db().delete_series(measurement=key, tags=tags)
+
+    # Chart related functions below
+
+    # TODO: Improve this!
```
nemesisdesign, line 173: improve how? If you can improve it further do so or remove the comment unless we can leave a more specific TODO comment that will be useful in the future.

Issue #108, opened 2020-05-29
[urgent] Checks stopped working 


nemesisdesign:
Active checks (ping) stopped working after #102 has been merged.

This is the output from celery:

```
[2020-05-29 05:05:20,728: INFO/MainProcess] Received task: openwisp_monitoring.check.tasks.run_checks[a616fd68-0593-4271-ad74-4b2e6544273d]  
[2020-05-29 05:05:20,729: DEBUG/MainProcess] TaskPool: Apply <function _fast_trace_task at 0x7f694c1e69e0> (args:('openwisp_monitoring.check.tasks.run_checks', 'a616fd68-0593-4271-ad74-4b2e6544273d', {'lang': 'py', 'task': 'openwisp_monitoring.check.tasks.run_checks', 'id': 'a616fd68-0593-4271-ad74-4b2e6544273d', 'shadow': None, 'eta': None, 'expires': None, 'group': None, 'retries': 0, 'timelimit': [None, None], 'root_id': 'a616fd68-0593-4271-ad74-4b2e6544273d', 'parent_id': None, 'argsrepr': '()', 'kwargsrepr': '{}', 'origin': 'gen3164@envy', 'reply_to': '76a72368-5591-3387-be9e-cb2d26bd5386', 'correlation_id': 'a616fd68-0593-4271-ad74-4b2e6544273d', 'hostname': 'celery@envy', 'delivery_info': {'exchange': '', 'routing_key': 'celery', 'priority': 0, 'redelivered': None}, 'args': [], 'kwargs': {}}, b'[[], {}, {"callbacks": null, "errbacks": null, "chain": null, "chord": null}]', 'application/json', 'utf-8') kwargs:{})
[2020-05-29 05:05:20,730: DEBUG/MainProcess] Task accepted: openwisp_monitoring.check.tasks.run_checks[a616fd68-0593-4271-ad74-4b2e6544273d] pid:3159
[2020-05-29 05:05:20,739: INFO/ForkPoolWorker-6] Task openwisp_monitoring.check.tasks.run_checks[a616fd68-0593-4271-ad74-4b2e6544273d] succeeded in 0.009382204996654764s: None
[2020-05-29 05:07:57,472: ERROR/MainProcess] Received unregistered task of type 'openwisp_monitoring.device.tasks.trigger_device_checks'.
The message has been ignored and discarded.

Did you remember to import the module containing this task?
Or maybe you're using relative imports?

Please see
[...]

PR #113, opened 2020-06-01
[swappable] Further improvements #104 #108

### LINKED ISSUE #104: [swappable] Further improvements after #102
Address the comments and suggestions in the following review:
https://github.com/openwisp/openwisp-monitoring/pull/102#pullrequestreview-420398266

nepython:
Closes #104 #108

I read the docs of celery and found that `celery` auto detects `tasks` of only `INSTALLED_APPS`. Hence, `openwisp_monitoring.device.tasks` was not getting auto-detected but I have explicitly declared it and now it's a registered task with CELERY. I think `openwisp_monitoring.check.tasks` was detected since `run_checks` task was declared for explicitly in `settings` for `celery beat` (I am unsure)

I also changed the test `test_last_called` in `sample_check` to take care of edge cases.


README.rst
```diff
@@ -795,7 +795,25 @@ file in the test project.
...
 
-13. Import the automated tests
+13. Import Celery Tasks
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Add the following in your settings.py to import celery tasks from ``device_monitoring`` app.
+
+.. code-block:: python
+
+    CELERY_IMPORTS = ('openwisp_monitoring.device.tasks')
```
PabloCastellano, line 805: This is a string and a tuple is expected. Either you add a trailing comma or convert it to a list

tests/openwisp2/sample_check/tests.py
```diff
@@ -29,10 +29,9 @@ def test_last_called(self):
             params={'count': 2, 'interval': 10, 'bytes': 10, 'timeout': 50},
         )
         check.perform_check(store=False)
-        # Left the seconds due to minor time lag
-        self.assertEqual(
-            check.last_called.strftime('%D %H %M'), now().strftime('%D %H %M')
-        )
+        # Avoid any failures due to minor lag
+        lag = int(now().strftime('%s')) - int(check.last_called.strftime('%s'))
+        assert lag < 4
```
PabloCastellano, line 34: There exist `self.assertLess` and `self.assertLessEqual`

tests/openwisp2/sample_monitoring/tests.py
```diff
@@ -17,7 +17,20 @@ class TestDatabase(BaseTestDatabase):
 
 
 class TestModels(BaseTestModels):
-    pass
+    def test_metric_details_field(self):
```
PabloCastellano, line 20: Why is this test specific for sample_monitoring?

README.rst
nepython: I had added a `tuple` not sure if it occured due to `black`.

tests/openwisp2/sample_monitoring/tests.py
nepython: I didn't get the query. I have added some additional fields for every model as an example for the user and @nemesisdesign suggested to have a test for testing the same. (You can check the suggestions in #102). Besides that, I am unsure if you implied something else.

PabloCastellano: Fair enough, it just seemed odd that this is just the only test in the sample_app

tests/openwisp2/settings.py
```diff
@@ -167,6 +167,9 @@
     },
 }
 
+# Required for SAMPLE_APP as celery auto detects tasks only from INSTALLED_APPS
+CELERY_IMPORTS = ('openwisp_monitoring.device.tasks',)
```
nemesisdesign, line 171: move this to the `if os.environ.get('SAMPLE_APP', False):` block, otherwise we won't spot problems if we mess the  autoloading of celery tasks

nepython:
>Missing:
>*  #102 (comment)

@nemesisdesign, I actually didn't get the need for it. If I understood it correctly from the docs, we can supply the `test` management command with **(1)** the names of the apps that need to be run one after the other as arguements or **(2)** the root directory containing all the apps will also suffice. In the non-sample apps case we are doing the **(2)** while in the sample-apps case `notifications` app's tests were not getting executed so I explicitly mentioned it **(1)**.

If I am missing or have misunderstood anything, please let me know.

PR #116, opened 2020-06-04
[enhancement] Show cpu, disk_usage and available_memory in device status

### LINKED ISSUE #107: [status] Show number of CPUs and details of disk partitions
Continuation of #76.

Since it's useful information, we should show the number of CPUs and the details of each disk partition in the device status, in human friendly form.
It should be relatively easy to do so since we should already have the data.

PS: show also available memory if present

nepython:
Closes #107


openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -26,30 +26,47 @@ <h2>{% trans 'Device Status' %}</h2>
...
+                <div class="readonly">
+                    {{ device_data.resources.memory.available|filesizeformat }}
+                </div>
+              </div>
+              {% endif %}
               <div class="form-row">
                   <label>{% trans 'Buffered' %}:</label>
                   <div class="readonly">
-                      {{ device_data.resources.memory.buffered }} MB
+                      {{ device_data.resources.memory.buffered|filesizeformat }}
```
PabloCastellano, line 63: :+1:

nepython:
This PR is dependent on #76 for `schema` and `test data`. Coverage should increase after #76 is merged, if I am correct.

openwisp_monitoring/device/base/models.py
```diff
@@ -93,6 +86,19 @@ def data_user_friendly(self):
         # reformat expiry in dhcp leases
         for lease in data.get('dhcp_leases', []):
             lease['expiry'] = datetime.fromtimestamp(lease['expiry'], tz=tz('UTC'))
+        if 'resources' in data and 'disk' in data['resources']:
+            # convert kibibytes to bytes
+            KiB = 1024
```
PabloCastellano, line 91: Since the metric is called used_bytes IMO the conversion should be done in https://github.com/openwisp/openwisp-monitoring/pull/76/ to make sure that we read bytes from influx

nepython: The above change is also not changing the way how the data is stored in influxdb but just changing the way it's to be displayed. I think, I shall remove the above one and rather change the lua script to pass data in bytes, right now there seems a mismatch between the name and the actual data (this is incorrect)

nepython: opened https://github.com/openwisp/lua-monitoring/pull/16 to correct this.

PabloCastellano: Merged

PabloCastellano:
@nepython Can you add a test to check that the rendered admin page contains the new information?

nepython:
> @nepython Can you add a test to check that the rendered admin page contains the new information?

@PabloCastellano, I have added checks that verifies that the three resources are present in the admin page in existing test. It is not a thorough test, I can add a test that shall verify the exact data is generated but I am unsure if that much is needed.

PR #119, opened 2020-06-08
[device] Added notification types for Device events

pandafy:
Related to https://github.com/openwisp/openwisp-notifications/issues/26

Notification Types Added:
- `unreachable_after_upgrade` - Notification specifically for a case like https://github.com/openwisp/openwisp-notifications/issues/26
- `device_up` - Notification when device becomes reachabel
- `device_down` - Generic notification when device is not reachable

When a device goes offline after an upgrade, two notifications are being generated

1. `threshold_crossed`
2. `unreachable_after_upgrade`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -435,13 +435,20 @@ it is not sent if the response was not successful.
 Default Alerts / Notifications
 ------------------------------
 
-+----------------------------+------------------------------------------------------------------------------------------------+
-|    Notification Types      | Uses                                                                                           |
-+----------------------------+------------------------------------------------------------------------------------------------+
-| ``threshold_crossed``      | Fires when a metric crosses the boundary defined in the threshold value of the alert settings. |
-+----------------------------+------------------------------------------------------------------------------------------------+
-| ``threshold_recovery``     | Fires when a metric goes back within the expected range.                                       |
-+----------------------------+------------------------------------------------------------------------------------------------+
++-------------------------------+------------------------------------------------------------------+
+| Notification Type             | Use                                                              |
++-------------------------------+------------------------------------------------------------------+
+| ``threshold_crossed``         | Fires when a metric crosses the boundary defined in the          |
+|                               | threshold value of the alert settings.                           |
++-------------------------------+------------------------------------------------------------------+
+| ``threshold_recovery``        | Fires when a metric goes back within the expected range.         |
++-------------------------------+------------------------------------------------------------------+
+| ``connection_is_working``     | Fires when the connection to a device is working.                |
++-------------------------------+------------------------------------------------------------------+
+| ``connection_is_not_working`` | Fires when the connection (eg: SSH) to a device stops working    |
+|                               | (eg: credentials are outdated, management IP address is          |
+|                               | outdated, or device is not reachable).                           |
++-------------------------------+------------------------------------------------------------------+
 
 Installing for development
 --------------------------
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -4,6 +4,7 @@ from django.db.models.signals import post_delete, post_save
 from django.utils.translation import gettext_lazy as _
 from django_netjsonconfig.signals import checksum_requested
 from openwisp_notifications.signals import notify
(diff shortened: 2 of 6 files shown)


pandafy:
> * memory usage critical
> * memory usage back to normal
> * cpu usage critical
> * cpu usage back to normal
> * disk usage critical
> * disk usage back to normal

I think these are already covered by `under_threshold` and `threshold_crossed`.

pandafy:
> The fact that we get the error messages of the firmware upgrader is nice but I believe this information should be contained in the body of the notificaiton and not in the title. Those messages happen when a connection does not work anymore

I think adding that information in email subject would help user to act more quickly.

README.rst
```diff
@@ -435,13 +435,25 @@ it is not sent if the response was not successful.
...
-+----------------------------+------------------------------------------------------------------------------------------------+
-| ``threshold_recovery``     | Fires when a metric goes back within the expected range.                                       |
-+----------------------------+------------------------------------------------------------------------------------------------+
++-------------------------------+------------------------------------------------------------------+
+| Notification Type             | Use                                                              |
++-------------------------------+------------------------------------------------------------------+
+| ``threshold_crossed``         | Used for creating notification when a metric exceeds             |
+|                               | threshold value.                                                 |
++-------------------------------+------------------------------------------------------------------+
+| ``under_threshold``           | Used for creating notification when a metric falls behind        |
```
nemesisdesign, line 444: this should have been renamed to `threshold_recovery`, please double check

openwisp_monitoring/device/apps.py
```diff
@@ -109,27 +111,79 @@ def is_working_changed_receiver(cls, instance, is_working, **kwargs):
...
         status = 'ok' if is_working else 'problem'
+        notification_opts = dict(sender=instance, target=device)
         if not is_working:
-            # Create a related notification explaining why it's not working
-            desc = instance.failure_reason
-            opts = dict(
-                sender=device, level='warning', verb='not working', description=desc
-            )
+            # Select appropriate notification type for reason of not working
+            if 'upgrade' in instance.failure_reason:
```
nemesisdesign, line 117: This is too fragile. It will stop working and be forgotten if the behavior in openwisp-controller or openwisp-firmware-upgrader changes.

Please let's not do it.

README.rst
```diff
@@ -435,13 +435,25 @@ it is not sent if the response was not successful.
...
+|                               | threshold value.                                                 |
++-------------------------------+------------------------------------------------------------------+
+| ``device_up``                 | Used for creating notification when a device gets connected.     |
++-------------------------------+------------------------------------------------------------------+
+| ``device_down``               | Used for creating notification when a device is unreachable.     |
++-------------------------------+------------------------------------------------------------------+
+| ``unreachable_after_upgrade`` | Used for creating notification when a device becomes unreachable |
+|                               | after firmware upgrade.                                          |
++-------------------------------+------------------------------------------------------------------+
+| ``connection_lost``           | Used for creating notification when the connection to a device   |
```
nemesisdesign, line 454: Please let's have only two notification types:

- `connection_is_not_working`: used to notify when a connection (eg: SSH) to a device stops working (eg: credentials are outdated, management IP address is outdated, or device is not reachable)
- `connection_is_working`: used to notify when a connection to a device goes back to working normally

Include the failure reason in the notification text.

Don't mention the connection is lost, because it may not be lost.

Use "PROBLEM" and "RECOVERY" in email subjects.

openwisp_monitoring/device/apps.py
```diff
@@ -109,27 +111,79 @@ def is_working_changed_receiver(cls, instance, is_working, **kwargs):
...
+        register_notification_type(
+            'device_up',
+            {
+                'name': 'Device Alert',
+                'verb': 'connected successfully',
+                'level': 'info',
+                'email_subject': '[{site.name}] INFO: Device "{notification.target}"',
+                'message': (
+                    '[Device: {notification.target}]({notification.target_link})'
+                    ' has {notification.verb} at {notification.timestamp}.'
```
nemesisdesign, line 145: Is `notification.timestamp` a human readable date? Maybe is better to just remove it.

PR #121, opened 2020-06-10
[charts] Add possibility to export histogram charts #56

### LINKED ISSUE #56: [enhancement] Export histogram chart data
Add a possibility of exporting data of `Histogram` charts.

**QUERY**:
* What is the problem currently?
* Why can't it be done similarly as `scatter` and `bar` charts?

nepython:
Closes #56
Currently it looks like this. I think we can:
* add a column to the left for serial numbers
* Add a row stating `Histogram charts below` or something similar.
Screenshot from 2020-06-10 14-45-34.
In the above image, the histogram chart's name is also `Histogram` (stating to avoid any confusion :sweat_smile: )
**PS**: One more thing I had changed top-fields in histogram configuration to 4.

* I am not sure about this line https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/api/views.py#L112 (in my opinion, histogram charts should not have data['x'] but the chart I have simulated for testing does have it so please confirm if I am correct regarding this.)

### DIFF
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -112,20 +112,34 @@ class DeviceMetricView(GenericAPIView):
 
     def _get_csv(self, data):
         header = ['time']
-        columns = [data['x']]
+        columns = [data.get('x')]
+        histograms = []
         for chart in data['charts']:
-            # TODO: add way to export data for histogram charts
             if chart['type'] == 'histogram':
+                histograms.append(chart)
                 continue
             for trace in chart['traces']:
                 header.append(self._get_csv_header(chart, trace))
                 columns.append(trace[1])
         rows = [header]
-        for index, element in enumerate(data['x']):
+        for index, element in enumerate(data.get('x', [])):
             row = []
             for column in columns:
                 row.append(column[index])
             rows.append(row)
+        for chart in histograms:
+            rows.append([])
+            rows.append([chart['title']])
+            # Export value as 0 if it is None
+            for key, value in chart['summary'].items():
+                if chart['summary'][key] is None:
+                    chart['summary'][key] = 0
+            # Sort Histogram on the basis of value in the descending order
+            sorted_charts = sorted(
+                chart['summary'].items(), key=lambda x: x[1], reverse=True
+            )
+            for field, value in sorted_charts:
+                rows.append([field, value])
         # write CSV to in-memory file object
         fileobj = StringIO()
         csv.writer(fileobj).writerows(rows)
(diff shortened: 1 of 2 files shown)


nepython:
@nemesisdesign, it's already sorted right?
https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/base/models.py#L506
Let me know, if I am missing anything.

nemesisdesign:
> @nemesisdesign, it's already sorted right?
> https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/base/models.py#L506
> Let me know, if I am missing anything.

It's sorted alphabetically which is not helpful in this case. Please address both of the points in my previous comment.

openwisp_monitoring/device/api/views.py
```diff
@@ -113,9 +113,10 @@ def _get_charts_data(self, charts, time, timezone):
     def _get_csv(self, data):
         header = ['time']
         columns = [data['x']]
+        histogram = []
```
PabloCastellano, line 116: nitpick: histogram**s**

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -291,31 +291,49 @@ def test_get_device_metrics_csv(self):
...
+                    '9.73',
+                    '0',
+                    '8.27',
+                ],
+            )
+        with self.subTest('Test export of histogram chart'):
+            self.assertEqual(rows[-4].strip(), '')
+            self.assertEqual(rows[-3].strip(), 'Histogram')
+            self.assertEqual(rows[-2].strip().split(','), ['ssh', '100'])
+            self.assertEqual(rows[-1].strip().split(','), ['http2', '90'])
```
PabloCastellano, line 336: There is no need for subtests in this case, just modify `last_line_before_histogram` as you already did and add the new asserts

openwisp_monitoring/device/api/views.py
```diff
@@ -112,20 +112,34 @@ def _get_charts_data(self, charts, time, timezone):
...
-            # TODO: add way to export data for histogram charts
             if chart['type'] == 'histogram':
+                histograms.append(chart)
                 continue
             for trace in chart['traces']:
                 header.append(self._get_csv_header(chart, trace))
                 columns.append(trace[1])
         rows = [header]
-        for index, element in enumerate(data['x']):
+        for index, element in enumerate(data.get('x', [])):
```
nepython, line 125: I just found that this is a bug, if there is only one histogram chart in Monitoring Charts (i.e. no time series charts) and you try to export it, this leads to KeyError.
```python
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/device/api/views.py", line 75, in get
    response = HttpResponse(self._get_csv(data), content_type='text/csv')
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/device/api/views.py", line 125, in _get_csv
    for index, element in enumerate(data['x']):
KeyError: 'x'
```
This change fixes it

nepython:
Done :+1:

openwisp_monitoring/device/api/views.py
nemesisdesign: Great, thanks

Issue #106, opened 2020-05-29
[alert settings] Add way to specify global defaults


nemesisdesign:
This is a draft and is open for discussion.
Related to #44, #52 and #93.

Right now, threshold objects are created whenever a metric is created.
If one wants to change its default values globally (value, operator, seconds), they would need to loop over all the available thresholds and change each one.

**Example**: the metrics being added in #76, we have different threshold values for memory, CPU, disk. What if a user having already 100 deployed devices wants to set the threshold value of the memory to 75%? With the change I'm proposing here the user would be able to do it just by changing a setting.

I've been thinking, what if we allow these fields to be NULL and we define their default values in the metric configuration (#44)?
For example, metrics that generate an alert, will specify this in their configuration, and will specify the default values of the fields that can be NULL.
We can then add model methods that retrieve the value, something like:

```python
def get_value(self):
    # if self.value is None, the default from the configuration will be used
    return self.value or self.config['value']
```

A solution like the one I described here would give us the possibility to change the defaults globally without having to change the database.

The user can then override specific thresholds by setting their own custom value if needed (eg: a device is particularly problematic and the user will want to change some parameters).

### Find a better name for the ``seconds`` field 

I also think we may use the chance to rename `seconds` to something more obvious, maybe `tolerance`.

### AlertSettings.threshold

Now that Threshold has been renamed to `AlertSettings`, it makes sense to rename `AlertSettings.value` to `AlertSettings.threshold`.

### Notification types

The goal of this section is to allow us to have more user friendly alerts/notifications.

After #43 is merged, we will have alerts for:

- ping problem/recovery
- RAM memory problem/recovery
- CPU usage problem/recovery
- disk usage problem/recovery

And after https://github.com/openwisp/openwisp-monitoring/issues/54 we will have another one.

But the alert text would be very similar, eg:

```
Metric Ping (device: cap-dev-01) returned within threshold limit.
Metric RAM memory (device: cap-dev-01) returned within threshold limit.
```

**Which is quite ugly.**

We have to achieve more user friendliness:

Ping PROBLEM
```
Subject: [openwisp] cap-dev-01 is not reachable

The device "cap-dev-01" (link) cannot be reached anymore by our ping messages since <UTC datetime>.
```

Ping RECOVERY
```
Subject: [openwisp] cap-dev-01 is reachable again

The device "cap-dev-01" (link) has become reachable again by our ping messages at <UTC datetime>.
```

Memory PROBLEM
```
[...]

PR #123, opened 2020-06-10
[feature] Add a check which inspects device configuration status periodically

### LINKED ISSUE #54: [feature] Add a check which inspects device configuration status periodically
The issue can be broken down as below :
Add a `Check`, `Metric` and a `relative alert setting` (created automatically for every `device`, it must take into account existing devices by using a data migration).

**Details of the Check**:
* The `Check` should run periodically (by default every 10 minutes via a celery task in the background).
* The `Check` shall find out if the configuration related to the `device` has been in the **modified** state for more than **X** minutes, where **X** is configurable via a setting and defaulting to **5**.
* It should also change the monitoring status of the device from **ok** to either **problem** or **critical**, depending on configuration (defaulting to **problem**).
* The `Check` should not fail but skip if the device has no defined configuration.
* Data stored in the timeseries DB which is older than **Y** days should be deleted, where **Y** is configurable with a django setting and defaults to **2** days.

Add documentation about this feature and the settings and also provide a setting (defaulting to `True`) for disabling the feature (in case it needs to be).

nepython:
Closes #54

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -213,6 +213,20 @@ command is used to collect these metrics.
 You may choose to disable auto creation of this check by setting
 `OPENWISP_MONITORING_AUTO_PING <#OPENWISP_MONITORING_AUTO_PING>`_ to ``False``.
 
+``Configuration applied``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This check ensures that the `openwisp-config agent <https://github.com/openwisp/openwisp-config/>`_
+is running and applying configuration changes in a timely manner.
+
+By default, if a configuration object stays in the ``modified`` state
+for more than 5 minutes (configurable via
+`OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME <#OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME>`_),
+the check will fail and set the health status of the device to ``PROBLEM``.
+
+You may choose to disable auto creation of this check by using the
+setting `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK <#OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK>`_.
+
 Settings
 --------
 
@@ -241,6 +255,33 @@ in terms of disk space.
 +--------------+-------------+
 
 Whether ping checks are created automatically for devices.
+
+``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+This setting allows you to choose whether `config_applied <#configuration-applied>`_ checks should be
+created automatically for newly registered devices. It's enabled by default.
+
+``OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    |  ``int``  |
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+Defines the maximum amount of minutes the device configuration can stay in the *modified*
+state before its health status is changed to ``PROBLEM`` and an alert is sent.
+
+**Note**: The setting will be ignored if ``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``
+is ``False``.
 
 ``OPENWISP_MONITORING_AUTO_CHARTS``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/openwisp_monitoring/check/apps.py
+++ b/openwisp_monitoring/check/apps.py
@@ -22,3 +22,12 @@ class CheckConfig(AppConfig):
                 sender=load_model('config', 'Device'),
                 dispatch_uid='auto_ping',
             )
+
+        if app_settings.AUTO_CONFIG_CHECK:
+            from .base.models import auto_config_check_receiver
+
+            post_save.connect(
+                auto_config_check_receiver,
+                sender=load_model('config', 'Device'),
+                dispatch_uid='auto_config_check',
+            )
--- a/openwisp_monitoring/check/base/models.py
+++ b/openwisp_monitoring/check/base/models.py
@@ -8,7 +8,7 @@ from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
 from jsonfield import JSONField
 from openwisp_monitoring.check import settings as app_settings
-from openwisp_monitoring.check.tasks import auto_create_ping
+from openwisp_monitoring.check.tasks import auto_create_config_check, auto_create_ping
 
 from openwisp_utils.base import TimeStampedEditableModel
 
@@ -92,6 +92,25 @@ def auto_ping_receiver(sender, instance, created, **kwargs):
     with transaction.atomic():
         transaction.on_commit(
             lambda: auto_create_ping.delay(
+                model=sender.__name__.lower(),
+                app_label=sender._meta.app_label,
+                object_id=str(instance.pk),
+            )
+        )
+
+
+def auto_config_check_receiver(sender, instance, created, **kwargs):
+    """
+    Implements OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK
+    The creation step is executed in the background
+    """
+    # we need to skip this otherwise this task will be executed
+    # every time the configuration is requested via checksum
+    if not created:
+        return
+    with transaction.atomic():
+        transaction.on_commit(
+            lambda: auto_create_config_check.delay(
                 model=sender.__name__.lower(),
                 app_label=sender._meta.app_label,
                 object_id=str(instance.pk),
--- a/openwisp_monitoring/check/classes/base.py
+++ b/openwisp_monitoring/check/classes/base.py
@@ -30,7 +30,7 @@ class BaseCheck(object):
     def check(self, store=True):
         raise NotImplementedError
 
-    def _get_or_create_metric(self):
+    def _get_or_create_metric(self, configuration=None):
         """
         Gets or creates metric
         """
@@ -44,7 +44,7 @@ class BaseCheck(object):
         options = dict(
             object_id=obj_id,
             content_type=ct,
-            configuration=self.__class__.__name__.lower(),
+            configuration=configuration or self.__class__.__name__.lower(),
         )
         metric, created = Metric._get_or_create(**options)
         return metric, created
--- a/openwisp_monitoring/check/classes/config_applied.py
+++ b/openwisp_monitoring/check/classes/config_applied.py
@@ -1,0 +1,35 @@
+from swapper import load_model
+
+from ...device.utils import SHORT_RP
+from ..settings import CONFIG_CHECK_MAX_TIME
+from .base import BaseCheck
+
+AlertSettings = load_model('monitoring', 'AlertSettings')
+
+
+class ConfigApplied(BaseCheck):
+    def check(self, store=True):
+        # If the device is down do not run config applied checks
+        if self.related_object.monitoring.status == 'critical':
+            return
+        if not hasattr(self.related_object, 'config'):
+            return
+        result = int(self.related_object.config.status == 'applied')
+        if store:
+            self._get_metric().write(
+                result, retention_policy=SHORT_RP,
+            )
+        return result
+
+    def _get_metric(self):
+        metric, created = self._get_or_create_metric(configuration='config_applied')
+        if created:
+            self._create_alert_setting(metric)
+        return metric
+
+    def _create_alert_setting(self, metric):
+        alert_s = AlertSettings(
+            metric=metric, operator='<', value=1, seconds=CONFIG_CHECK_MAX_TIME * 60
+        )
+        alert_s.full_clean()
+        alert_s.save()
--- a/openwisp_monitoring/check/settings.py
+++ b/openwisp_monitoring/check/settings.py
@@ -3,7 +3,17 @@ from django.conf import settings
 CHECK_CLASSES = getattr(
     settings,
     'OPENWISP_MONITORING_CHECK_CLASSES',
-    (('openwisp_monitoring.check.classes.Ping', 'Ping'),),
+    (
+        ('openwisp_monitoring.check.classes.Ping', 'Ping'),
+        ('openwisp_monitoring.check.classes.ConfigApplied', 'Configuration Applied'),
+    ),
 )
 AUTO_PING = getattr(settings, 'OPENWISP_MONITORING_AUTO_PING', True)
+AUTO_CONFIG_CHECK = getattr(
+    settings, 'OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK', True
+)
+# Input in minutes
+CONFIG_CHECK_MAX_TIME = getattr(
+    settings, 'OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME', 5
+)
 MANAGEMENT_IP_ONLY = getattr(settings, 'OPENWISP_MONITORING_MANAGEMENT_IP_ONLY', True)
--- a/openwisp_monitoring/check/tasks.py
+++ b/openwisp_monitoring/check/tasks.py
@@ -52,3 +52,27 @@ def auto_create_ping(model, app_label, object_id):
     check = Check(name='Ping', check=ping_path, content_type=ct, object_id=object_id)
     check.full_clean()
     check.save()
+
+
+@shared_task
+def auto_create_config_check(model, app_label, object_id):
+    """
+    Called by openwisp_monitoring.check.models.auto_config_check_receiver
+    """
+    Check = load_model('check', 'Check')
+    config_check_path = 'openwisp_monitoring.check.classes.ConfigApplied'
+    has_check = Check.objects.filter(
+        object_id=object_id, content_type__model='device', check=config_check_path,
+    ).exists()
+    # create new check only if necessary
+    if has_check:
+        return
+    ct = ContentType.objects.get(app_label=app_label, model=model)
+    check = Check(
+        name='Configuration Applied',
+        check=config_check_path,
+        content_type=ct,
+        object_id=object_id,
+    )
+    check.full_clean()
+    check.save()
--- a/openwisp_monitoring/check/tests/test_models.py
+++ b/openwisp_monitoring/check/tests/test_models.py
@@ -1,17 +1,27 @@
+from datetime import timedelta
+from unittest.mock import patch
+
 from django.core.exceptions import ValidationError
 from django.test import TransactionTestCase
+from django.utils.timezone import now
+from freezegun import freeze_time
 from swapper import load_model
 
 from ...device.tests import TestDeviceMonitoringMixin
-from ..classes import Ping
-from ..settings import CHECK_CLASSES
+from .. import settings as app_settings
+from ..classes import ConfigApplied, Ping
+from ..tasks import auto_create_config_check, auto_create_ping
 
 Check = load_model('check', 'Check')
 Metric = load_model('monitoring', 'Metric')
+AlertSettings = load_model('monitoring', 'AlertSettings')
+Device = load_model('config', 'device')
+Notification = load_model('openwisp_notifications', 'Notification')
 
 
 class TestModels(TestDeviceMonitoringMixin, TransactionTestCase):
-    _PING = CHECK_CLASSES[0][0]
+    _PING = app_settings.CHECK_CLASSES[0][0]
+    _CONFIG_APPLIED = app_settings.CHECK_CLASSES[1][0]
 
     def test_check_str(self):
         c = Check(name='Test check')
@@ -29,8 +39,14 @@ class TestModels(TestDeviceMonitoringMixin, TransactionTestCase):
         self.assertEqual(str(c), expected)
 
     def test_check_class(self):
-        c = Check(name='Ping class check', check=self._PING)
-        self.assertEqual(c.check_class, Ping)
+        with self.subTest('Test Ping check Class'):
+            c = Check(name='Ping class check', check=self._PING)
+            self.assertEqual(c.check_class, Ping)
+        with self.subTest('Test Configuration Applied check Class'):
+            c = Check(
+                name='Configuration Applied class check', check=self._CONFIG_APPLIED
+            )
+            self.assertEqual(c.check_class, ConfigApplied)
 
     def test_base_check_class(self):
         path = 'openwisp_monitoring.check.classes.base.BaseCheck'
@@ -42,34 +58,131 @@ class TestModels(TestDeviceMonitoringMixin, TransactionTestCase):
 
     def test_check_instance(self):
         obj = self._create_device(organization=self._create_org())
-        c = Check(
-            name='Ping class check', check=self._PING, content_object=obj, params={}
-        )
-        i = c.check_instance
-        self.assertIsInstance(i, Ping)
-        self.assertEqual(i.related_object, obj)
-        self.assertEqual(i.params, c.params)
+        with self.subTest('Test Ping check instance'):
+            c = Check(
+                name='Ping class check', check=self._PING, content_object=obj, params={}
+            )
+            i = c.check_instance
+            self.assertIsInstance(i, Ping)
+            self.assertEqual(i.related_object, obj)
+            self.assertEqual(i.params, c.params)
+        with self.subTest('Test Configuration Applied check instance'):
+            c = Check(
+                name='Configuration Applied class check',
+                check=self._CONFIG_APPLIED,
+                content_object=obj,
+                params={},
+            )
+            i = c.check_instance
+            self.assertIsInstance(i, ConfigApplied)
+            self.assertEqual(i.related_object, obj)
+            self.assertEqual(i.params, c.params)
 
     def test_validation(self):
-        check = Check(name='Ping check', check=self._PING, params={})
-        try:
-            check.full_clean()
-        except ValidationError as e:
-            self.assertIn('device', str(e))
-        else:
-            self.fail('ValidationError not raised')
+        with self.subTest('Test Ping check validation'):
+            check = Check(name='Ping check', check=self._PING, params={})
+            try:
+                check.full_clean()
+            except ValidationError as e:
+                self.assertIn('device', str(e))
+            else:
+                self.fail('ValidationError not raised')
+        with self.subTest('Test Configuration Applied check validation'):
+            check = Check(name='Config check', check=self._CONFIG_APPLIED, params={})
+            try:
+                check.full_clean()
+            except ValidationError as e:
+                self.assertIn('device', str(e))
+            else:
+                self.fail('ValidationError not raised')
 
-    def test_auto_ping(self):
+    def test_auto_check_creation(self):
         self.assertEqual(Check.objects.count(), 0)
         d = self._create_device(organization=self._create_org())
-        self.assertEqual(Check.objects.count(), 1)
-        c = Check.objects.first()
-        self.assertEqual(c.content_object, d)
-        self.assertIn('Ping', c.check)
+        self.assertEqual(Check.objects.count(), 2)
+        with self.subTest('Test AUTO_PING'):
+            c1 = Check.objects.filter(check=self._PING).first()
+            self.assertEqual(c1.content_object, d)
+            self.assertEqual(self._PING, c1.check)
+        with self.subTest('Test AUTO_CONFIG_CHECK'):
+            c2 = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+            self.assertEqual(c2.content_object, d)
+            self.assertEqual(self._CONFIG_APPLIED, c2.check)
 
     def test_device_deleted(self):
         self.assertEqual(Check.objects.count(), 0)
         d = self._create_device(organization=self._create_org())
-        self.assertEqual(Check.objects.count(), 1)
+        self.assertEqual(Check.objects.count(), 2)
         d.delete()
         self.assertEqual(Check.objects.count(), 0)
+
+    @patch('openwisp_monitoring.check.settings.AUTO_PING', False)
+    def test_config_modified_device_problem(self):
+        self._create_admin()
+        self.assertEqual(Check.objects.count(), 0)
+        self._create_config(status='modified', organization=self._create_org())
+        d = Device.objects.first()
+        self.assertEqual(Check.objects.count(), 2)
+        self.assertEqual(Metric.objects.count(), 0)
+        self.assertEqual(AlertSettings.objects.count(), 0)
+        check = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        with freeze_time(now() - timedelta(minutes=10)):
+            check.perform_check()
+        self.assertEqual(Metric.objects.count(), 1)
+        self.assertEqual(AlertSettings.objects.count(), 1)
+        # Check needs to be run again without mocking time for threshold crossed
+        check.perform_check()
+        m = Metric.objects.first()
+        self.assertEqual(m.content_object, d)
+        self.assertEqual(m.key, 'config_applied')
+        dm = d.monitoring
+        self.assertFalse(m.is_healthy)
+        self.assertEqual(dm.status, 'problem')
+        self.assertEqual(Notification.objects.count(), 1)
+
+    @patch(
+        'openwisp_monitoring.device.base.models.app_settings.CRITICAL_DEVICE_METRICS',
+        [{'key': 'config_applied', 'field_name': 'config_applied'}],
+    )
+    @patch('openwisp_monitoring.check.settings.AUTO_PING', False)
+    def test_config_check_critical_metric(self):
+        self._create_config(status='modified', organization=self._create_org())
+        self.assertEqual(Check.objects.count(), 2)
+        d = Device.objects.first()
+        dm = d.monitoring
+        check = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        check.perform_check()
+        self.assertEqual(Metric.objects.count(), 1)
+        self.assertEqual(AlertSettings.objects.count(), 1)
+        m = Metric.objects.first()
+        self.assertTrue(dm.is_metric_critical(m))
+        with freeze_time(now() - timedelta(minutes=10)):
+            check.perform_check()
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'critical')
+
+    def test_no_duplicate_check_created(self):
+        self._create_config(organization=self._create_org())
+        self.assertEqual(Check.objects.count(), 2)
+        d = Device.objects.first()
+        auto_create_config_check.delay(
+            model=Device.__name__.lower(),
+            app_label=Device._meta.app_label,
+            object_id=str(d.pk),
+        )
+        auto_create_ping.delay(
+            model=Device.__name__.lower(),
+            app_label=Device._meta.app_label,
+            object_id=str(d.pk),
+        )
+        self.assertEqual(Check.objects.count(), 2)
+
+    def test_device_unreachable_no_config_check(self):
+        self._create_config(status='modified', organization=self._create_org())
+        d = self.device_model.objects.first()
+        d.monitoring.update_status('critical')
+        self.assertEqual(Check.objects.count(), 2)
+        c2 = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        c2.perform_check()
+        self.assertEqual(Metric.objects.count(), 0)
+        self.assertIsNone(c2.perform_check())
--- a/openwisp_monitoring/db/backends/influxdb/client.py
+++ b/openwisp_monitoring/db/backends/influxdb/client.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.utils.functional import cached_property
+from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from influxdb import InfluxDBClient
 
@@ -112,11 +113,10 @@ class DatabaseClient(object):
             'tags': kwargs.get('tags'),
             'fields': values,
         }
-        timestamp = kwargs.get('timestamp')
+        timestamp = kwargs.get('timestamp') or now()
         if isinstance(timestamp, datetime):
-            timestamp = timestamp.strftime('%Y-%m-%dT%H:%M:%SZ')
-        if timestamp:
-            point['time'] = timestamp
+            timestamp = timestamp.isoformat(sep='T', timespec='microseconds')
+        point['time'] = timestamp
         self.get_db.write(
             {'points': [point]},
             {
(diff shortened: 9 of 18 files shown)


nepython:
**QUERIES**:
* When is the status supposed to be made critical? The issue metioned that status should change to `problem/critical` depending upon the configuration. I didn't understood this. 
* I think that I should rename all instances of config_status to config_modified, it looks more appropriate. Can I please know if it seems a better name?

PabloCastellano:
> **QUERIES**:
> 
> * When is the status supposed to be made critical? The issue metioned that status should change to `problem/critical` depending upon the configuration. I didn't understand this.

Check out the README file. According to the current version, a status will be critical when the metric is defined in `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS`. Critical status is reserved for some metrics such as ping.

README.rst
```diff
@@ -247,6 +247,47 @@ in terms of disk space.
...
+``OPENWISP_MONITORING_AUTO_CONFIG_MODIFIED``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+Whether ``config_modified`` checks are created automatically for devices.
```
PabloCastellano, line 259: This should be explained in a better way, since `config_modified check` doesn't say much at all to newcomers

README.rst
```diff
@@ -247,6 +247,47 @@ in terms of disk space.
...
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+After ``config`` is ``modified``, if the ``modified`` status does not change  after a
+fixed **duration** then ``device`` health status shall change to ``problem``.
+This fixed **duration** is 5 minutes by defaut and can be changed with the help of this setting.
+The input represents corresponding duration in minutes.
+
+**Note**: If the setting ``AUTO_CONFIG_STATUS`` is disabled then this setting need not be declared.
```
PabloCastellano, line 275: What is this AUTO_CONFIG_STATUS setting? Please review that everything makes sense when making changes and submitting PRs

openwisp_monitoring/check/settings.py
```diff
@@ -3,7 +3,21 @@
...
+AUTO_CONFIG_MODIFIED = getattr(
+    settings, 'OPENWISP_MONITORING_AUTO_CONFIG_MODIFIED', True
+)
+# Input in minutes
+CONFIG_MODIFIED_MAX_TIME = getattr(
+    settings, 'OPENWISP_MONITORING_CONFIG_MODIFIED_MAXIMUM_TIME', 5
+)
+# Input in days
+CONFIG_MODIFIED_RETENTION_POLICY = getattr(
+    settings, 'OPENWISP_MONITORING_CONFIG_STATUS_RETENTION_POLICY', '48h0m0s'
```
PabloCastellano, line 21: Please follow the standard we use for setting names: the user defines `OPENWISP_MONITORING_X_Y_Z` but we use `X_Y_Z` internally

openwisp_monitoring/check/tasks.py
```diff
@@ -54,3 +54,32 @@ def auto_create_ping(model, app_label, object_id, created):
...
+    Called by openwisp_monitoring.check.models.auto_config_modified_receiver
+    """
+    Check = load_model('check', 'Check')
+    config_modified_path = 'openwisp_monitoring.check.classes.ConfigModified'
+    has_check = (
+        Check.objects.filter(
+            object_id=object_id,
+            content_type__model='device',
+            check=config_modified_path,
+        ).count()
```
PabloCastellano, line 71: Tip: there is `.exists()` which replaces `.count() > 0`

openwisp_monitoring/check/base/models.py
```diff
@@ -75,27 +76,45 @@ def perform_check(self, store=True):
         return self.check_instance.check(store=True)
 
 
-if app_settings.AUTO_PING:
```
PabloCastellano, line 78: I preferred the previous approach to be honest because previously we only registered the dispatcher when necessary (AUTO_PING=True). Now it is always registered (no matters the value of AUTO_PING) and hence executed on each post_save signal, causing the function to do nothing when AUTO_PING=False, note the difference.

README.rst
```diff
@@ -247,6 +247,47 @@ in terms of disk space.
...
+
++--------------+-----------+
+| **type**:    |  ``int``  |
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+After ``config`` is ``modified``, if the ``modified`` status does not change  after a
+fixed **duration** then ``device`` health status shall change to ``problem``.
+This fixed **duration** is 5 minutes by defaut and can be changed with the help of this setting.
```
PabloCastellano, line 272: typo: defaut

README.rst
```diff
@@ -247,6 +247,47 @@ in terms of disk space.
...
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+Whether ``config_modified`` checks are created automatically for devices.
+
+``OPENWISP_MONITORING_CONFIG_MODIFIED_MAX_TIME``
```
PabloCastellano, line 261: Incorrect: you have defined `OPENWISP_MONITORING_CONFIG_MODIFIED_MAXIMUM_TIME` in settings.py

README.rst
```diff
@@ -247,6 +247,47 @@ in terms of disk space.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    |  ``int``  |
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+After ``config`` is ``modified``, if the ``modified`` status does not change  after a
```
PabloCastellano, line 270: Sounds very technical too.

What about something like

> This setting allows configuring the time that the check is allowed to fail until the device health status changes to ``problem``

It is important to take the necessary time to re-read the documentation multiple times to make sure it is written properly and others can understand it. It is not easy though!

openwisp_monitoring/check/tasks.py
```diff
@@ -54,3 +54,32 @@ def auto_create_ping(model, app_label, object_id, created):
...
+        > 0
+    )
+    # create new check only if necessary
+    if has_check:
+        return
+    ct = ContentType.objects.get(app_label=app_label, model=model)
+    check = Check(
+        name='Config Modified',
```
PabloCastellano, line 79: What about 'Device configuration' or 'Configuration Modified'? The name shouldn't contain abbreviations

openwisp_monitoring/check/tasks.py
nepython: I do know :sweat_smile:. I thought of using it but then dropped the idea to be consistent with `ping`'s code. Thanks, I shall change for both the checks, then :)

openwisp_monitoring/check/base/models.py
nepython: Makes sense but then I shall have to add a check in corresponding tasks. The reason being, the dispatcher shall get registered when the models are loaded. So, there is no option of mocking `AUTO_PING`, `AUTO_CONFIG_STATUS` it seems. Except for the tests, there is no real need of the check in the tasks.

README.rst
nepython: I was trying to be consistent with the `ping` check but yes you are right. I shall work upon improving the documentation :smiley:

openwisp_monitoring/check/settings.py
nepython: Can there be an exception here. I have named the setting as `OPENWISP_MONITORING_DEVICE_CONFIGURATION_X_Y_Z` but internally the settings are named as `CONFIG_X_Y_Z`. The reason was to explicitly say to newcomers that the setting is related to `DEVICE_CONFIGURATION` (there's also chart config and soon there shall be metric_config) but internally i.e. in the `check settings` we know that it means device config hence it's shortened. Let me know if it seems right?

nepython:
@nemesisdesign, I realized it now that there is no migration for `ping`, `config_modified` checks in the `sample_check`. I don't think there's a need since the module is not yet released (and the migrations were meant for backward compatibilty). Can you please confirm this?

openwisp_monitoring/check/tasks.py
```diff
@@ -35,22 +35,43 @@ def perform_check(uuid):
 
 
 @shared_task
-def auto_create_ping(model, app_label, object_id, created):
+def auto_create_ping(model, app_label, object_id):
```
nepython, line 38: I think it's safe to remove the the keyword argument `created` , since it's not being used at all.

README.rst
```diff
@@ -247,6 +247,53 @@ in terms of disk space.
...
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+``config_modified`` checks are run peridically to monitor whether device
+configuration is applied properly.
+This setting is used to govern whether ``config_modified`` checks are
```
nemesisdesign, line 261: govern? Can you find a different way of explaining this?

README.rst
```diff
@@ -247,6 +247,53 @@ in terms of disk space.
 
 Whether ping checks are created automatically for devices.
 
+``OPENWISP_MONITORING_AUTO_DEVICE_CONFIGURATION_MODIFIED``
```
nemesisdesign, line 250: what about `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK`?

README.rst
```diff
@@ -247,6 +247,53 @@ in terms of disk space.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    |  ``int``  |
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+This setting allows you to configure the time that the check is allowed to
+fail after which the device health status changes to ``problem``.
```
nemesisdesign, line 274: I don't understand this setting from its explanation, we should also find a shorter name, eg `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME`

README.rst
```diff
@@ -247,6 +247,53 @@ in terms of disk space.
...
+
+This default **duration** is 5 minutes. The input represents corresponding
+duration in minutes.
+
+**Note**: If the setting ``OPENWISP_MONITORING_AUTO_DEVICE_CONFIGURATION_MODIFIED``
+is disabled then this setting need not be declared.
+
+``OPENWISP_MONITORING_DEVICE_CONFIGURATION_MODIFIED_RETENTION_POLICY``
```
nemesisdesign, line 282: too long.. `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_RP`?

openwisp_monitoring/check/apps.py
```diff
@@ -1,8 +1,13 @@
...
+from .utils import manage_config_modified_retention_policy
+
 
 class CheckConfig(AppConfig):
     name = 'openwisp_monitoring.check'
     label = 'check'
     verbose_name = _('Network Monitoring Checks')
+
+    def ready(self):
+        manage_config_modified_retention_policy()
```
nemesisdesign, line 13: how does this play with the abstraction? Can you make it play well with the changes introduced in #103?

openwisp_monitoring/check/classes/base.py
```diff
@@ -0,0 +1,38 @@
...
+from django.core.exceptions import ValidationError
+from swapper import load_model
+
+from openwisp_controller.config.models import Device
+
+Check = load_model('check', 'Check')
+Metric = load_model('monitoring', 'Metric')
+
+
+class BaseCheck(object):
```
nemesisdesign, line 11: I would add:

```python
def validate(self):
    self.validate_instance()
    self.validate_params()

def validate_params(self):
    pass
```

README.rst
```diff
@@ -247,6 +247,53 @@ in terms of disk space.
...
++--------------+-------------+
+| **default**: | ``48h0m0s`` |
++--------------+-------------+
+
+This setting allows to modify the duration for which the metric data generated
+by ``config_modified`` check is to be retained.
+
+**Note**: If the setting ``OPENWISP_MONITORING_AUTO_DEVICE_CONFIGURATION_MODIFIED``
+is disabled then this setting need not be declared.
+
```
nemesisdesign, line 296: I think we should start creating a section which describes the available checks, which can go above, for now we'll have two checks: ping and this new one, we need a description for each one, and each section can then link the available settings for each check.

I would like to see a prose description of this new check.

openwisp_monitoring/check/classes/base.py
nepython: Thanks!

openwisp_monitoring/check/apps.py
nepython: `manage_config_modified_retention_policy()` makes use of `timeseries_db.get_list_retention_policies` after rebasing which has been defined by us so it's abstract enough, I think.

openwisp_monitoring/check/classes/base.py
```diff
@@ -0,0 +1,45 @@
...
+            message = 'A related device is required to perform this operation'
+            raise ValidationError({'content_type': message, 'object_id': message})
+
+    def validate(self):
+        self.validate_instance()
+        self.validate_params()
+
+    def validate_params(self):
+        pass
+
```
nemesisdesign, line 25: I forgot something else, which is just something which is usually done when developing these kind of object oriented patterns:

```python
def check(self, store=True):
    raise NotImplementedError
```

It's usage is recommended in the python docs, I suggest to read this passage: https://docs.python.org/3/library/exceptions.html#NotImplementedError

README.rst
```diff
@@ -224,6 +224,32 @@ has a value which is not in the expected range
 Example: ping is by default a critical metric which is expected to be always 1
 (reachable).
 
+``Available Checks``
+--------------------
+
+``Ping``
+~~~~~~~~
+
+This check returns information on device ``uptime``, ``downtime``,
```
PabloCastellano, line 233: downtime is redundant, saying uptime is enough

README.rst
```diff
@@ -224,6 +224,32 @@ has a value which is not in the expected range
...
+``RTT (Round trip time) maximum, minimum and average``.
+It creates a ``Ping`` metric to store this information and an ``AlertSettings``
+to monitor them.
```
PabloCastellano, line 235: > It creates a ``Ping`` metric to store this information and an ``AlertSettings`` to monitor them.

These are technical details that a normal user prefers not to know

README.rst
```diff
@@ -224,6 +224,32 @@ has a value which is not in the expected range
...
+The Charts ``uptime``, ``packet loss`` and ``rtt`` are created. The ``fping``
+command is used to collect these metrics.
+You may choose to disable auto creation of this check by using the setting
```
PabloCastellano, line 238: Instead of `by using the setting ...` I'd say `by setting ... to False` which is more concrete

README.rst
```diff
@@ -224,6 +224,32 @@ has a value which is not in the expected range
...
+`OPENWISP_MONITORING_AUTO_PING <#OPENWISP_MONITORING_AUTO_PING>`_.
+
+``Configuration Modified``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It checks the configuration status of the device periodically and changes
+the device health status to ``PROBLEM``, if config status stays modified
+for more than `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME <#OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME>`_.
+You may choose to disable auto creation of this check by using the
+setting `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK <#OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK>`_.
```
PabloCastellano, line 248: Same: say "setting it to False"

README.rst
```diff
@@ -224,6 +224,32 @@ has a value which is not in the expected range
...
+to monitor them.
+The Charts ``uptime``, ``packet loss`` and ``rtt`` are created. The ``fping``
+command is used to collect these metrics.
+You may choose to disable auto creation of this check by using the setting
+`OPENWISP_MONITORING_AUTO_PING <#OPENWISP_MONITORING_AUTO_PING>`_.
+
+``Configuration Modified``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It checks the configuration status of the device periodically and changes
```
PabloCastellano, line 244: What about this rewording

"This will periodically check the configuration present in the device and will change the device health status to ``PROBLEM`` if the configuration is not in sync with the expected by OpenWISP. The time until this check changes the health status can be configured using the setting `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME`"

README.rst
```diff
@@ -253,6 +279,49 @@ in terms of disk space.
...
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+``config_modified`` check changes the device health status to ``PROBLEM`` if it's
+configuration status has been in the modified state for more than  the time defined in
+``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``. It's 5 minutes by default.
+
+**Note**: If the setting ``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``
+is disabled then this setting need not be declared.
```
PabloCastellano, line 307: "This setting will be ignored if `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK = False`"

README.rst
```diff
@@ -253,6 +279,49 @@ in terms of disk space.
...
+
++--------------+-----------+
+| **type**:    |  ``int``  |
++--------------+-----------+
+| **default**: |   ``5``   |
++--------------+-----------+
+
+``config_modified`` check changes the device health status to ``PROBLEM`` if it's
+configuration status has been in the modified state for more than  the time defined in
+``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``. It's 5 minutes by default.
```
PabloCastellano, line 304: "Defines the amount of time that the device configuration can be different from what is expected until its health status is changed to PROBLEM"

README.rst
```diff
@@ -253,6 +279,49 @@ in terms of disk space.
...
+**Note**: If the setting ``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``
+is disabled then this setting need not be declared.
+
+``OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_RETENTION_POLICY``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    |  ``time``   |
++--------------+-------------+
+| **default**: | ``48h0m0s`` |
```
PabloCastellano, line 315: I think this is actually too low. Can you set it to 30 days by default?

openwisp_monitoring/check/classes/base.py
```diff
@@ -0,0 +1,48 @@
...
+from django.core.exceptions import ValidationError
+from swapper import load_model
+
+from openwisp_controller.config.models import Device
+
+Check = load_model('check', 'Check')
+Metric = load_model('monitoring', 'Metric')
+
+
+class BaseCheck(object):
```
PabloCastellano, line 11: While refactoring BaseCheck is needed, this kind of changes make it more difficult to review PRs and detect problems later. It is usually a good thing to make one change per PR so please consider it next time (1 PR for refactoring and another one to add this new check).

openwisp_monitoring/check/classes/base.py
```diff
@@ -0,0 +1,48 @@
...
+            obj_id = str(check.id)
+            ct = ContentType.objects.get_for_model(Check)
+        options = dict(
+            name=check.name,
+            object_id=obj_id,
+            content_type=ct,
+            field_name=field_name,
+            key=self.__class__.__name__.lower(),
+        )
+        metric, created = Metric.objects.get_or_create(**options)
```
PabloCastellano, line 47: Shouldn't we use `Metric._get_or_create` ?

README.rst
nepython: It sounds a bit technical.

README.rst
nepython: I can set it but the issue (derived from measurable outcomes) mentioned a default retention period of 2 days. It made sense since I don't think the user would be interested to know this info at all and it's for internal usage.

openwisp_monitoring/device/apps.py
```diff
@@ -162,3 +164,23 @@ def register_notifcation_types(self):
...
+                dispatch_uid='config_modified',
+            )
+
+    @classmethod
+    def config_modified_receiver(cls, sender, instance, **kwargs):
+        from ..check.tasks import perform_check
+
+        DeviceData = load_model('device_monitoring', 'DeviceData')
+        device = DeviceData.objects.get(config=instance)
+        check = device.checks.get(name='Configuration Modified')
```
nemesisdesign, line 185: never query in the code by name, it's too fragile!

The name right now can also be changed by users so this would stop working in that case.
Use another field, eg: `configuration`.

Moreover, what if the Check has been deleted? Please write a test for this case and ensure it's handled without crashes.

README.rst
nemesisdesign:
```
Defines the maximum amount of minutes the device configuration can stay in the *modified* state before its health status is changed to PROBLEM and an alert is sent.
```

nepython:
> What if the device is already in critical state (eg: turned off)?
> 
> The scenario is the following.
> 
> User has 10 devices using the same configuration template.
> 3 devices are offline (unplugged, turned off, whatever).
> User edits one of the common templates, hence the configuration status of all the devices changes to modified.
> The 3 devices that are offline will not get their config status updated because they're offline.
> 
> The user does not expect (nor wants) to not be notified about this.
> The user does expect to know if there's any issue once the devices do come back online.

How's this. If the device health status becomes `critical` (due to any reason), config modified checks won't be run thus there won't be metric write, threshold_crossed and thus no alert generated.
When the device comes back online (not critical anymore), `config_modified` checks shall be triggered and if status stays `modified` for more than threshold period, an alert will be sent.
Thus besides the celery tasks, `config_modified` checks shall be triggered when config gets modified / device comes back online.

This should cover the case, so let me know :smile:

nepython:
> Is this ready for review/testing?

Can you please confirm that https://github.com/openwisp/openwisp-monitoring/pull/123#issuecomment-654134969 covers the use case well, since I am unsure about it.

nepython:
@nemesisdesign, I went ahead with https://github.com/openwisp/openwisp-monitoring/pull/123#issuecomment-654134969, this is the code that I have added
https://github.com/openwisp/openwisp-monitoring/blob/bf712870125a05d347436857ccc7fc52289bae16/openwisp_monitoring/check/classes/config_modified.py#L12-L14
It checks device status, if it's `critical` then `config_modified` checks won't be run and thus no corresponding alerts. I think it covers the use case well :smile:.

README.rst
```diff
@@ -242,6 +251,33 @@ in terms of disk space.
...
+``OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+This setting allows you to choose whether `config_modified <#Configuration>` checks should be
```
nemesisdesign, line 263: see how it looks: https://github.com/openwisp/openwisp-monitoring/blob/125f1c7fb11280cb3435f6545cf7171b893d9aaa/README.rst#id2

Always do these manual visual checks before publishing docs.

openwisp_monitoring/check/base/models.py
```diff
@@ -91,3 +94,22 @@ def auto_ping_receiver(sender, instance, created, **kwargs):
                 object_id=str(instance.pk),
             )
         )
+
+
+def auto_config_modified_receiver(sender, instance, created, **kwargs):
```
nemesisdesign, line 99: instead of referring to `config_modified`, let's refer to `config_check`, follow the same suggestion for all other occurrences

openwisp_monitoring/monitoring/metrics.py
```diff
@@ -12,6 +12,12 @@
         'field_name': 'reachable',
         'related_fields': ['loss', 'rtt_min', 'rtt_max', 'rtt_avg'],
     },
+    'configmodified': {
```
nemesisdesign, line 15: here I would use `config_applied` and same for `key`and `field_name`

nepython:
> Thank you @nepython, I tested it and it works! +1
> 
> After using the feature and analyzing the code, I realized that the naming is odd.
> 
> The check would be a lot more easy to understand if it's something like:
> 
> name: Configuration applied
> alert settings operator: less than
> alert settings threshold value: 1
> alert settings threshold value: 600
> 
> So if the configuration is not applied for more than 5 minutes, we get an alert.
> 
> PS: while testing this, I also realized this: #172

@nemesisdesign, I am slightly confused. You are saying it should be 5 min and 600s at the same time :thinking:.
Currently it defaults to 5 minutes with a setting `OPENWISP_MONITORING_CONFIG_CHECK_MAX_TIME` to allow user to configure if needed (as per the issue description). I think the current approach of default 5 min with a setting to configure is better. Wdyt?

openwisp_monitoring/check/migrations/0004_create_config_applied.py
```diff
@@ -0,0 +1,33 @@
...
+        auto_create_config_check.delay(
+            model=Device.__name__.lower(),
+            app_label=Device._meta.app_label,
+            object_id=str(device.pk),
+        )
+
+
+def remove_config_applied_checks(apps, schema_editor):
+    Check = apps.get_model('config', 'Device')
+    Check.objects.filter(name='Configuration Applied').delete()
```
nemesisdesign, line 20: you should never rely on fields that can be changed like names and descriptions, you should query based on data which  cannot be altered by users.

openwisp_monitoring/check/migrations/0004_create_config_applied.py
```diff
@@ -0,0 +1,33 @@
...
+    for device in Device.objects.all():
+        auto_create_config_check.delay(
+            model=Device.__name__.lower(),
+            app_label=Device._meta.app_label,
+            object_id=str(device.pk),
+        )
+
+
+def remove_config_applied_checks(apps, schema_editor):
+    Check = apps.get_model('config', 'Device')
```
nemesisdesign, line 19: **TOTALLY BOGUS**. 19 days should have been enough to test this and make it work but yet, you fail at this Hardik. :-1: you're still sending broken code for final review after all the discussions we had. You have to stop doing this.

README.rst
```diff
@@ -213,6 +213,15 @@ command is used to collect these metrics.
...
 `OPENWISP_MONITORING_AUTO_PING <#OPENWISP_MONITORING_AUTO_PING>`_ to ``False``.
 
+``Configuration Applied``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It checks the configuration status of the device periodically and changes
+the device health status to ``PROBLEM``, if config status stays modified
+for more than `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME <#OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME>`_.
+You may choose to disable auto creation of this check by using the
+setting `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK <#OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK>`_.
```
nemesisdesign, line 223: I think this is better:

```
``Configuration applied``
~~~~~~~~~~~~~~~~~~~~~~~~~

This check ensures that the `openwisp-config agent
<https://github.com/openwisp/openwisp-config/>`_ is running and applying
configuration changes in a timely manner.

By default, if a configuration object stays in the ``modified`` state
for more than 5 minutes (configurable via
`OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME
<#OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME>`_), the check will
fail and set the health status of the device to ``PROBLEM``.

You may choose to disable auto creation of this check by using the
setting `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK
<#OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK>`_.
```

openwisp_monitoring/check/migrations/0004_create_config_applied.py
nemesisdesign: 1. try to run this
2. make sure it deletes also the Metric (and related alert setting cascades)

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -42,34 +56,128 @@ def test_base_check_class(self):
...
-        c = Check.objects.first()
-        self.assertEqual(c.content_object, d)
-        self.assertIn('Ping', c.check)
+        self.assertEqual(Check.objects.count(), 2)
+        with self.subTest('Test AUTO_PING'):
+            c1 = Check.objects.get(name='Ping')
+            self.assertEqual(c1.content_object, d)
+            self.assertIn('Ping', c1.check)
+        with self.subTest('Test AUTO_CONFIG_CHECK'):
+            c2 = Check.objects.get(name='Configuration Applied')
```
nemesisdesign, line 106: don't query by name, use `Check.objects.filter(check=self._CONFIG_APPLIED).first()`

fix the same issue for all other occurrences across the codebase!

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -42,34 +56,128 @@ def test_base_check_class(self):
...
+    def test_auto_check_creation(self):
         self.assertEqual(Check.objects.count(), 0)
         d = self._create_device(organization=self._create_org())
-        self.assertEqual(Check.objects.count(), 1)
-        c = Check.objects.first()
-        self.assertEqual(c.content_object, d)
-        self.assertIn('Ping', c.check)
+        self.assertEqual(Check.objects.count(), 2)
+        with self.subTest('Test AUTO_PING'):
+            c1 = Check.objects.get(name='Ping')
```
nemesisdesign, line 102: don't query by name, use the check field, please correct the same for any other occurrence of this everywhere!
Do not query by fields that can be changed by users, do not use get() for querying objects that can return multiple items!

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -42,34 +56,128 @@ def test_base_check_class(self):
...
+            self.assertEqual(c1.content_object, d)
+            self.assertIn('Ping', c1.check)
+        with self.subTest('Test AUTO_CONFIG_CHECK'):
+            c2 = Check.objects.get(name='Configuration Applied')
+            self.assertEqual(c2.content_object, d)
+            self.assertIn('ConfigApplied', c2.check)
```
nemesisdesign, line 108: why don't you use equal and you compare it to `self._CONFIG_APPLIED`

openwisp_monitoring/check/classes/config_applied.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def get_metric(self):
+        metric, created = self._get_or_create_metric(configuration='config_applied')
+        if created:
+            self._create_alert_setting(metric)
+        return metric
+
+    def _create_alert_setting(self, metric):
+        alert_s = AlertSettings(
+            metric=metric, operator='<', value=1, seconds=CONFIG_CHECK_MAX_TIME * 60
```
nemesisdesign, line 32: keep in mind this will have to be changed when we do https://github.com/openwisp/openwisp-monitoring/issues/106

openwisp_monitoring/check/tests/test_models.py
nepython: > do not use get() for querying objects that can return multiple items!

I expect one `Ping` and one `ConfigApplied` check per device. Since, the test had only one device, I thought using `get` was safe enough. Though yes, querying by `name` (which can be changed) is bad practice on my side :(

nepython:
> Also, the check doesn't seem to work to me if the tolerance is not zero seconds, I'll do more tests and let you know more details if I can. **UPDATE**: I confirm this, here's how I test it:
> 
>     * run dev server, run worker, run celery beat
> 
>     * set the config of a device to `modified`
> 
>     * wait more than 5 minutes, no alert is ever generated
> 
>     * if I set `seconds` to `0` in the alert settings, the alert is generated

@nemesisdesign, no it's not a bug. You had asked to cover this [use case](https://github.com/openwisp/openwisp-monitoring/pull/123#pullrequestreview-442628533) earlier :confused:, I replied back with a solution https://github.com/openwisp/openwisp-monitoring/pull/123#issuecomment-654134969. Awaited your response, pinged you two days later https://github.com/openwisp/openwisp-monitoring/pull/123#issuecomment-655560372 and on IM. Still no response, maybe you were slightly busy back then. Hence, went ahead with the approach and informed nevertheless https://github.com/openwisp/openwisp-monitoring/pull/123#issuecomment-656330919.

As you had requested, the `ConfigApplied` check won't send any alerts if the device goes offline (health is thus `critical`). The user will already have one for `Ping`. The reason being to avoid irritating the user with too many alerts. It will do alert the user once device comes back online and `config` still stays modified.

The zero second case I think is an edge case where device goes offline. `ConfigApplied` checks are triggered prior to health status getting updated to `CRITICAL`. We might never set a `0s` threshold for `ConfigApplied` I think, since config status stays modified for some time whenever a template is updated (it's not spontaneous atleast for me :sweat_smile:)

**Note**: There is one specific test (`test_config_modified_device_problem`) that I have added in `check.tests.TestModels` which tests that the `ConfigApplied` check is working!

openwisp_monitoring/check/tests/test_models.py
nepython: This test was already written for `Ping` check. So, I have just adapted it for `ConfigApplied` checks. Made the change.

openwisp_monitoring/check/classes/config_applied.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ConfigApplied(BaseCheck):
+    def check(self, store=True):
+        # If the device is down do not run config applied checks
+        if self.related_object.monitoring.status == 'critical':
+            return
+        if not hasattr(self.related_object, 'config'):
+            return
+        result = 1 if self.related_object.config.status == 'applied' else 0
```
nemesisdesign, line 17: this could be rewritten as:

```python
int(self.related_object.config.status == 'applied')
```

not sure which one is more readable, I think this form is compact and readable

nepython:
Mentioning so if anyone faces same issue in future, then can avoid the trouble, hopefully.
> Please repeat the same tests and let me know:
> 
>     1. if you can replicate the same issue successfully

Yes
>     2. if yes, why it is happening? Why writing with the short retention policy doesn't write anything?

When a retention policy is used besides `default` in `influxdb` then it needs to be specified during query.
https://community.influxdata.com/t/influxdb-not-writing-data-if-rp-is-specified/9861. So, the issue is not with `write` but with `read`. To query any measurement in `influxdb` with a non-default retention_policy in place, please use.
```
> select * from <retention_policy>.<measurement>;   # `select * from short.config_applied` in this case 
```
>     3. if yes, write a failing test for it

Added `openwisp_monitoring.db.backends.influxdb.tests.TestDatabaseClient.test_read_with_rp`
>     4. if yes, how you fixed it

Added `retention_policy` as a parameter in `metric.read()` and `timeseries_db.read()`. So, you can query like this,
```python
In [7]: m.read(retention_policy='short')
Out[7]: 
[{'time': 1596102060, 'config_applied': 1},
 {'time': 1596102120, 'config_applied': 1},
...
```
Personally, I don't like this approach since user will need to specify the `retention_policy` everytime in read / write. Maybe we can refactor `retention_policy` to be a field in `Metric` model (later on).

The alert didn't came immediately after 5 minutes (took roughly 6-7 minutes, I think) but yes tested twice and added `Notification count` in  `test_config_modified_device_problem`
Screenshot from 2020-07-31 02-15-21

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -130,11 +130,15 @@ def read(self, key, fields, tags, **kwargs):
...
+        retention_policy = kwargs.get('retention_policy')
         if extra_fields and extra_fields != '*':
             fields = ', '.join([fields] + extra_fields)
         elif extra_fields == '*':
             fields = '*'
-        q = f'SELECT {fields} FROM {key}'
+        if not retention_policy:
+            q = f'SELECT {fields} FROM {key}'
+        else:
+            q = f'SELECT {fields} FROM {retention_policy}.{key}'
```
nemesisdesign, line 141: do not repeat the whole query, follow the DRY principle please:

```python
from_clause = key
if retention_policy:
    from_clause = f'{retention_policy}.{key}'
q = f'SELECT {fields} FROM {from_clause}'
```

or a more compact version:

```python
from_clause = f'{rt}.{key}' if rt else key
q = f'SELECT {fields} FROM {from_clause}'
```

nepython:
> > > ```
> > > 4. if yes, how you fixed it
> > > ```
> > 
> > 
> > Added `retention_policy` as a parameter in `metric.read()` and `timeseries_db.read()`.
> 
> Ok, I see you added support for this in the tsdb backend, but the short retention policy was already being used to retrieve the device data, and that code has not been updated. Can you update that bit of code as well to make it use the new way so we can probably simplify the code? I'm referring to these:
> 
> https://github.com/openwisp/openwisp-monitoring/blob/87e7135b4b6e0b4a0c744f0aaa87b021912eae50/openwisp_monitoring/device/base/models.py#L99
> 
> https://github.com/openwisp/openwisp-monitoring/blob/238a17bcfebef619c5c125a130abcb076d1a9c4b/openwisp_monitoring/db/backends/influxdb/queries.py#L67

@nemesisdesign, as you have correctly pointed it out, `timseries_db.query()` was being used and not `timeseries_db.read()` for `device_data`. This method of how we query `device_data` will change fully after #164 (reason is elasticsearch doesn't support querying large text fields (efficiency issues) so had to adopt a different approach). It will simplify and transform as below,
https://github.com/openwisp/openwisp-monitoring/blob/9c53057826befd121d92035885a972163b93134c/openwisp_monitoring/device/base/models.py#L100-L102
also abstraction was not complete as `device_data_query` is a `dict` in case of `elasticsearch`. I think if needed, we can simplify it there :smiley:.

> In order to have robust tests, I think we should switch to use freezegun to arbitrarily manipulate the time, we already do this in other modules, can you please take advantage of this chance to change all the tests of this test suite which have this approach so that they use freezegun.freeze_time instead?
> 
> You would freeze the time 10 minutes earlier to trigger the check, that would trigger the insertion in the timeseries DB without the need of this line, this will make the testcases a lot more robust and will help us catch bugs in the future.

Ok, thanks freezegun works great. Currently, implemented only for this test, will do for all others in the suite. Though we will need to run checks twice in order to validate alerts after this. The reason being `datetime.now()` is then mocked throughout the code to point 10 minutes backwards. Hence, `alert_settings._is_crossed` also gets `now()` as `now() - timedelta(minutes=10)`. So, two times `perform_check` needs to be run, one with `freeze_time` (context manager) and one without it. It works perfectly with this.

To get it working, I had to alter tsdb backend `write` method, this seems to be causing failure of many other tests (maybe clock differences (unsure)). I am checking out why, change is simple enough :thinking:.
```python
timestamp = kwargs.get('timestamp') or now()  # <--- now() is extra here
if isinstance(timestamp, datetime):
    timestamp = timestamp.strftime('%Y-%m-%dT%H:%M:%SZ')
point['time'] = timestamp
```

nepython:
> To get it working, I had to alter tsdb backend write method, this seems to be causing failure of many other tests (maybe clock differences (unsure)). I am checking out why, change is simple enough thinking.

Check this build (https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/713662695#L1163) to know the failure.

Found it, it was related to `precision` of the timestamp. This issue was hidden till now and came forward now due to below change.
```python
timestamp = kwargs.get('timestamp') or now()  # now() is added here
if isinstance(timestamp, datetime):
    timestamp = timestamp.strftime('%Y-%m-%dT%H:%M:%SZ')  # This was the reason, seconds precision
point['time'] = timestamp    # Passing timestamp was optional earlier
```
I will try to add a failing test for this.

PR #126, opened 2020-06-11
[bug] Fix alerts sent for new metrics within threshold

### LINKED ISSUE #125: [alerts] Alert being sent on first metric collection
I've been testing the new metrics added by @nepython and I noticed something that needs to be improved.

Expected behavior:

- metrics are collected and are within the allowed values (eg: CPU is lower than 90%, etc)
- no alert is sent unless

Actual behavior

- RECOVERY alerts are sent

The system some how consider the first metric collection as a recovery, we must avoid this.

@nepython can you please help me to find the cause of this bug, write a test for it and propose a fix?

nepython:
Closes #125


openwisp_monitoring/monitoring/tests/test_notifications.py
```diff
@@ -78,6 +78,67 @@ def test_general_check_threshold_deferred_not_crossed(self):
...
+            self.create_test_adata()
+            m = Metric.objects.get(name='Memory usage')
+            self.assertEqual(Notification.objects.count(), 1)
+            n = Notification.objects.first()
+            self.assertEqual(n.recipient, admin)
+            self.assertEqual(n.actor, m)
+            self.assertEqual(n.action_object, m.threshold)
+            self.assertEqual(n.level, 'warning')
+
+    def test_disk_metric_threshold_crossed(self):
```
PabloCastellano, line 122: Not sure if we need to test each 'resource' metric, isn't it enough with just one?

nepython: Though it is always better to test all cases but in this case it seems that including test for only one metric is sufficient to test the same for all three :smiley:

README.rst
```diff
@@ -406,6 +406,28 @@ information by performing lookups on the OUI
...
+``Disk usage``, ``Memory usage`` and ``CPU usage``.
+For example, if we want to set thresholds for ``Memory usage`` and ``Disk usage`` metrics to
+``0.90`` and ``0.85`` respectively, then you can use the following configuration:
+
+.. code-block:: python
+
+    OPENWISP_MONITORING_DEVICE_RESOURCES_THRESHOLDS = {
+        'memory': 0.90,
+        'cpu': 0.90,
+        'disk': 0.85,
```
nemesisdesign, line 428: Please don't. We need a generic solution for this as mentioned in https://github.com/openwisp/openwisp-monitoring/issues/106
Remove this logic please.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -83,6 +83,7 @@ def _get_or_create(cls, **kwargs):
             metric = cls.objects.get(**kwargs)
             created = False
         except cls.DoesNotExist:
+            kwargs['is_healthy'] = True
```
nemesisdesign, line 86: This field is NULL by default which means unknown.
Setting it to True is flagging the metric as healthy but it's hiding an underlying issue: it is not a good solution. The broken code is still lying around.

Why are recovery alerts being sent when the metric is created?

That shouldn't happen and I'm not sure if it's an underlying issue that was present before or if it's new, have you found out? Is this happening also with ping metrics?

Please look for a general solution, try also with ping metrics.

nepython: I think it was present from the beginning. It should be solved by the latest iteration (which should be full-proof hopefully :smiley: )

README.rst
nepython: I forgot that there's a separate issue existing which states it more clearly, thanks for the reminding :smile:

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -119,7 +119,7 @@ def content_type_key(self):
         except AttributeError:
             return None
 
-    def check_threshold(self, value, time=None):
+    def check_threshold(self, value, time=None, notify=True):
```
PabloCastellano, line 122: Looks like it should be a local variable instead of a parameter

nepython: I thought upon it that this could be helpful in tests, similar to the `store` parameter in `perform_check`. Wdyt, @PabloCastellano, can it be helpful in testing?

nemesisdesign: is this argument being used at all anywhere? It looks like is not. If that's the case, it should be removed. We must refrain from adding code we don't really need.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -138,6 +138,9 @@ def check_threshold(self, value, time=None):
             notification_type = 'threshold_crossed'
         # ok: returned within threshold limit
         elif not crossed and self.is_healthy in [False, None]:
+            # First metric write and within threshold, do not raise alert
+            if self.is_healthy is None:
+                notify = False
```
nemesisdesign, line 143: this would still trigger the sending of the `threshold_crossed` signal, which will lead to inconsistent behavior:

software consuming that signal will manifest the same issue we're seeing now with alerts being sent unnecessarily.

I would just return after `save()`.

openwisp_monitoring/monitoring/tests/test_notifications.py
```diff
@@ -80,6 +80,29 @@ def test_general_check_threshold_deferred_not_crossed(self):
         self.assertTrue(m.is_healthy)
         self.assertEqual(Notification.objects.count(), 0)
 
+    def test_resources_metric_threshold_deferred_not_crossed(self):
+        # Verify that newly created metrics within threshold limit don't send any notifications
+        self._create_admin()
+        self.create_test_adata()
+        self.assertEqual(Notification.objects.count(), 0)
```
nemesisdesign, line 87: can you please add a test for the `threshold_crossed` signal to ensure it's not emitted in this case?

openwisp_monitoring/monitoring/base/models.py
nepython: The signal `threshold_crossed` and it's receiver are being used in both situations (this can be improved). The [receiver](https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/base/models.py#L233) of this signal also takes care of updating the status for a new device (unknown --> ok). This should be done by a separate method, I think. Without that, device health status might never get updated as can be seen from this [travis build](https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/698984432#L1047).

**Proposal**: Lets's have a separate `threshold_recovery` signal and separate the code in `threshold_crossed` receiver which is taking care of updating status during recoveries currently.

nepython:
**Note**:
I went ahead with this https://github.com/openwisp/openwisp-monitoring/pull/126#discussion_r441006284, since it became necessary after [requested changes](https://github.com/openwisp/openwisp-monitoring/pull/126#discussion_r440488557).

PR #134, opened 2020-06-15
[tests] Optimize unit test runtime speed #69

### LINKED ISSUE #69: Optimize unit test runtime speed
Locate slow tests (tests taking longer than 0.3 secs) and find out if mocking can be used to optimize them.
These can be helpful pointers:
* Checkout [locate slow tests](https://github.com/ninuxorg/nodeshot/blob/master/nodeshot/core/base/tests_speedup.py), it should help to locate slower tests.
* Check `--parallel` provided provided by django for testing (provided that it does not affect test-coverage sharply )
* and mocking whatever possible (but not duplicating the entire logic :sweat_smile: )

**outdated** ~~Just for the record it took~~
```C
Ran 142 tests in 37.713s
```
on my latest travis-build

nepython:
Closes #69


openwisp_monitoring/device/tests/__init__.py
```diff
@@ -62,8 +62,9 @@ def create_test_adata(self):
             m = Metric.objects.get(key=ifname, field_name='clients', object_id=d.pk)
             points = m.read(limit=10, order='time DESC')
             self.assertEqual(len(points), len(iface['wireless']['clients']))
+        return dd
 
-    def _create_multiple_measurements(self, create=True):
+    def _create_multiple_measurements(self, create=True, double=False):
```
nemesisdesign, line 67: I have a doubt regarding the naming of this parameter, what does `double` mean?

openwisp_monitoring/measure_speed.py
```diff
@@ -0,0 +1,51 @@
+# Temporarily added to identify slow tests
```
nemesisdesign, line 1: rename this file to `tests.py` for clarity (even if it's temporary, is good to use standard file names).

openwisp_monitoring/measure_speed.py
```diff
@@ -0,0 +1,51 @@
...
+        for name, elapsed in self.test_timings:
+            if elapsed > self.slow_test_threshold:
+                slow_tests_counter += 1
+                name, module = name.split()
+                if module != self._module:
+                    self._module = module
+                    print(f'\033[33;1m{module}\033[0m')
+                color = '\033[31;1m' if elapsed > 1 else '\033[32;1m'
+                print(f'  ({color}{elapsed:.2f}s\033[0m) {name}\033[0m')
+        print(f'\n\033[1mTotal slow tests detected: {slow_tests_counter}\033[0m')
```
nemesisdesign, line 41: This works, but is quite unreadable. Which will not be good for future maintainers.

Why don't you make a function for outputting colored text?

Pseudo code example:

```python
def colored_text(text, color):
    colors = {
       'green': [,
    }
    assert color in colors
    color_code = colors[color]
    return f'<begin-code><text><end-code>'
```

Your future you (and future maintainers) will thank you for this.

runtests.py
```diff
@@ -9,12 +9,20 @@
...
+    from django.conf import settings
 
     args = sys.argv
     args.insert(1, 'test')
     if not os.environ.get('SAMPLE_APP', False):
+        # Temporarily added to identify slow tests
+        setattr(
+            settings,
+            'TEST_RUNNER',
+            'openwisp_monitoring.measure_speed.TimeLoggingTestRunner',
```
nemesisdesign, line 21: why is it only enabled in this part of the if and not in the other?

BTW, maybe we can put this in the settings.py of the test project, it should be easier to find it, here is kind of hidden and can generate confusion.

nepython:
One slow test `test_check_ping_params` has been left unlocked on purpose because we need to verify that `_command` method is working as expected atleast in one test and for code coverage.

nepython:
This PR is now ready for review :
* There are 2 very slow tests `test_check_ping_params`, `test_device_admin`. The first one has been left unmocked as explained above https://github.com/openwisp/openwisp-monitoring/pull/134#issuecomment-644997755.
* I have reduced the number of API requests being made wherever possible.
* `Ping _command` has been mocked wherever possible.
* I am not sure why but `test_admin` (device_monitoring) and `test_notifications` (monitoring) where left out from SAMPLE_APP tests, This explains the major difference in sample_app tests and actual module's tests, earlier. I have included them both (`test_admin`, `test_notifications`) in the sample app tests
* Merged three tests in `test_notifications` which were testing same feature in 3 different conditions (the increase in speed is approx. 0.6s on travis). Please check these specially to unsure I didn't break them in doing so :/
* Overall testing runtime is almost halved now :)

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -62,32 +69,40 @@ def create_test_adata(self):
             m = Metric.objects.get(key=ifname, field_name='clients', object_id=d.pk)
             points = m.read(limit=10, order='time DESC')
             self.assertEqual(len(points), len(iface['wireless']['clients']))
+        return dd
 
-    def _create_multiple_measurements(self, create=True):
+    def _create_multiple_measurements(self, create=True, no_resources=False, count=4):
```
PabloCastellano, line 74: It would be better to have this kind of refactors in a different PR to make the reviewing process easier

openwisp_monitoring/check/tests/test_ping.py
```diff
@@ -24,15 +25,13 @@ class TestPing(TestDeviceMonitoringMixin, TransactionTestCase):
         '',
         bytes("fping: option requires an argument -- 'z'", encoding='utf8'),
     )
-    _FPING_OUTPUT = (
+    _FPING_UNREACHABLE = (
```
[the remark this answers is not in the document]
nepython: I had moved `_FPING_OUTPUT` to `__init__.py` since it was required by three Test classes. This need didn't arose in the case of `_FPING_UNREACHABLE`, hence I hadn't moved it but I can move it.

openwisp_monitoring/device/tests/test_recovery.py
```diff
@@ -27,11 +29,14 @@ def _create_device_monitoring(self):
...
 
-    def test_trigger_device_recovery_task(self):
+    @patch.object(Ping, '_command', return_value=_FPING_OUTPUT)
+    def test_trigger_device_recovery_task(self, mocked_method):
         d = self._create_device(organization=self._create_org())
         d.management_ip = '10.40.0.5'
         d.save()
         data = self._data()
+        del data['resources']
+        del data['interfaces']
```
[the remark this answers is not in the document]
nepython: The test is supposed to test whether device recovery task is being triggered spontaneously. In that case, we could save the requests being made to create `clients`, `traffic`, `resources` metric, charts, alertsettings. It speeds up the tests as we are testing only what we want to test, specifically. Nothing less, nothing more :)

tests/openwisp2/sample_check/tests.py
```diff
@@ -18,15 +22,13 @@ class TestPing(BaseTestPing, TestDeviceMonitoringMixin, TransactionTestCase):
...
+    def test_last_called(self, mocked_method):
         device = self._create_device(organization=self._create_org())
         # will ping localhost
         device.management_ip = '127.0.0.1'
         check = Check(
-            name='Ping check',
-            check=self._PING,
-            content_object=device,
-            params={'count': 2, 'interval': 10, 'bytes': 10, 'timeout': 50},
+            name='Ping check', check=self._PING, content_object=device, params={},
```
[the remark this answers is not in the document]
nepython: Yes, you are correct.

openwisp_monitoring/device/tests/__init__.py
nepython: If it wouldn't had increased the speed, I would have desisted from making the change. I have compared the speeds previously with and without `resources` `metrics`, `charts`, `alertsettings` being created.
Increase in individual test's speed was slightly more than 0.1s. If `0.3s` is the threshold then `0.1s` counts according to me. Anyways, for the record, the fastest travis build when there was no changes in `device` tests was slightly less than **29s**. So, I think reducing requests is increasing speed by approximately **5s** overall.
Also, this parameter is only being used when there is no need for resources metrics at all in the test. For example, `test_200_traffic_counter_incremented`, `test_200_traffic_counter_reset`, etc.

If you want to compare the speed increase in slow tests:~
Speed log before changes : [travis-build](https://travis-ci.com/github/nepython/openwisp-monitoring/jobs/348784253#L1072)
Speed log after changes : [travis-build](https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/699070313#L1071)

openwisp_monitoring/device/tests/test_recovery.py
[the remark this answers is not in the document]
nemesisdesign: add a brief comment to let maintainers know why those `del` statements are there please

openwisp_monitoring/tests.py
```diff
@@ -0,0 +1,67 @@
+# Temporarily added to identify slow tests
+
+from time import time
+from unittest import TextTestResult
+
+from django.test.runner import DiscoverRunner
+
+
+class TimeLoggingTestResult(TextTestResult):
```
nemesisdesign, line 9: you can now import this from openwisp-utils master

nepython:
Done :+1:

nepython:
> It's good to merge, I tried to fix conflicts with the current master but something went wrong with a test which is failing, can you fix it please? Also squash the commits this time please.

The test module was named incorrectly. I have corrected it `test_device_notifcations` --> `test_device_notifications`.
I have made 2 small changes in sample_app, relocated `TestDeviceNotifications` from `sample_monitoring` --> `sample_device_monitoring`
Added `TestMonitoringNotifications` in `sample_monitoring` tests.

Issue #131, opened 2020-06-15
[bug] All logs are being duplicated


nepython:
On the master branch of openwisp-monitoring, I found that logs are duplicated as can be seen below. I am observing this for the first time.

![Screenshot from 2020-06-15 18-30-20](https://user-images.githubusercontent.com/54471024/84661333-bedde100-af09-11ea-90bd-9ca3441e0734.png)

PabloCastellano:
This change might be related https://github.com/openwisp/openwisp-monitoring/commit/a51467c1effc3f66ae63431ad2d498c485343129

Since we have a root logger now, we probably have to stop propagating from the other loggers:

```
    'loggers': {
        'django': {
            'handlers': ['console'],
            'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
            'propagate': False,
        },
```

nepython:
Closed by #137

PR #137, opened 2020-06-18
[fix] Avoid log propagation when using root logger #131

nepython:
I have checked that celery logs aren't being duplicated but don't know for `py.warnings` log but it should be safe to have `propagate` disabled for it.


nepython:
I tried to write a test for this bug but could not reproduce it. There was some interference from django TestCase for `django logger` but even after overcoming them, I could record only a single log with and without the fix. The fix is simple and works but a test to ensure this does not repeat might had been better :/

Issue #66, opened 2020-05-05
[enhancement] Add support for another timeseries DB


nepython:
Add support for another timeseries DB, it could be `prometheus`, `opentsdb` or another good option if there's any.
The decision regarding which one to choose must be backed up by convincing research.

Check the discussion [here](https://groups.google.com/forum/#!topic/openwisp/tinsyzJt3BE)

Points to remember:
- we need to implement the current features and make them both work for `Prometheus` and `Influxdb`
- the new timeseries DB that we will support needs to provide an **open source solution for horizontal scaling**, because influxdb does not provide it (has a paid cloud service instead)

**New Findings**:
- There are many timeseries db which are built on top of prometheus and cover up for many essential features missing in prometheus. The best amongst these are [Victoria Metrics](https://github.com/VictoriaMetrics/VictoriaMetrics), [Cortex](https://github.com/cortexproject/cortex) and [Thanos](https://github.com/thanos-io/thanos).
All three projects took off in past 4-5 years, are currently active and open-source (Victoria went open-source [last year](https://blog.usejournal.com/open-sourcing-victoriametrics-f31e34485c2b)).
* Here's a quick comparision between [Cortex and Thanos](https://grafana.com/blog/2019/11/21/promcon-recap-two-households-both-alike-in-dignity-cortex-and-thanos/).
* After going through multiple reviews/ docs/ benchmark comparisions, Victoria Metrics seems impressive. (Check it's [benchmark comparision with influxdb](https://medium.com/@valyala/insert-benchmarks-with-inch-influxdb-vs-victoriametrics-e31a41ae2893))

**NOTE**: This requires #65 to be complete

nepython:
Found this https://github.com/prometheus/influxdb_exporter. It might make things easier. :smile:

PabloCastellano:
We should discuss how to properly setup the metrics exporter, as I have different ideas but I don't know what is the best at the moment.

1. Creating a exporter to be installed in the device is the standard, however due to device limited resources sometimes it might not be the best idea
2. Using something like the pushgateway

There are already exporters for cpu, load, memory, network, etc. They are usually written in Golang, resulting in a quite big executable to put on small devices. So we would have to implement our own ones... in Lua? Doesn't sound very exciting reinventing the wheel in Lua.

By using pushgateway we can reuse the Lua code we have already written.

These are my views, I will try to get some time to think about it so that we can properly discuss it.

nemesisdesign:
[...]

PR #142, opened 2020-06-22
[metrics] Merge traffic metrics #133

### LINKED ISSUE #133: [metrics] rx_bytes vs tx_bytes
Right now two `Metric` objects are being created for each traffic measurements.
I wonder if we can avoid this and have just one metric created. It should be possible.

nepython:
Closes #133
* I have created a migration script to delete traffic metrics (just in case there are too many interfaces so to avoid doing it manually)
* The new metric created automatically is able to communicate with the old measurement (thus the new chart is able to display old data, no data loss I mean :) )

### DIFF
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -43,7 +43,6 @@ class DeviceMetricView(GenericAPIView):
     serializer_class = serializers.Serializer
     permission_classes = [DevicePermission]
     schema = schema
-    statistics_stored = ['rx_bytes', 'tx_bytes']
 
     def get(self, request, pk):
         # ensure valid UUID
@@ -191,19 +190,29 @@ class DeviceMetricView(GenericAPIView):
         ct = ContentType.objects.get_for_model(Device)
         for interface in data.get('interfaces', []):
             ifname = interface['name']
-            for key, value in interface.get('statistics', {}).items():
-                if key not in self.statistics_stored:
-                    continue
-                name = '{0} {1}'.format(ifname, key)
+            ifstats = interface.get('statistics', {})
+            # Explicitly stated None to avoid skipping in case the stats are zero
+            if (
+                ifstats.get('rx_bytes') is not None
+                and ifstats.get('rx_bytes') is not None
+            ):
+                field_value = self._calculate_increment(
+                    ifname, 'rx_bytes', ifstats['rx_bytes']
+                )
+                extra_values = {
+                    'tx_bytes': self._calculate_increment(
+                        ifname, 'tx_bytes', ifstats['tx_bytes']
+                    )
+                }
+                name = f'{ifname} traffic'
                 metric, created = Metric._get_or_create(
                     object_id=pk,
                     content_type=ct,
-                    configuration=f'traffic_{key}',
+                    configuration='traffic',
                     name=name,
                     key=ifname,
                 )
-                increment = self._calculate_increment(ifname, key, value)
-                metric.write(increment)
+                metric.write(field_value, extra_values=extra_values)
                 if created:
                     self._create_traffic_chart(metric)
             try:
@@ -320,10 +329,7 @@ class DeviceMetricView(GenericAPIView):
         """
         create "traffic (GB)" chart
         """
-        if (
-            metric.field_name != 'tx_bytes'
-            or 'traffic' not in monitoring_settings.AUTO_CHARTS
-        ):
+        if 'traffic' not in monitoring_settings.AUTO_CHARTS:
             return
         chart = Chart(metric=metric, configuration='traffic')
         chart.full_clean()
--- a/openwisp_monitoring/device/tests/__init__.py
+++ b/openwisp_monitoring/device/tests/__init__.py
@@ -51,23 +51,19 @@ class DeviceMonitoringTestCase(TestDeviceMonitoringMixin, TestCase):
         dd = DeviceData(pk=d.pk)
         self.assertDictEqual(dd.data, data)
         if no_resources:
-            metric_count, chart_count = 6, 4
+            metric_count, chart_count = 4, 4
         else:
-            metric_count, chart_count = 9, 7
+            metric_count, chart_count = 7, 7
         self.assertEqual(Metric.objects.count(), metric_count)
         self.assertEqual(Chart.objects.count(), chart_count)
         if_dict = {'wlan0': data['interfaces'][0], 'wlan1': data['interfaces'][1]}
         for ifname in ['wlan0', 'wlan1']:
             iface = if_dict[ifname]
-            for field_name in ['rx_bytes', 'tx_bytes']:
-                m = Metric.objects.get(
-                    key=ifname, field_name=field_name, object_id=d.pk
-                )
-                points = m.read(limit=10, order='-time')
-                self.assertEqual(len(points), 1)
-                self.assertEqual(
-                    points[0][m.field_name], iface['statistics'][field_name]
-                )
+            m = Metric.objects.get(key=ifname, field_name='rx_bytes', object_id=d.pk)
+            points = m.read(limit=10, order='-time', extra_fields=['tx_bytes'])
+            self.assertEqual(len(points), 1)
+            for field in ['rx_bytes', 'tx_bytes']:
+                self.assertEqual(points[0][field], iface['statistics'][field])
             m = Metric.objects.get(key=ifname, field_name='clients', object_id=d.pk)
             points = m.read(limit=10, order='-time')
             self.assertEqual(len(points), len(iface['wireless']['clients']))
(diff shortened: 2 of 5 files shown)


openwisp_monitoring/monitoring/migrations/0013_delete_traffic_metrics.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def delete_traffic_metrics(apps, schema_editor):
+    Metric = apps.get_model('monitoring', 'Metric')
+    Metric.objects.filter(field_name='rx_bytes').delete()
```
nemesisdesign, line 6: can we rename these instead of deleting them?

We have threshold configuration set for some metrics, and if we delete these we'll lose them.

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -51,23 +51,19 @@ def create_test_adata(self, no_resources=False):
...
-                m = Metric.objects.get(
-                    key=ifname, field_name=field_name, object_id=d.pk
-                )
-                points = m.read(limit=10, order='time DESC')
-                self.assertEqual(len(points), 1)
-                self.assertEqual(
-                    points[0][m.field_name], iface['statistics'][field_name]
-                )
+            m = Metric.objects.get(key=ifname, field_name='rx_bytes', object_id=d.pk)
+            points = m.read(limit=10, order='time DESC', extra_fields=['tx_bytes'])
```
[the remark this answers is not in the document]
nepython: Yes, I think it's InfluxDB specific but we might need something similar for the `read` method.
I mean we would definitely need to read metric data using params similar to `limit` and `order` for the other timeseries db to create required query.


* I think we can rename these params.
* have a list of keywords like [time, DESC, ASC, value, etc.] for order, and while working on #66, create queries for the other db based upon it.

Both are best done when working on #66, I think.

openwisp_monitoring/device/api/views.py
```diff
@@ -191,15 +191,26 @@ def _write(self, request, pk):
         ct = ContentType.objects.get_for_model(Device)
         for interface in data.get('interfaces', []):
             ifname = interface['name']
+            extra_values = {}
+            field = None
             for key, value in interface.get('statistics', {}).items():
```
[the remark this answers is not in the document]
nepython: I thought of doing so but didn't do so due to `statistics_stored`, it can have more fields in the future, that's what I assumed but I can proceed now :smiley:

openwisp_monitoring/monitoring/migrations/0013_delete_traffic_metrics.py
nepython: Is AlertSettings being used along with traffic metrics? (it's not the case atleast by default).
So if AlertSettings have been created for traffic metrics, there shall be a small problem. Currently, AlertSettings defaults to `field_name` for the value and there is no way to pass `extra_fields` to AlertSettings I think. A limitation which can be overcomed, so let me know if I should go ahead with it.

nemesisdesign: There's no default alert settings but they have been added in some cases.
Moreover, if a device is temporarily down, the metrics and graphs will not show up after this migration, which will leave the user thinking there's a bug.
Whenever we introduce changes, we must make sure it doesn't affect negatively the people who are already using the software.
You can manipulate the existing metrics by looping over them right?
And the ones that are not needed anymore can be deleted as you're doing here. It shouldn't be a big change, isn't it?

openwisp_monitoring/device/tests/__init__.py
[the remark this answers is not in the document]
nemesisdesign: We could do it kind like django does it: `order='time'` would translate in influxdb to `time ASC` while `order='-time'` would translate to `time DESC`. `limit` can stay. Regarding `extra_fields`, would that be addressed by the other PR which introduces metric configuration?

Rather than working on it in #66, I would start opening a new PR for this change, which being smaller it will be faster for me and @PabloCastellano to review and get merged.

openwisp_monitoring/device/api/views.py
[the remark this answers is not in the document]
nemesisdesign: yes that was the initial idea but turned out to be the usual brilliant idea that never materializes and only adds unneeded complexity (yes, I also do this mistake sometimes) so we can remove it :+1:

openwisp_monitoring/monitoring/migrations/0013_delete_traffic_metrics.py
nepython: The small problem here is, there are 2 metrics, 2 AlertSettings. Now we can go ahead and merge the two metrics but we can't do so with AlertSettings (in the current form). The reason being AlertSettings currently supports only one threshold value. Since, metric supports multiple fields, it would be better if AlertSettings can also do the same.

**Proposal**: Let's have a TextField called additional_thresholds (this field shall hold the name and value of the extra_field's threshold as a list of dict).
```python
m.alertsettings.additional_fields = [{'tx_bytes': <threshold_value>}] # rx_bytes is covered by default, being the field_name
```
We can modify the change form to make this UI friendly and we aren't modifying the current approach, so it should be backward compatible. Metric configuration PR won't be much helpful here but yes we can make use of `metric.related_fields` to verify the fields entered in `alertsettings.additional_fields` exist. Let me know, if I should go ahead with this solution.

openwisp_monitoring/device/tests/__init__.py
```diff
@@ -95,8 +91,8 @@ def _create_multiple_measurements(self, create=True, no_resources=False, count=4
         data3 = deepcopy(data)
         data3['interfaces'][0]['statistics']['rx_bytes'] = 500000000
         data3['interfaces'][0]['statistics']['tx_bytes'] = 300000000
-        data3['interfaces'][1]['statistics']['rx_bytes'] = 0
-        data3['interfaces'][1]['statistics']['tx_bytes'] = 0
+        data3['interfaces'][1]['statistics']['rx_bytes'] = 1
+        data3['interfaces'][1]['statistics']['tx_bytes'] = 1
```
PabloCastellano, line 95: Why did these values change?

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -151,18 +143,15 @@ def test_200_multiple_measurements(self):
...
+        # TODO: Find why influxdb returns 3 points without this change
+        expected = [999999999, 1, 1999997725, 2275]
         for i, point in enumerate(points):
             self.assertEqual(point['rx_bytes'], expected[i])
-        # wlan1 tx_bytes
-        m = Metric.objects.get(key='wlan1', field_name='tx_bytes', object_id=dd.pk)
-        points = m.read(limit=10, order='time DESC')
-        self.assertEqual(len(points), 4)
-        expected = [500000000, 0, 999999174, 826]
+        expected = [499999999, 1, 999999174, 826]
```
PabloCastellano, line 154: Why did it change from 500000000 to 499999999?

openwisp_monitoring/monitoring/migrations/0013_delete_traffic_metrics.py
PabloCastellano: I like it but I think it's less confusing if we add an extra metric_name field and keep having one AlertSettings per threshold. Otherwise you need also to define additional_operators, right?

This way you can get a notification only when rx_bytes goes above X and you don't really mind about tx_bytes.

openwisp_monitoring/device/tests/__init__.py
nepython: I had added a comment above it stating that I am working on it. I wasn't able to pinpoint the issue then. The problem was with the views, fixed it and added a comment for the same :smiley:.

openwisp_monitoring/monitoring/migrations/0013_delete_traffic_metrics.py
nepython: @PabloCastellano, @nemesisdesign let me know if the current iteration implements your suggestions well.

nepython:
Coverage is decreasing as metrics got merged and the relevant code is thus reduced.

openwisp_monitoring/monitoring/migrations/0013_merge_traffic_metrics.py
```diff
@@ -0,0 +1,23 @@
+from django.db import migrations
+
+
+def merge_traffic_metrics(apps, schema_editor):
+    Metric = apps.get_model('monitoring', 'Metric')
+    rx_metrics = Metric.objects.filter(field_name='rx_bytes')
+    for metric in rx_metrics:
+        metric.name = f'{metric.name.split()[0]} traffic'
+        metric.save()
```
PabloCastellano, line 9: I think it would be useful to provide feedback to the user:

```
new_name = f'{metric.name.split()[0]} traffic'
print('Renamed metric "{}" to "{}"'.format(metric.name, new_name))
metric.name = new_name
```

nepython:
Done :+1:

nepython:
> @nepython I have one doubt here and in general after the metric configurations have been introduced.
> 
> What happens to metrics that were being stored previously? Do they keep working right?
> Please ensure this.

That is a migration script to fill configuration similar to chart configuration one!

PR #146, opened 2020-06-24
[timeseries] Refactor order parameter to only allow 'time' / '-time'

nepython:
Please refer to https://github.com/openwisp/openwisp-monitoring/pull/142#discussion_r444553485

### DIFF
--- a/openwisp_monitoring/db/backends/influxdb/client.py
+++ b/openwisp_monitoring/db/backends/influxdb/client.py
@@ -146,6 +146,16 @@ class DatabaseClient(object):
             conditions = 'WHERE %s' % ' AND '.join(conditions)
             q = f'{q} {conditions}'
         if order:
+            # InfluxDB only allows ordering results by time
+            if order == 'time':
+                order = 'time ASC'
+            elif order == '-time':
+                order = 'time DESC'
+            else:
+                raise self.client_error(
+                    f'Invalid order "{order}" passed.\nYou may pass "time" / "-time" to get '
+                    'result sorted in ascending /descending order respectively.'
+                )
             q = f'{q} ORDER BY {order}'
         if limit:
             q = f'{q} LIMIT {limit}'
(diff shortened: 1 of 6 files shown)


openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -146,6 +146,15 @@ def read(self, key, fields, tags, **kwargs):
...
             q = f'{q} {conditions}'
         if order:
+            if order == 'time':
+                order = 'time ASC'
+            elif order == '-time':
+                order = 'time DESC'
+            else:
+                raise self.client_error(
+                    f'Invalid order "{order}" passed.\nYou may pass "time" / "-time" to get '
+                    'result sorted in ascending /descending order respectively.'
```
PabloCastellano, line 156: What  if you want to order by a different value?

I think the logic here is checking if `order` starts with`-` or not, and then constructing `{order} ASC/DESC`

PabloCastellano: https://github.com/openwisp/openwisp-monitoring/blob/1825d7788a07be6a2afdf4eeeb1c29c8bd571e1f/openwisp_monitoring/device/tests/test_api.py#L401

nepython: Anything other than time is not supported for `influxdb`, https://stackoverflow.com/questions/58733165/is-there-a-way-to-sort-influxdb-by-a-field-value-other-than-time.
I also didn't think there's a requirement for it, currently. Let me know if I am missing anything.

PabloCastellano: As you pointed out in the gitter chat, influxdb only allows ordering results by time, so it's ok - https://stackoverflow.com/questions/58733165/is-there-a-way-to-sort-influxdb-by-a-field-value-other-than-time

Please add the following comment so that future developers can save a search about it ;)

```
# As of now, influxdb only allows to order by time
```

nepython:
Done :+1:

PR #147, opened 2020-06-24
[fix] Fixed broken chart js in metric admin change form #110

### LINKED ISSUE #110: [bug] admin: Metric > Graphs: chart JS is broken
From the browser JS console:
```
plotly-cartesian.min.js:21 Uncaught Error: DOM element provided is null or undefined
    at Object.getGraphDiv (plotly-cartesian.min.js:21)
    at Object.r.newPlot (plotly-cartesian.min.js:21)
    at window.createGraph (graph.js:134)
    at (index):478
```

nepython:
Closes #110
**PS**: There is a better and more full proof approach but it depends on the query below.

### DIFF
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -407,7 +407,10 @@ class AbstractChart(TimeStampedEditableModel):
 
     def json(self, time=DEFAULT_TIME, **kwargs):
         try:
-            return json.dumps(self.read(time=time), **kwargs)
+            # unit needs to be passed for chart_inline
+            data = self.read(time=time)
+            data.update({'unit': self.unit})
+            return json.dumps(data, **kwargs)
         except KeyError:
             # TODO: this should be improved
             pass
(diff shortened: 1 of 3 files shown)


nepython:
The actual reason for the bug is https://github.com/openwisp/openwisp-monitoring/blob/238a17bcfebef619c5c125a130abcb076d1a9c4b/openwisp_monitoring/monitoring/templates/admin/chart_inline.html#L24-L31
I can't note any difference if it's removed except for the fact that error isn't raised in the browser console.
@nemesisdesign, can you please let me know, what is the code supposed to do and is it required?

nepython:
>@nepython Instead of just ignoring the error, can you investigate why the chart is not shown in the Metric details page? I think what we are really expecting is seeing the chart displayed there too.

I did mention there's a more full proof approach, just wanted this confirmation that is a chart expected there or not :smiley:
Screenshot from 2020-06-30 22-47-35
It's fixed now.

nepython:
I noted after fixing this that there are some other issues (like chart color).

PR #148, opened 2020-06-25
[feature] Add Monitoring Checks section in Device Admin #53

### LINKED ISSUE #53: [feature] Add Monitoring Checks section in Device Admin
Add a `Monitoring Checks` section in the device admin which allows to see what **checks** are available for a `device` and add new ones if needed.

nepython:
Closes #53 
Screenshot from 2020-06-25 19-07-10

**PS** I have added check_inline.css as the default font-size of `Ping (Device: 08:00:27:04:8E:1C)` given by GenericTabularInline seems a bit small as can be seen in the pic above.

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -1,6 +1,11 @@
 import uuid
 
 from django.contrib import admin
+from django.contrib.contenttypes.admin import GenericStackedInline
+from django.contrib.contenttypes.forms import BaseGenericInlineFormSet
+from django.contrib.contenttypes.models import ContentType
+from django.db.models import TextField
+from django.forms import Textarea
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
@@ -9,6 +14,7 @@ from swapper import load_model
 
 from openwisp_controller.config.admin import DeviceAdmin as BaseDeviceAdmin
 from openwisp_controller.config.models import Device
+from openwisp_utils.admin import TimeReadonlyAdminMixin
 
 from ..monitoring.admin import MetricAdmin
 from . import settings as app_settings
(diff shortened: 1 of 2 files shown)


nepython:
> Did you test adding a new check?

Unfortunately, I didn't, since I had priorly done something similar with no glitch :sweat:.
**GenericTabularInline** takes care of auto filling `contenttype` fields but this takes place while saving the model.
In case of check, the above failure was occurring due to validation before saving the model.
https://github.com/openwisp/openwisp-monitoring/blob/23f3105c539950d1f547f5dcefcda1228c09bcf1/openwisp_monitoring/check/classes/ping.py#L61-L66
It's solved now by filling these fields prior validation by over riding default formset's full clean method and I have added a test for the same :)

openwisp_monitoring/device/static/check/css/check_inline.css
```diff
@@ -0,0 +1,3 @@
+.inline-group .tabular td.original p {
+    font-size: 14px;
+}
```
nemesisdesign, line 3: I would remove this because this will be in our way when we'll do a general redesign of the theme and use different font-sizes (we could use em instead of px for example).

nepython: It's displayed properly without this after using `GenericStackedInline`, so removed it :smiley:

nepython:
>* let's not use the tabular form, it won't work well on narrow screens

:+1: 
>* make check type readonly for existing objects

I will need to look into this, there's already a 9 year old [django ticket](https://code.djangoproject.com/ticket/15602).
>* order of fields should be: name, type, active, parameters, created, modified

Ok
>* I think we can get rid of description or rename it to notes, I'm inclined to just remove it

It's not displayed now.

Issue #67, opened 2020-05-05
[enhancement] Add a way to disable alerts


nepython:

NoumbissiValere:
@nepython  please  what do you mean by alert? Do you mean notifications or what?

nemesisdesign:
Closed by #150

PR #150, opened 2020-06-29
[feature] Add an Alert Settings section in device admin #52

### LINKED ISSUE #52: [feature] Add an Alert Settings section in device admin
Add an `Alert Settings` section in the device admin which allows to see what alerts are available for a `device` and which also allow to modify the `threshold` or **disable/enable** the alarm.

nepython:
### Blocked by #162
Closes #52, #67
This is how it looks, right now:
Screenshot from 2020-07-07 22-01-16


nepython:
> Does it look like the following?

Yes with an external dependency `django-nested-admin`, @PabloCastellano had requested to create a draft PR and display all possible fields related to `alertsettings` without the dependency (to stay close to the UNIX philosophy, I assume).

I can display only the info as displayed in the pic (https://github.com/openwisp/openwisp-monitoring/pull/150#issue-441333083) without the dependency. There is a hacky approach you create your own form/formset/factory/template/css/js (in short it would be repeating the code of `nested inline` (for which there are multiple options). It appeared an overkill to me :sweat_smile: )

nepython:
I had to put `metric_send_alerts` migration above `merge_traffic_metrics` as there was conflict and exception of `send_alerts` field not existing was being raised which is expected.

I tried adding `send_alerts` field to `Alert Settings` model instead of `Metric`, I can't figure out why but the corresponding column in `monitoring_alertsettings` is not created despite migration showing `OK` status. Though everything is working fine with `Metric` model, learning what actually went wrong with `AlertSettings` would be beneficial in the longer run, I think.

nepython:
> Why are notifications being shown? That shouldn't happen.

I though it should, no worries removed it.
> Also, we should not show metrics that do not have related alert settings.
> Can you change the queryset so that only metrics which have related settings are added?
> Eg:
> 
> ```python
> Metric.objects.filter(alertsettings__isnull=False)
> ```

Done, thanks.
> Also, the alert settings inline should not be shown for new devices, we did something similar in the firmware upgrader module recently: https://github.com/openwisp/openwisp-firmware-upgrader/blob/d4f31d46e64af82e936f6917854420731b26436a/openwisp_firmware_upgrader/admin.py#L291-L307
> We will have to call inlines = super(DeviceAdmin, self).get_inlines() in the first line though, instead of doing inlines = self.inlines to ensure the two modules will not override each other.
> In this case, the inline should be removed for new objects (obj._state.adding).

Didn't knew about this, it's implemented now though inline would also be skipped when `obj=None`, since I was getting that when adding a new `Device`.
> I don't understand your reasoning: we renamed Threshold to Alert Settings exactly because we wanted to highlight that the alert settings are concentrated there and now we add a send_alert field on Metric, which is used for the purpose of mapping timeseries DB metrics to the data in django?
> 
> It doesn't make any sense to me.
> 
> I think this should be moved to AlertSettings and should be simply called is_active.
> Use "Alerts enabled" as the field label, add an help_text like "Whether alerts are enabled for this metric, uncheck to disable the alert without deleting the alert settings.".

I was trying to say that when running the migration (for adding the field to `AlertSettings` model) the corresponding SQL script was not being run by django thus no changes were being made in the database (I don't know the cause). I have added a `RunSQL` in the migration, the same one that was supposed to be executed by the automatically generated migration script (and everything works fine but I would like to dig for the cause as to why the migration script was unable to add a field to `AlertSettings` model in the first place.)

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -26,9 +26,12 @@ def test_device_admin(self):
         r = self.client.get(url)
         self.assertContains(r, '<h2>Status</h2>')
         self.assertContains(r, '<h2>Charts</h2>')
+        self.assertContains(r, '<h2>\n        Alert Settings\n    </h2>')
```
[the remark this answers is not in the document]
nepython: I am not sure, I guess we can simply test whether the response contains `AlertSettings` (might not be a full proof approach since if in future more tabs are added which also contain the term `AlertSettings` then this might be obsolete.)

nepython:
> Make sure this field is show as first thing in the alert settings, when disabled the other fields should become readonly (readonly attribute and there's also a readonly css class to use) to improve UX.

@nemesisdesign, I have added a toggle which will hide all the fields of `AlertSettings` besides `Alerts enabled` when it's disabled, as can be seen below

Screenshot from 2020-07-08 01-41-01
The above seemed more better to me, since if the user is not interested in the alert then simply don't display the fields.

openwisp_monitoring/device/admin.py
```diff
@@ -90,9 +125,26 @@ def get_form(self, request, obj=None, **kwargs):
         return super().get_form(request, obj, **kwargs)
 
 
-DeviceAdmin.inlines.append(CheckInline)
+def device_admin_get_inlines(self, request, obj):
+    # copy the list to avoid modifying the original data structure
+    inlines = super(DeviceAdmin, self).get_inlines(request, obj)
```
nemesisdesign, line 130: are you sure this copies the list?

this is safer:

```python
inlines = list(super(DeviceAdmin, self).get_inlines(request, obj))
```

openwisp_monitoring/monitoring/migrations/0015_alertsettings_is_active.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('monitoring', '0014_merge_traffic_metrics_and_fill_metric_config'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
nemesisdesign, line 13: why are you using RunSQL?? I don't understand why we can't use a regular migration like all the other ones? Please explain.

tests/openwisp2/sample_monitoring/migrations/0007_alertsettings_is_active.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sample_monitoring', '0006_metric_configuration'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
nemesisdesign, line 13: same here

openwisp_monitoring/monitoring/migrations/0015_alertsettings_is_active.py
nepython: I stated it above https://github.com/openwisp/openwisp-monitoring/pull/150#issuecomment-654786288 (last para), maybe I should use inline comments :sweat_smile:

openwisp_monitoring/device/admin.py
```diff
@@ -5,20 +5,68 @@
...
+        return False
+
+
+class AlertSettingsInline(TimeReadonlyAdminMixin, NestedStackedInline):
+    model = AlertSettings
+    extra = 0
+    max_num = 0
+    sortable_options = dict()
 
+    def has_add_permission(self, request, obj=None):
```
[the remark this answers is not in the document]
nepython: It's `inline` not `ModelAdmin`. Confirmed that `obj` parameter is passed.
https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#django.contrib.admin.InlineModelAdmin.has_add_permission

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -464,6 +465,15 @@ class AbstractAlertSettings(TimeStampedEditableModel):
         '({1} days)'.format(_SECONDS_MAX, int(_SECONDS_MAX / 60 / 60 / 24))
     )
     _ALERTSETTINGS_OPERATORS = (('<', _('less than')), ('>', _('greater than')))
+    is_active = models.BooleanField(
+        _('Alerts enabled'),
+        default=True,
+        db_index=True,
```
nemesisdesign, line 471: I don't think we need `db_index=True` here, why did you add it?

openwisp_monitoring/device/static/monitoring/js/alertsettings_inline.js
```diff
@@ -0,0 +1,12 @@
+(function ($) {
+    $(function () {
+        checks = $('[id^="id_monitoring-metric-content_type-object_id"][id $="alertsettings-0-is_active"]');
+        checks.parentsUntil('fieldset.djn-fieldset.module').siblings('h2').css('display', 'none');
+        checks.closest('fieldset').siblings('h3').css('display', 'none');
```
nemesisdesign, line 5: keep in mind is always better to use CSS to do these kind of things

Issue #55, opened 2020-04-27
[feature] Add a Check which inspects links and nodes & creates relevant charts for the same


nepython:
The issue can be broken down as:
### 1. Add check, metric and a relative alert settings
Add a `check`, `metric` and a `relative alert settings` that will be available only if `openwisp-network-topology` is in **INSTALLED_APPS** (created automatically for every topology, it must take into account existing **topologies** by using a data migration) that is run periodically (by default every 5 minutes).
### 2. The Check should:-
* write to the timeseries DB the number of links that are down, the number of links that are up, the total number of nodes and the number of isolated nodes (nodes without links).
* check if the topology has **X %** or more of its links in **down** state, in which case it should send an alert. **X** should be configurable with a django setting defaulting to **50%**.
### 3. Add Relevant charts
Two charts should be created: one which shows the amount of links (one line for the **up** state and one line for the **down** state) and another chart for the amount of nodes (one line for total nodes and one line for isolated nodes).
### 4. Documentation
Document this feature and add illustrative pictures of the `charts`.

nepython:
I was able to create a topology and update it's strategy thus creating relevant links and nodes.
I have few **queries** below as I don't have a deep understanding of how this works :sweat_smile:.

* How to link a device to a `Topology`? (I mean by passing `last_ip` as `gateway id` or anything similar so later on we can find topology related to a device).

> Two charts should be created: one which shows the amount of links (one line for the up state and one line for the down state) and another chart for the amount of nodes (one line for total nodes and one line for isolated nodes).

* I think the former chart should be similar to `uptime` chart while the latter might be similar to `rtt` chart (not sure). I don't know what is meant by `isolated nodes`, a brief description please.

I have these queries due to lack of deeper understanding of networks, I am looking for relevant resources to understand key concepts but am unable to classify what is useful and what is not. So, if you have any resources to suggest, please do it. Thanks! :smile:

nemesisdesign:
@nepython let's put this in standby. Do not work further on this please.

nemesisdesign:
I'm closing this. Not needed right now. We'll work on other things.

PR #152, opened 2020-07-02
[check] Refactor Check classes

nepython:
Related to #54, #55. Breaks #123 into smaller PRs as per @PabloCastellano's suggestion.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -224,6 +224,18 @@ has a value which is not in the expected range
 Example: ping is by default a critical metric which is expected to be always 1
 (reachable).
 
+``Available Checks``
+--------------------
+
+``Ping``
+~~~~~~~~
+
+This check returns information on device ``uptime`` and ``RTT (Round trip time)``.
+The Charts ``uptime``, ``packet loss`` and ``rtt`` are created. The ``fping``
+command is used to collect these metrics.
+You may choose to disable auto creation of this check by setting
+`OPENWISP_MONITORING_AUTO_PING <#OPENWISP_MONITORING_AUTO_PING>`_ to ``False``.
+
 Settings
 --------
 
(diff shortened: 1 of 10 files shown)


openwisp_monitoring/check/tests/test_models.py
```diff
@@ -26,6 +32,14 @@ def test_check_class(self):
         c = Check(name='Ping class check', check=self._PING)
         self.assertEqual(c.check_class, Ping)
 
+    def test_base_check_class(self):
```
PabloCastellano, line 35: I think this test does not provide much value. I would remove it

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -16,6 +17,11 @@ def test_check_str(self):
         c = Check(name='Test check')
         self.assertEqual(str(c), c.name)
 
+    def test_check_no_content_type(self):
```
PabloCastellano, line 20: What are you testing here?

nepython: This happens to be a rare case when you create a check without passing any content_object (currently no such check classes exist). The test is verifying that everything works fine in that case as a metric was created by passing `Check` as the **content_object** and yes it covers 2 uncovered lines :sweat_smile:

openwisp_monitoring/check/tests/test_models.py
nepython: I had to add these tests since there were 4 uncovered lines and coverage fell by 0.1% (currently it's increased by 0.2%). If coverage is not much important then I can remove the tests though this one is verifying that you cannot derive a `Check` class from `BaseCheck` without defining `check` method (it was suggested by @nemesisdesign to raise `NotImplementedError` in all such cases. )

README.rst
```diff
@@ -224,6 +224,18 @@ has a value which is not in the expected range
 Example: ping is by default a critical metric which is expected to be always 1
 (reachable).
 
+``Available Checks``
+--------------------
+
+``Ping``
```
nemesisdesign, line 230: Nice, thank you :+1:

PR #155, opened 2020-07-05
[models] Swapper for openwisp-controller & openwisp-users

atb00ker:


openwisp_monitoring/device/models.py
```diff
@@ -1,12 +1,16 @@
 from django.contrib.contenttypes.fields import GenericRelation
 from swapper import get_model_name, swappable_setting
 
-from openwisp_controller.config.models import Device
+# TODO:N Fix me
+from openwisp_controller.config.models import Device as BaseDevice
```
atb00ker, line 5: For some reason, this works but swappper.load_model is not working because "Model's are not loaded". I am looking into it if it's possible.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -364,19 +358,19 @@ def test_device_metrics_received_signal(self):
...
         with redirect_stderr(StringIO()) as stderr:
-            c.configuration = 'invalid'
-            c.save()
-            r = self.client.get(self._url(d.pk.hex, d.key))
-        self.assertIn('InvalidChartConfigException', stderr.getvalue())
-        self.assertEqual(r.status_code, 200)
+            conf.configuration = 'invalid'
+            conf.save()
+            # TODO:N not working testcase
+            self.assertIn('InvalidChartConfigException', stderr.getvalue())
```
atb00ker, line 379: @nepython this test is failing, I checked couple of things and I doesn't seem to failing because of openwisp-controller update.
Infact, when I run this manually, typing `conf.configuration` does throw the error, but `stderr.getvalue()` is still empty, is this failing because of some other work I should be aware of or should I proceed to modify the test to fix it?

nepython: You need to send the request prior to checking the exception, the reason being the exception is raised after `clean` method is called. So, you can simply do it as is being done currently.
```python
response = self.client.get(self._url(device.pk.hex, device.key))
self.assertIn('InvalidChartConfigException', stderr.getvalue())
```

Also, note :point_up: above you have used `device` but there is no such variable defined (it's still `d`, so change either as seems better :smiley:)

atb00ker: Changing the order fixed it. I may have mistakenly changed the order while fixing another testcase, thanks! :smile:

tests/openwisp2/sample_monitoring/migrations/0001_initial.py
```diff
@@ -131,7 +132,7 @@ class Migration(migrations.Migration):
                     'metric',
                     models.OneToOneField(
                         on_delete=django.db.models.deletion.CASCADE,
-                        to='sample_monitoring.Metric',
+                        to=swapper.get_model_name('monitoring', 'Metric'),
```
atb00ker, line 135: This change has no effect on `sample_monitoring`, I only did it because if user tries to copy-paste the migration file in their own sample app(because people may copy the entire folder and start working on it), it would still work.

openwisp_monitoring/check/base/models.py
```diff
@@ -77,10 +75,9 @@ def perform_check(self, store=True):
 
 if app_settings.AUTO_PING:
```
PabloCastellano, line 76: Since you are now connecting the post_save signal instead of using the 'receiver' decorator and you already check this condition in the _connect_signals method, I think it's safe to remove it now from here

openwisp_monitoring/check/tasks.py
```diff
@@ -37,7 +37,8 @@ def perform_check(uuid):
 @shared_task
 def auto_create_ping(model, app_label, object_id):
     """
-    Called by openwisp_monitoring.check.models.auto_ping_receiver
+    Called by django signal (dispatch_uid: auto_ping)
+    registered  in check app's apps.py
```
PabloCastellano, line 41: double space

openwisp_monitoring/device/admin.py
```diff
@@ -8,17 +8,19 @@
...
 from . import settings as app_settings
 
 DeviceData = load_model('device_monitoring', 'DeviceData')
 DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
 Chart = load_model('monitoring', 'Chart')
+Device = load_model('config', 'Device')
 
 
 class DeviceAdmin(BaseDeviceAdmin):
+    change_form_template = 'admin/config/device/change_form.html'
```
PabloCastellano, line 22: Why has it changed?

openwisp_monitoring/device/apps.py
```diff
@@ -25,7 +27,7 @@ def ready(self):
         self.device_recovery_detection()
 
     def connect_device_signals(self):
-        from openwisp_controller.config.models import Device
+        Device = load_model('config', 'Device')
```
PabloCastellano, line 30: I think we can move the following definitions to the top of this file:

```
        Device = load_model('config', 'Device')
        DeviceData = load_model('device_monitoring', 'DeviceData')
        DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
```

nemesisdesign: unfortunately it can't be done because the models are not loaded at that point yet, trying to load them will raise an exception

openwisp_monitoring/device/apps.py
```diff
@@ -96,13 +97,10 @@ def trigger_device_recovery_checks(cls, instance, **kwargs):
...
-    def connect_is_working_changed(self):
-        from openwisp_controller.connection.models import DeviceConnection
-        from openwisp_controller.connection.signals import is_working_changed
-
+    def connect_is_working_changed(cls):
         is_working_changed.connect(
-            self.is_working_changed_receiver,
-            sender=DeviceConnection,
+            cls.is_working_changed_receiver,
+            sender=load_model('connection', 'DeviceConnection'),
```
nemesisdesign, line 103: thanks!

tests/openwisp2/settings.py
```diff
@@ -86,6 +88,7 @@
 ]
 
 MIDDLEWARE = [
+    # 'debug_toolbar.middleware.DebugToolbarMiddleware',
```
nemesisdesign, line 91: seems you forgot some debugging lines in this file

openwisp_monitoring/device/admin.py
nemesisdesign: I also would like to know this

openwisp_monitoring/check/tasks.py
atb00ker: Fixed, thanks! :smile:

openwisp_monitoring/check/base/models.py
atb00ker: Fixed! :smile:

openwisp_monitoring/device/admin.py
atb00ker: Without this change, `change_form_template` is fetched from [admin/config/change_form.html](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/templates/admin/config/change_form.html) from openwisp-controller, Previously, this wasn't required because path to this `change_form` was something like `admin/django_netjsonconfig/.../...`.

tests/openwisp2/settings.py
atb00ker: Yeah, sorry for causing the confusion again, when I mentioned on gitter, it wasn't ready for review! :smile:

openwisp_monitoring/device/admin.py
nemesisdesign: I think I understand

PR #158, opened 2020-07-06
[feature] Added notification widget from openwisp-notifications

pandafy:


README.rst
```diff
@@ -69,24 +69,26 @@ Follow the setup instructions of `openwisp-controller
...
         'openwisp_controller.connection',
         # monitoring
-        'notifications',
         'openwisp_monitoring.monitoring',
         'openwisp_monitoring.device',
         'openwisp_monitoring.check',
+        # notifications
+        'openwisp_notifications',
+        # openwisp2 admin theme (must be loaded here)
+        'openwisp_utils.admin_theme',
```
nemesisdesign, line 91: can you quickly explain why this change is needed? I guess is for the template but just double checking.

pandafy: Yes, it is for template. We are now using `base_site.html` instead of `base.html` in `openwisp-notifications`, hence reordering of installed apps was required. :smile:

nemesisdesign: Ok, please remove `apptemplates.Loader` from `settings.py` (it should fail now that django-apptemplates is removed from the dependencies)

install-dev.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# TODO: removed when openwisp-controller 0.8.0 is released
+pip install -U https://github.com/openwisp/openwisp-controller/tarball/master
+# TODO: removed when openwisp-users 0.3.0 is released
+pip install -U https://github.com/openwisp/openwisp-users/tarball/master
+# TODO: remove when openwisp-notifications 0.1 is released
+pip install -U https://github.com/openwisp/openwisp-notifications/tarball/master
```
nemesisdesign, line 8: here we'll need to point to dev

PR #162, opened 2020-07-10
[fix] Create new models Chart, AlertSettings, fixes migrations #161

### LINKED ISSUE #161: [bug] Migrations broken for AlertSettings and Chart models
I tried this on `master`.

Tried adding any field on `AlertSettings` or `Chart`, the migration won't be executed, as found in https://github.com/openwisp/openwisp-monitoring/pull/150#issuecomment-654786288.

This has started happening after the renaming of the models.

I went back to commit 8035f76 and the problem disappeared.

@nepython this is a release blocker.

nepython:
Checks:

Closes #161

I am not sure why are there issues with renaming. These are the ones we have observed till now and this might be tip of the ice berg,
1. `table_name` was not getting renamed (so we took care of it in the migrations)
2. `ContentType` was not getting registered (once again we took care of it in the migrations)
3. Unable to add field (we discovered this now)

Normally `django` was supposed to take care of it, yet we had these issues. On `stackoverflow` and elsewhere I did find that others faced issues (some similar to ours) while using `migrations.RenameModel` and they ended up using different strategies (prominently **South**) but I couldn't find a explanation of why this behavior was observed in the first place.

I will try to replicate this on some dummy django apps (to see if it can be reproduced easily) and search if there are any existing tickets for the same (if not then I will go ahead and open a ticket).

Right now, the best approach I think might be that we should create new models `Chart` and `AlertSettings`, perform data migration and then delete `Graph`, `Threshold` models. The above three issues don't appear with this appraoch (tested it).
Since, we have already renamed `Graph` -> `Chart` and `Threshold` -> `AlertSettings`, I had two add two lines initially to revert these changes.

**Note: After these changes, previous ones (renaming migrations in monitoring app) become obsolete but I didn't modify them for making it backward compatible in already deployed systems.**


tests/openwisp2/sample_monitoring/migrations/0006_create_chart_alertsettings.py
```diff
@@ -0,0 +1,130 @@
+# Generated by Django 3.0.3 on 2020-07-10 08:40
+
```
nemesisdesign, line 2: we do not need to add migrations to the sample app, we can just destroy everything and recreate them.
@nepython @atb00ker @PabloCastellano @NoumbissiValere @pandafy :point_up_2: :warning: :warning: 
Please read this. I have seen you all doing it. Focus on the main migrations, sample app is not supposed to be maintained, it's only an example and used mainly for testing.

nepython:
> This should be merge into dev, right?

I am unsure, the bug is valid on both branches. If it's merged in master then I will need to make small changes on dev while rebasing to resolve conflicts. If the target branch needs to be changed to `dev` then let me know.

PabloCastellano:
@nepython Ok that's right. Better fix it in master.

PR #165, opened 2020-07-16
[fix] Do not call update_status directly if any check is present

nemesisdesign:
Added recovery argument to allow to be used also as a way to signal device issues.
Needed for https://github.com/openwisp/openwisp-network-topology/issues/75

Added mention of this integration in the README.
Will test this manually soon and then merge.

Checks:

### DIFF
--- a/openwisp_monitoring/device/tasks.py
+++ b/openwisp_monitoring/device/tasks.py
@@ -5,17 +5,20 @@ from ..check.tasks import perform_check
 
 
 @shared_task
-def trigger_device_checks(pk):
+def trigger_device_checks(pk, recovery=True):
     """
     Retrieves all related checks to the passed ``device``
     and calls the ``perform_check`` task from each of them.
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/device/tasks.py
```diff
@@ -5,17 +5,20 @@
 
 
 @shared_task
-def trigger_device_checks(pk):
+def trigger_device_checks(pk, recovery=True):
```
nepython, line 8: This might be better as we can choose as to what should happen in case checks don't exist. Maybe the parameter `recovery` could be renamed as it seems slightly misleading to me but I can't think of a better name right now.

nemesisdesign: if another part of openwisp detects a recovery (eg: management VPN status is up again) it will pass True, otherwise it will pass False, it's also consistent with "device recovery detection".

PR #170, opened 2020-07-25
Chores: End of July 2020 cleanup

nemesisdesign:
Follow up to #150 and #148.


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -460,18 +460,15 @@ def _round(value, decimal_places):
 class AbstractAlertSettings(TimeStampedEditableModel):
     _SECONDS_MAX = 60 * 60 * 24 * 7  # 7 days
     _SECONDS_HELP = (
-        'for how long should the alert_settings value be crossed before '
-        'triggering an alert? The maximum allowed is {0} seconds '
+        'for how long should the value be crossed before '
+        'an alert is sent? The maximum allowed is {0} seconds '
```
nepython, line 464: A corresponding change in `sample_check` migrations is missing

openwisp_monitoring/device/static/monitoring/js/alert-settings.js
```diff
@@ -0,0 +1,11 @@
+(function ($) {
+    'use strict';
+    $(function () {
+        var checks = $('[id^="id_monitoring-metric-content_type-object_id"][id $="alertsettings-0-is_active"]');
```
nepython, line 4: I don't recall why I named it `checks`. Maybe we should rename it (something like `is_active` simply), since `checks` can create confusion that it's related to the `Check` model. Please feel free to rename it, if you think it's a better name.

openwisp_monitoring/monitoring/migrations/0017_merge_traffic_metrics_and_fill_metric_config.py
```diff
@@ -13,8 +13,8 @@
 
 
 def merge_traffic_metrics(apps, schema_editor):
-    Metric = load_model('monitoring', 'Metric')
-    Chart = load_model('monitoring', 'Chart')
+    Metric = apps.get_model('monitoring', 'Metric')
+    Chart = apps.get_model('monitoring', 'Chart')
```
nepython, line 17: Using `apps.get_model` is always preferred in Migrations, don't know how I missed it :sweat_smile: (maybe due to #161, but it's resolved now so no worries). Thanks!

PR #180, opened 2020-07-29
[admin] Improve device checks tab

nepython:
Checks:

To fix:
* Check is showing `help_text` for `read_only field`, need to remove this.


nepython:
This is how it's looking currently, I have renamed `active` to `is_active` but retained admin display name as `Active` since it seemed better than `Is active`. The help_text below `Check type` looks pretty bad. Tried to remove it but faced `generic_inline` limitations. I wanted to suggest shall we remove the `help_text` from the field since it's not adding much value.
Screenshot from 2020-07-30 21-36-30

openwisp_monitoring/check/base/models.py
```diff
@@ -15,7 +15,14 @@
 
 class AbstractCheck(TimeStampedEditableModel):
     name = models.CharField(max_length=64, db_index=True)
-    active = models.BooleanField(default=True, db_index=True)
+    is_active = models.BooleanField(
+        _('active'),
+        default=True,
+        db_index=True,
+        help_text=_(
+            'whether the check should be run, related metrics be collected and alerts sent'
```
nemesisdesign, line 23: `related metrics be collected` doesn't sound correct to me, I think it should be `related metrics collected`

Issue #70, opened 2020-05-05
[docs] Update README


nepython:
Ensure the `README` is up to date, lists all the features available, provides some `screenshots` and a detailed explaination of the usage of the main features added in this project.

### Things to be documented:
* [ ] Badges as other OpenWISP modules
* [ ] `run_checks` command (maybe something similar to https://github.com/openwisp/django-freeradius/blob/master/docs/source/general/management_commands.rst)
* [ ] Signals Missing:
 * pre_metric_write
 * post_metric_write
 * health_status_changed
* [ ] Images (`AlertSettings Inline`, `Checks Inline`, `Metric Health`) and GIF need to be updated as per latest changes.
* [ ] Reference to `OPENWISP_MONITORING_DEVICE_CONFIG_CHECK_MAX_TIME` setting needs to be removed from https://github.com/openwisp/openwisp-monitoring/tree/dev#configuration-applied (somehow missed to notice it until now :/).

nepython:
I will surely document as I introduce the changes but this issue is for updating the **README** to make it more effective for the end-user to understand the module :smile:

nemesisdesign:
Let's coordinate with @atb00ker since he's been doing this for different modules and we've been experimenting with animated gifs which seem a great way to show what the module can do at glance.

@pandafy related to you as well :point_up:

atb00ker:
My workflow for creating the GIFs, 
1. List all the features in order that I want to show them in the video.
2. Do some demo runs to ensure that GIF is as short as possible in the final shot.
3. Use Firefox, because F11 takes you to full screenwithout any borders.
4. Use `simplescreenrecorder` recording `mkv` in full screen at 7 FPS
5. Convert mkv to an efficient GIF: 
```
ffmpeg -i  vid.mkv -vf fps=7,scale=1366:-1:flags=lanczos,palettegen palette.png
ffmpeg -i vid.mkv -i palette.png -filter_complex "fps=7,scale=1366:-1:flags=lanczos[x];[x][1:v]paletteuse" demo.gif
```

6. Use https://ezgif.com/optimize to optimize further according to requirement.
7. If the size is still big, some steps specific to the GIF may be required.

atb00ker:
@nepython @pandafy , let me know if you need any further assistance with creating GIFs! :smile:

nepython:
Thanks, @atb00ker :smiley:

nemesisdesign:
[...]

PR #185, opened 2020-08-06
[enhancement] Add way to specify global defaults

### LINKED ISSUE #106: [alert settings] Add way to specify global defaults

nepython:
Checks:

Closes #106


nepython:
The build is failing after I rebased it to the latest changes on `dev` branch. The last commit on `dev` https://github.com/openwisp/openwisp-monitoring/commits/dev too was [failing with the same assertion error](https://travis-ci.org/github/openwisp/openwisp-monitoring/jobs/715362589) (related to changing `http` --> `https`). I am checking what's wrong.

README.rst
```diff
@@ -1083,7 +1179,7 @@ If you want to extend this view, you will have to perform the additional steps b
         # add your customizations here ...
         pass
 
-**Step 2: remove the following line from your root ``urls.py`` file:**
+Step 2: remove the following line from your root ``urls.py`` file:
```
nepython, line 1182: **Note**: This is a very small refactoring in `Readme`, not related to this PR. Same for below change.

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,512 @@
...
+        'notification': {
+            'problem': {
+                'name': 'Ping PROBLEM',
+                'verb': 'cannot be reached anymore',
+                'level': 'warning',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} is not reachable'
+                ),
+                'message': _(
+                    'The device [{notification.target}] {notification.verb} anymore by our ping '
```
nemesisdesign, line 104: Watch out, the markdown link has been removed and only the square brackets have remained:

Screenshot from 2020-08-07 21-11-40

The device should be linked as the word "default" in the other sample notifications.
Applies to the other alerts as well.

If in doubt, ask @pandafy.

nepython: This has been fixed now.

openwisp_monitoring/monitoring/migrations/0020_make_alertsettings_fields_null.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db import migrations
+
+
+def make_fields_null(apps, schema_editor):
+    """
+    This function checks the values of alertsettings fields,
+    if the values match default values provided in the metric configuration
+    then those values will be set as None
+    """
+    AlertSettings = apps.get_model('monitoring', 'AlertSettings')
```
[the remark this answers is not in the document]
nepython: Thanks, the migration functions now won't make use of `AlertSettings`'s methods.

openwisp_monitoring/monitoring/admin.py
```diff
@@ -9,8 +10,21 @@
...
 
 
+class AlertSettingsForm(ModelForm):
+    def __init__(self, *args, **kwargs):
+        instance = kwargs.get('instance')
+        if instance:
+            kwargs['initial'] = {
+                'tolerance': instance.get_tolerance,
+                'operator': instance.get_operator,
+                'threshold': instance.get_threshold,
```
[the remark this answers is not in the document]
nepython: Yep, I too noticed this yesterday night and was working on `full_clean`. Added a regression test for the same,
https://github.com/openwisp/openwisp-monitoring/blob/3f8215fc982161344eea40b390bb17934188595d/openwisp_monitoring/monitoring/tests/test_models.py#L254-L272

nepython:
So currently 3 things need to be done:
* [ ] Making the **docs** more friendly by giving more details and reordering few alike topics after some discussion.
* [ ] Changing **tolerance field** input from `seconds` to `minutes`.
* [ ] Fixing the percentage display in **notifications** (it should display `70%` and not `0.7`)

openwisp_monitoring/monitoring/migrations/0020_make_alertsettings_fields_null.py
```diff
@@ -0,0 +1,32 @@
...
+    alert_settings_queryset = AlertSettings.objects.all()
+    for alert_settings in alert_settings_queryset:
+        if alert_settings.threshold == alert_setting.config_dict['threshold']:
+            alert_settings.threshold = None
+        if alert_settings.tolerance == alert_setting.config_dict['tolerance']:
+            alert_settings.tolerance = None
+        if alert_settings.operator == alert_setting.config_dict['operator']:
+            alert_settings.operator = None
+        alert_settings.save()
```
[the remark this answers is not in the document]
nepython: I have added backward migrations but please note that there is a conversion of `seconds` --> `minutes` (for **tolerance** field) which will show a minor change. So, say if you had `seconds=61` for one `AlertSettings`  before migrating. After migrating and then running reverse migrations you will have `seconds=60`.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -482,26 +483,68 @@ class AbstractAlertSettings(TimeStampedEditableModel):
...
+    @property
+    def get_threshold(self):
+        return self.threshold or self.config_dict['threshold']
+
+    @property
+    def get_tolerance(self):
+        return self.tolerance or self.config_dict['tolerance']
+
+    @property
+    def get_operator(self):
```
nemesisdesign, line 536: methods named `get_` should not be decorated with `@property`.

I see mainly two possibilities:

- either leave these methods undecorated and call them like `instance.get_threshold()`
- rename the model fields as `custom_<field_name>` and rename these decorated methods to remove `get_` 

eg:

```python
@property
def tolerance(self):
    return self.custom_tolerance or self.config_dict['tolerance']
```

nepython:
> **Question**: what if a user wants to change the default tolerance of the ping metric?
Do they have to redefine everything or can we find a way to allow them to only specify the differences?

Well for `AlertSettings` specifically we have `custom_threshold`, `custom_operator`, `custom_tolerance` but I get your query, what if the user wants to change say `notification`, right? I got you covered, see below. :smiley:
> Maybe we could add a new function specifically for this, eg:
>```python
># sloppy pseudocode alert!!
>
>modify_thing('ping', {'alerts': {'tolerance': 3}}
>```
>It's just an idea, if you have a better idea let me know.

@nemesisdesign, I took inspiration from your work with `charts` and adapted the setting `OPENWISP_MONITORING_CHARTS` for metrics. This should cover your use case and also maintains consistency, I think https://github.com/openwisp/openwisp-monitoring/tree/issues/106-global-defaults#openwisp_monitoring_metrics. I know `README` needs improvement :sweat_smile:

tests/openwisp2/sample_monitoring/migrations/0001_initial.py
```diff
@@ -167,19 +171,33 @@ class Migration(migrations.Migration):
...
+                        verbose_name='operator',
                     ),
                 ),
-                ('value', models.FloatField(help_text='threshold value')),
                 (
-                    'seconds',
+                    'custom_threshold',
+                    models.FloatField(
+                        blank=True,
+                        help_text='sets the threshold for collected metric data to send alerts, you can set threshold value as a float',
```
nemesisdesign, line 187: I would just keep `threshold value`, which will be clear to most users

tests/openwisp2/sample_monitoring/migrations/0001_initial.py
```diff
@@ -167,19 +171,33 @@ class Migration(migrations.Migration):
...
+                    ),
+                ),
+                (
+                    'custom_tolerance',
                     models.PositiveIntegerField(
+                        blank=True,
                         default=0,
-                        help_text='for how long should the value be crossed before an alert is sent? The maximum allowed is 604800 seconds (7 days)',
-                        validators=[django.core.validators.MaxValueValidator(604800)],
+                        help_text='for how long should the value be crossed before an alert is sent? The maximum allowed is 10080 minutes (7 days)',
```
nemesisdesign, line 197: let's change this to something more explicit and useful:

```
for how many minutes should the threshold value be crossed before an alert is sent? A value of zero means the alert is sent immediately
```

README.rst
```diff
@@ -741,14 +726,165 @@ An example usage is shown below.
...
+
++--------------------------+------------------------------------------------------+
+|      **Parameter**       |                   **Description**                    |
++--------------------------+------------------------------------------------------+
+|     **metric_name**:     | A ``str`` defining name of the metric configuration. |
++--------------------------+------------------------------------------------------+
+|**metric_configuration**: | A ``dict`` defining configuration of the metric.     |
++--------------------------+------------------------------------------------------+
+
+An example usage has been shown below.
```
nemesisdesign, line 757: Please summarize what the following example does.

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+            'problem': {
+                'name': 'Ping PROBLEM',
+                'verb': 'cannot be reached anymore',
+                'level': 'warning',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} is not reachable'
+                ),
+                'message': _(
+                    'The device [{notification.target}]({notification.target_link}) '
+                    '{notification.verb} anymore by our ping messages.'
```
nemesisdesign, line 105: please remove ` anymore by our ping messages`

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                ),
+            },
+            'recovery': {
+                'name': 'Ping RECOVERY',
+                'verb': 'has become reachable',
```
nemesisdesign, line 110: let's change this to `is reachable again`

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                'level': 'info',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} is reachable again'
```
nemesisdesign, line 113: here you can use the verb instead of duplicating "is reachable again"

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                ),
+                'message': _(
+                    'The device [{notification.target}]({notification.target_link}) '
+                    '{notification.verb} again by our ping messages.'
```
nemesisdesign, line 117: let's remove ` again by our ping messages`

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                'unit': f' {_("ms")}',
+                'order': 220,
+                'query': chart_query['rtt'],
+            },
+        },
+        'alert_settings': {'operator': '<', 'threshold': 1, 'tolerance': 0},
+        'notification': {
+            'problem': {
+                'name': 'Ping PROBLEM',
+                'verb': 'cannot be reached anymore',
```
nemesisdesign, line 98: `is not reachable`

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                'level': 'warning',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} is not reachable'
```
nemesisdesign, line 101: `[{site.name}] PROBLEM: {notification.target} {notification.verb}`

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                ],
+                'unit': f' {_("ms")}',
+                'order': 220,
+                'query': chart_query['rtt'],
+            },
+        },
+        'alert_settings': {'operator': '<', 'threshold': 1, 'tolerance': 0},
+        'notification': {
+            'problem': {
+                'name': 'Ping PROBLEM',
```
nemesisdesign, line 97: what's this key used for? Is this shown to the user?

openwisp_monitoring/monitoring/configuration.py
nemesisdesign: make string translatable

openwisp_monitoring/monitoring/configuration.py
nemesisdesign: make string translatable

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+        'key': 'config_applied',
+        'field_name': 'config_applied',
+        'alert_settings': {'operator': '<', 'threshold': 1, 'tolerance': 5},
+        'notification': {
+            'problem': {
+                'name': 'Configuration Applied PROBLEM',
+                'verb': 'is not applied',
+                'level': 'warning',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} has problem in configuration'
```
nemesisdesign, line 134: missing PROBLEM

it should say something like: 

PROBLEM device X configuration status issue

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+                ),
+                'message': _(
+                    'The configuration for device [{notification.target}]'
+                    '({notification.target_link}) {notification.verb} properly.'
```
nemesisdesign, line 138: The configuration **of** device X ......... has not been applied in a timely manner.

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -0,0 +1,519 @@
...
+            'recovery': {
+                'name': 'Configuration Applied RECOVERY',
+                'verb': 'configuration has been applied',
+                'level': 'info',
+                'email_subject': _(
+                    '[{site.name}] {notification.target} configuration applied successfully'
+                ),
+                'message': _(
+                    'The device [{notification.target}]({notification.target_link}) '
+                    '{notification.verb} successfully.'
```
nemesisdesign, line 150: has been applied again successfully

nepython:
Addressed all reviews /  issues besides docs (working on it).

> Why did you remove RECOVERY: and PROBLEM: from the email subjects?
Please add these back.
Every email alert must have PROBLEM or RECOVERY stated clearly in its subject

The issue had 3 notification examples. None had `RECOVERY` or `PROBLEM` in email subjects, hence I removed them. :sweat_smile:. Reverting.

openwisp_monitoring/monitoring/configuration.py
nepython: I am not sure how it is used internally, I didn't observe it via the admin. @pandafy might be able to confirm / correct me.

nemesisdesign:
> **PS**: Docs updated as per the review but topics can be made more ordered to improve a new user's experience. Maybe we can do this in #70 since it's a dedicated issue for improving `README`.

As a general rule, it's better to work on documentation while working on a specific feature, because our minds are already focused on it and therefore it requires less effort compared to postponing it.

openwisp_monitoring/monitoring/configuration.py
nemesisdesign: @nepython according to the [openwisp-notifications README](https://github.com/openwisp/openwisp-notifications/tree/dev#register-notification-type), this key is not required.
Please double check and remove it if it's not needed.
@pandafy anything to add?

pandafy: @nepython this is what is shown to the user. Please rename it to `verbose_name`. It was set incorrectly earlier. :sweat_smile: 
I fixed it in [#189](https://github.com/openwisp/openwisp-monitoring/pull/189/files#diff-583ce6230e4e945550ab510f1b8be8d8R154), but please make such changes here also. It will be used while [displaying notification settings](https://user-images.githubusercontent.com/32094433/89739293-2b560600-da9d-11ea-9bdb-ce92c89dc3be.gif).

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -7,14 +7,66 @@
...
 AlertSettings = load_model('monitoring', 'AlertSettings')
 
+default_message = (
+    '{notification.actor.name} for device [{notification.target}]'
+    '({notification.target_link}) {notification.verb}.'
+)
+
+test_notification = {
+    'problem': {
+        'name': 'Monitoring Alert',
```
nemesisdesign, line 30: should be `verbose_name`, right?

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -7,14 +7,66 @@
...
+        'verb': 'crossed the threshold',
+        'level': 'warning',
+        'email_subject': '[{site.name}] PROBLEM: {notification.actor.name} {notification.target}',
+        'message': default_message,
+    },
+    'recovery': {
+        'name': 'Monitoring Alert',
```
nemesisdesign, line 37: should be `verbose_name`, right?

README.rst
```diff
@@ -618,71 +709,173 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+                ],
+                'unit': f' {_("ms")}',
+                'order': 220,
+                'query': chart_query['rtt'],
+            },
+        },
+        'alert_settings': {'operator': '<', 'threshold': 1, 'tolerance': 0},
+        'notification': {
+            'problem': {
+                'name': 'Ping PROBLEM',
```
nemesisdesign, line 808: should be `verbose_name`, right?

README.rst
```diff
@@ -618,71 +709,173 @@ In case you just want to change the colors used in a chart here's how to do it:
...
+                'email_subject': _(
+                    '[{site.name}] {notification.target} is not reachable'
+                ),
+                'message': _(
+                    'The device [{notification.target}] {notification.verb} anymore by our ping '
+                    'messages.'
+                ),
+            },
+            'recovery': {
+                'name': 'Ping RECOVERY',
```
nemesisdesign, line 820: should be `verbose_name`, right?

README.rst
```diff
@@ -547,6 +532,112 @@ network topology module will trigger the monitoring checks.
...
+
+.. code-block:: python
+
+    from django.utils.translation import gettext_lazy as _
+
+    OPENWISP_MONITORING_METRICS = {
+        'config_applied': {
+            'notification': {
+                'problem': {
+                    'verb': _('has not been applied'),
```
nemesisdesign, line 599: does it have any effect changing `verbose_name` here? If it does, please include it in the example, eg: "PING problem changed name" or something like that.

README.rst
```diff
@@ -547,6 +532,112 @@ network topology module will trigger the monitoring checks.
...
+                        '[{site.name}] PROBLEM: {notification.target} configuration '
+                        'status issue'
+                    ),
+                    'message': _(
+                        'The configuration for device [{notification.target}]'
+                        '({notification.target_link}) {notification.verb} in a timely manner.'
+                    ),
+                },
+                'recovery': {
+                    'verb': _('configuration has been applied again'),
```
nemesisdesign, line 610: same here

README.rst
nepython: Yes, it will have. Earlier I wasn't very sure how helpful it might be earlier. It should be after `Notification Settings`, I think. Added the `verbose_name` as `Configuration PROBLEM` (default is `Configuration Applied PROBLEM` so just shortened it).

PR #190, opened 2020-08-10
[timeseries] Make metric collection resilient to failures #166

### LINKED ISSUE #166: [feature] Metric collection resilient to failures
If for some reason the timeseries DB is overloaded, the write fails and the data is lost.

I think this is avoidable and the system can become resilient to these kind of failures, here's how I think it has to be done (it's not a literal plan but a guideline, things can be moved around if needed):

- the write to the timeseries DB should be passed to a background worker (celery task), we should pass the data to write as a dictionary (containing only integers, strings, floats, things that can be serialized and passed from one process to the other), the data should contain the information about the datetime which must persist in case of failure
- now that the write happens in a celery worker, in case of failures we have to designate an exception which will be retried automatically (with exponential backoff, like we have in [openwisp-firmware-upgrade for recoverable failures during firmware upgrade operations](https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/tasks.py#L15-L33)); we shall create a new exception dedicated for this case, we could name it `TimeseriesWriteException` or anything that communicates the concept well
- the method of the timeseries abstraction which ultimately performs the write operation should catch any exception got from the underlying level library (eg: influxdb or elasticsearch) and re-raise `TimeseriesWriteException` wrapping the lower level exception

This way, whenever `TimeseriesWriteException` is raised in the background worker, the operation will be retried indefinitely until the problem is solved and no monitoring data will be lost.
As I already mentioned above, it's important that the information about the datetime persists so that it will be written explicitly, because if the write succeeds with a delay, the datetime of the points could be stored with a wrong datetime, which would have negative consequences on the reliability of the data.

The [retry options of celery should be configurable as in openwisp-firmware-upgrader](https://github.com/openwisp/openwisp-firmware-upgrader/#openwisp_firmware_upgrader_retry_options).

nepython:
Checks:

Closes #166

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -678,6 +678,37 @@ information by performing lookups on the OUI
 
 This feature is enabled by default.
 
+``OPENWISP_MONITORING_WRITE_RETRY_OPTIONS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    | ``dict``  |
++--------------+-----------+
+| **default**: | see below |
++--------------+-----------+
+
+.. code-block:: python
+
+    # default value of OPENWISP_MONITORING_RETRY_OPTIONS:
+
+    dict(
+        max_retries=None,
+        retry_backoff=True,
+        retry_backoff_max=600,
+        retry_jitter=True,
+    )
+
+Retry settings for recoverable failures during metric writes.
+
+By default if a metric write fails (eg: due to excessive load on timeseries database at that moment)
+then the operation will be retried indefinitely with an exponential random backoff and a maximum delay of 10 minutes.
+
+This feature makes the monitoring system resilient to temporary outages and helps to prevent data loss.
+
+For more information regarding these settings, consult the `celery documentation
+regarding automatic retries for known errors
+<https://docs.celeryproject.org/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions>`_.
+
 Registering / Unregistering Chart Configuration
 -----------------------------------------------
 
--- a/openwisp_monitoring/db/backends/influxdb/client.py
+++ b/openwisp_monitoring/db/backends/influxdb/client.py
@@ -10,9 +10,10 @@ from django.utils.functional import cached_property
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from influxdb import InfluxDBClient
+from influxdb.exceptions import InfluxDBClientError
 
+from ...exceptions import TimeseriesWriteException
 from .. import TIMESERIES_DB
-from .exception import DatabaseException
 
 logger = logging.getLogger(__name__)
 
@@ -53,7 +54,7 @@ class DatabaseClient(object):
     def __init__(self, db_name=None):
         self._db = None
         self.db_name = db_name or TIMESERIES_DB['NAME']
-        self.client_error = DatabaseException.client_error
+        self.client_error = InfluxDBClientError
 
     def create_database(self):
         """ creates database if necessary """
@@ -117,13 +118,17 @@ class DatabaseClient(object):
         if isinstance(timestamp, datetime):
             timestamp = timestamp.isoformat(sep='T', timespec='microseconds')
         point['time'] = timestamp
-        self.get_db.write(
-            {'points': [point]},
-            {
-                'db': kwargs.get('database') or self.db_name,
-                'rp': kwargs.get('retention_policy'),
-            },
-        )
+        try:
+            self.get_db.write(
+                {'points': [point]},
+                {
+                    'db': kwargs.get('database') or self.db_name,
+                    'rp': kwargs.get('retention_policy'),
+                },
+            )
+        except Exception as exception:
+            logger.warning(f'got exception while writing to tsdb: {exception}')
+            raise TimeseriesWriteException
 
     def read(self, key, fields, tags, **kwargs):
         extra_fields = kwargs.get('extra_fields')
--- a/openwisp_monitoring/db/backends/influxdb/tests.py
+++ b/openwisp_monitoring/db/backends/influxdb/tests.py
@@ -1,13 +1,20 @@
 from datetime import datetime, timedelta
+from unittest.mock import patch
 
+from celery.exceptions import Retry
 from django.core.exceptions import ValidationError
 from django.test import TestCase
 from django.utils.timezone import now
+from freezegun import freeze_time
+from influxdb import InfluxDBClient
+from influxdb.exceptions import InfluxDBServerError
 from openwisp_monitoring.device.settings import SHORT_RETENTION_POLICY
 from openwisp_monitoring.device.utils import SHORT_RP, manage_short_retention_policy
 from openwisp_monitoring.monitoring.tests import TestMonitoringMixin
+from pytz import timezone as tz
 from swapper import load_model
 
+from ...exceptions import TimeseriesWriteException
 from .. import timeseries_db
 
 Chart = load_model('monitoring', 'Chart')
@@ -240,3 +247,42 @@ class TestDatabaseClient(TestMonitoringMixin, TestCase):
         m.write('00:14:5c:00:00:00', time=datetime(2020, 7, 31, 22, 5, 47, 235142))
         m.write('00:23:4a:00:00:00', time=datetime(2020, 7, 31, 22, 5, 47, 235152))
         self.assertEqual(len(m.read()), 2)
+
+    @patch.object(
+        InfluxDBClient, 'write', side_effect=InfluxDBServerError('Server error')
+    )
+    def test_write_retry(self, mock_write):
+        with self.assertRaises(TimeseriesWriteException):
+            timeseries_db.write('test_write', {'value': 1})
+        m = self._create_general_metric(name='Test metric')
+        with self.assertRaises(Retry):
+            m.write(1)
+
+    @patch.object(
+        InfluxDBClient, 'write', side_effect=InfluxDBServerError('Server error')
+    )
+    def test_timeseries_write_params(self, mock_write):
+        with freeze_time('Jan 14th, 2020') as frozen_datetime:
+            m = self._create_general_metric(name='Test metric')
+            with self.assertRaises(Retry) as e:
+                m.write(1)
+            frozen_datetime.tick(delta=timedelta(minutes=10))
+            self.assertEqual(
+                now(), datetime(2020, 1, 14, tzinfo=tz('UTC')) + timedelta(minutes=10)
+            )
+            task_signature = e.exception.sig
+            with patch.object(timeseries_db, 'write') as mock_write:
+                self._retry_task(task_signature)
+            mock_write.assert_called_with(
+                'test_metric',
+                {'value': 1},
+                database=None,
+                retention_policy=None,
+                tags={},
+                # this should be the original time at the moment of first failure
+                timestamp='2020-01-14T00:00:00Z',
+            )
+
+    def _retry_task(self, task_signature):
+        task_kwargs = task_signature.kwargs
+        task_signature.type.run(**task_kwargs)
--- a/openwisp_monitoring/db/exceptions.py
+++ b/openwisp_monitoring/db/exceptions.py
@@ -1,0 +1,2 @@
+class TimeseriesWriteException(Exception):
+    pass
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -6,9 +6,11 @@ from datetime import datetime
 import swapper
 from cache_memoize import cache_memoize
 from dateutil.relativedelta import relativedelta
+from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.dispatch import receiver
+from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from jsonschema import draft7_format_checker, validate
 from jsonschema.exceptions import ValidationError as SchemaError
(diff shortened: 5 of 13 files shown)


openwisp_monitoring/db/exceptions.py
```diff
@@ -0,0 +1,7 @@
+class TimeseriesWriteException(Exception):
+    """
+    Caused by excessive load on timeseries server
+    leading to failure in storing metrics
```
nemesisdesign, line 4: this is misleading, we can't know the reason the tsdb is failing in advance, it may fail for any reason. The point is that when the system finds this exception will retry.

openwisp_monitoring/monitoring/tasks.py
```diff
@@ -0,0 +1,12 @@
+from celery import shared_task
+
+from ..db import timeseries_db
+from ..db.exceptions import TimeseriesWriteException
+
+
+@shared_task(bind=True, autoretry_for=(TimeseriesWriteException,), retry_backoff=True)
```
nemesisdesign, line 7: looks good here, we need to be able to specify more options here and be able to configure them by settings:
https://github.com/openwisp/openwisp-firmware-upgrader/#openwisp_firmware_upgrader_retry_options

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -204,14 +205,13 @@ def write(
...
-            name=self.key,
-            values=values,
+        options = dict(
             tags=self.tags,
-            timestamp=time,
+            timestamp=time or now(),
             database=database,
             retention_policy=retention_policy,
         )
+        timeseries_write(name=self.key, values=values, **options)
```
nemesisdesign, line 214: you do need to call `delay` here

openwisp_monitoring/db/exceptions.py
nepython: I will remove it but I thought the reason for failure was known to us, as it was stated in the issue:~
> If for some reason the timeseries DB is overloaded, the write fails and the data is lost.

nemesisdesign: you can never know the reason of all the possible failures that will happen when the service is used in the wild
the important bit is that these exceptions will be handled

nepython: Agreed :smiley:

nepython:
The retry mechanism seems to work perfectly now :)
Screenshot from 2020-08-10 23-54-12

README.rst
```diff
@@ -678,6 +678,25 @@ information by performing lookups on the OUI
 
 This feature is enabled by default.
 
+``OPENWISP_MONITORING_RETRY_OPTIONS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------------------------------------------------------------------------------------------+
+| **type**:    |                                            ``dict``                                              |
```
nemesisdesign, line 685: please do not do this thing of aligning the text to the center because it will be hard to maintain (plus not useful)

README.rst
```diff
@@ -678,6 +678,25 @@ information by performing lookups on the OUI
...
++--------------+--------------------------------------------------------------------------------------------------+
+| **default**: | ``{'max_retries': None, 'retry_backoff': True, 'retry_backoff_max': 600, 'retry_jitter': True}`` |
++--------------+--------------------------------------------------------------------------------------------------+
```
nemesisdesign, line 688: Please show the default value as it's done here for consistency and also for better readability:
https://github.com/openwisp/openwisp-firmware-upgrader/#openwisp_firmware_upgrader_retry_options

README.rst
```diff
@@ -678,6 +678,25 @@ information by performing lookups on the OUI
...
+
+Retry settings for recoverable failures during metric writes.
+
+By default if a metric write fails (eg. due to excessive load on timeseries server at that moment)
+then the operation will be retried indefinitely until it succeeds so that no metric data is lost
+with an exponential random backoff and a maximum delay of 10 minutes.
```
nemesisdesign, line 694: By default if a metric write fails (eg: due to excessive load on timeseries database at that moment)
then the operation will be retried indefinitely with an exponential random backoff and a maximum delay of 10 minutes.

This feature makes the monitoring system resilient to temporary outages and helps to prevent data loss.

openwisp_monitoring/device/base/models.py
```diff
@@ -167,15 +169,10 @@ def save_data(self, time=None):
...
-        timeseries_db.write(
-            name=self.__key,
-            values={'data': self.json()},
-            tags={'pk': self.pk},
-            timestamp=time,
-            retention_policy=SHORT_RP,
+        options = dict(
+            tags={'pk': self.pk}, timestamp=time or now(), retention_policy=SHORT_RP
         )
+        timeseries_write(name=self.__key, values={'data': self.json()}, **options)
```
nemesisdesign, line 175: why is this not asynchronous?

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -117,13 +118,16 @@ def write(self, name, values, **kwargs):
...
-        )
+        try:
+            self.get_db.write(
+                {'points': [point]},
+                {
+                    'db': kwargs.get('database') or self.db_name,
+                    'rp': kwargs.get('retention_policy'),
+                },
+            )
+        except InfluxDBServerError:
```
nemesisdesign, line 129: why catch only `InfluxDBServerError`? 

I tried stopping the server on purpose and the task just failed without retrying.

I think here we should catch just `Exception`.

nepython: I thought catching every exception might not be a good practice since if there is some other issue besides server load (say the `dict` we provide to the `write` method is of **invalid** format), it will stay hidden :no_mouth:. `InfluxDBClientError` was being raised in the case I mentioned but I agree if server goes down then the task won't be resilient which will defeat the purpose.

I am a bit unsure here, tbh as to what might be better for the long term. Currently, it will catch **every Exception**.

nepython:
**TODO:**
* [ ] https://github.com/openwisp/openwisp-monitoring/pull/190#pullrequestreview-464585622
* [ ] Write a test that ensures that `timestamp` is retained during retries.

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -117,13 +118,16 @@ def write(self, name, values, **kwargs):
...
+        try:
+            self.get_db.write(
+                {'points': [point]},
+                {
+                    'db': kwargs.get('database') or self.db_name,
+                    'rp': kwargs.get('retention_policy'),
+                },
+            )
+        except Exception:
+            raise TimeseriesWriteException
```
nemesisdesign, line 130: what about:

```python
except Exception as exception:
    logger.warning(f'got exception while writing to tsdb: {exception}')
    raise TimeseriesWriteException(exception)
```

This way we should see what's going on in the logs, please try it and adjust it so that we can clearly see what's wrong from the logs.

nepython:
> while testing, I tried to stop the timeseries DB to see if the device metric API write is really resilient but I found out some code needs to retrieve some data before writing: `_init_previous_data` fails.
> 
> How about trying the following?
> 
> After saving the device data, that last save gets stored in the cache as previous data for that device.
> On subsequent calls, the previous data is retrieved from the cache and only if not present from the timeseries DB
> 
> Can you try this and test the metric collection API with the timeseries DB temporarily stopped and let me know if it succeeds?

I have taken care of but I would like to note that still this doesn't cover all cases. For example, I noted below exception being raised on celery when `threshold_crossed` method (that means when `timeseries.read`) was called
Screenshot from 2020-08-11 19-00-14.

Similarly, charts too won't load when the tsdb server is down. There might be other cases too, I know you might be aware of them as they are obvious but I am noting nevertheless as to what can be achieved and what not after this **PR** :sweat_smile:.

I have tested that metric data is not lost when tsdb server is down. Here is how I tested it.
*Started everything celery worker/ beat, django runserver, OpenWRT instance.*
*Stopped InfluxDB*
`sudo systemctl stop influxdb`

**Celery started task retries on failure and displayed cause of failures too**
Screenshot from 2020-08-11 16-34-08
**Device admin was loading successfully with `STATUS` tab being loaded from cache**
**NOTE**: I have set cache timeout to 1 day.

*Waited for 15 minutes*
*started InfluxDB*
`sudo systemctl start influxdb`

Checked data of all metrics (ping, config_applied and resources) and device_data. No data was lost and celery retries were successful. Chart loaded correctly with no gaps.

I think this is what was needed for the issue, let me know if I tested incorrectly, thanks.

> * [ ] Write a test that ensures that timestamp is retained during retries.

I am unable to write a test for this :disappointed:.

openwisp_monitoring/monitoring/settings.py
```diff
@@ -2,3 +2,11 @@
 
 ADDITIONAL_CHARTS = getattr(settings, 'OPENWISP_MONITORING_CHARTS', {})
 ADDITIONAL_METRICS = getattr(settings, 'OPENWISP_MONITORING_METRICS', {})
+
+RETRY_OPTIONS = getattr(
+    settings,
+    'OPENWISP_MONITORING_RETRY_OPTIONS',
```
nemesisdesign, line 8: I think this should be called `OPENWISP_MONITORING_WRITE_RETRY_OPTIONS`, otherwise it can be misleading

openwisp_monitoring/device/base/models.py
```diff
@@ -97,7 +100,13 @@ def data(self):
         if self.__data:
             return self.__data
         q = device_data_query.format(SHORT_RP, self.__key, self.pk)
-        points = timeseries_db.get_list_query(q, precision=None)
+        try:
+            points = timeseries_db.get_list_query(q, precision=None)
```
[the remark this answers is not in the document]
nepython: Nice idea, thanks!

openwisp_monitoring/db/backends/influxdb/tests.py
```diff
@@ -240,3 +245,27 @@ def test_metric_write_microseconds_precision(self):
...
+
+    @patch.object(
+        InfluxDBClient, 'write', side_effect=InfluxDBServerError('Server error')
+    )
+    def test_write_retry(self, mock_write):
+        with self.assertRaises(TimeseriesWriteException):
+            timeseries_db.write('test_write', {'value': 1})
+        m = self._create_general_metric(name='Test metric')
+        with self.assertRaises(Retry):
+            m.write(1)
```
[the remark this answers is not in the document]
nepython: >I think it's here that you should do with patch.object(timeseries_db, 'write') as mock_write and then call mock_write.assert_called_with.

@nemesisdesign, actually this test is also covering two lines below in `timeseries_db.write` and hence I didn't mock it and went for another separate test :sweat_smile:.
https://github.com/openwisp/openwisp-monitoring/blob/bcf86f72950707cf4c54bbca06f8818798f083ef/openwisp_monitoring/db/backends/influxdb/client.py#L129-L131

>The test below is not very useful.
>
>I suggest also to freeze the time with freezun, make pass several minutes artificially (always with freezegun) and then ensure the time is the initial one. This is how the kind of rock solid test which gives us confidence when introducing changes.
```python
    @patch.object(timeseries_db, 'write')
    def test_timeseries_write_params(self, mock_write):
        m = self._create_general_metric(name='Test metric')
        with freeze_time('Jan 14th, 2020', auto_tick_seconds=15):
            # First tick, time should jump by 15 seconds
            first_time = now()
            # Two more ticks, metric.write and check_threshold, time should increment by 30s
            m.write(1)
            # Ensure clock is ticking on every call to datetime
            self.assertEqual(now(), first_time + timedelta(seconds=45))
        mock_write.assert_called_with(
            'test_metric',
            {'value': 1},
            database=None,
            retention_policy=None,
            tags={},
            timestamp='2020-01-14T00:00:15Z',  # this should only be incremented by 15s from frozen time
        )
```
I have added above test where clock will tick whenever datetime is called. So we are sure that even after clock has ticked, timestamp is retained (more realistic, imo).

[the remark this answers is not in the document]
nemesisdesign: The choice of adding a new test was the right one :+1:

openwisp_monitoring/db/backends/influxdb/tests.py
```diff
@@ -240,3 +246,32 @@ def test_metric_write_microseconds_precision(self):
...
+
+    @patch.object(timeseries_db, 'write')
+    def test_timeseries_write_params(self, mock_write):
+        m = self._create_general_metric(name='Test metric')
+        with freeze_time('Jan 14th, 2020', auto_tick_seconds=15):
+            # First tick, time should jump by 15 seconds
+            first_time = now()
+            # Two more ticks, metric.write and check_threshold, time should increment by 30s
+            m.write(1)
```
nemesisdesign, line 267: We want to be 100% sure, now and in the future, that when a write fails, it's retried (we already test this) **but when the write is retried, the time used for the write is the original one**.

This is the full scenario, but I don't see the retry mechanism involved in this test. So it looks like something important is missing to me.

I think it should be something like the following (**pseudocode alert!**):

```python
@freeze_time('Jan 14th, 2020'):
def test_timeseries_write_params(self, mock_write):
   m = self._create_general_metric(name='Test metric')
   with patch(MAKE_IT_FAIL) as whatever:
       m.write(1)
       # expect retry but find a way to not call the retry

   time_tick_10_minutes()  # do it manually, it's more explicit
   self.assertEqual(now(), first_time + timedelta(minutes=10))

   with patch.object(timeseries_db, 'write') as mock_write:
       trigger_retry()
       mock_write.assert_called_with(
           'test_metric',
           {'value': 1},
           database=None,
           retention_policy=None,
           tags={},
           
           # =================================================================
           # this must be the original time at the moment of the first failure
           # =================================================================
           timestamp='2020-01-14T00:00:00Z',
       )
```

openwisp_monitoring/db/backends/influxdb/tests.py
```diff
@@ -240,3 +247,42 @@ def test_metric_write_microseconds_precision(self):
...
+                self._retry_task(task_signature)
+            mock_write.assert_called_with(
+                'test_metric',
+                {'value': 1},
+                database=None,
+                retention_policy=None,
+                tags={},
+                # this should be the original time at the moment of first failure
+                timestamp='2020-01-14T00:00:00Z',
+            )
```
nemesisdesign, line 284: Great :+1:

PR #193, opened 2020-08-12
[refactor] Load BaseDevice using swappable

nemesisdesign:
Temporary solution, may be obsoleted by
https://github.com/wq/django-swappable-models/pull/23

### DIFF
--- a/openwisp_monitoring/device/models.py
+++ b/openwisp_monitoring/device/models.py
@@ -1,9 +1,10 @@
 from django.contrib.contenttypes.fields import GenericRelation
 from swapper import get_model_name, swappable_setting
 
-from openwisp_controller.config.models import Device as BaseDevice
-
 from .base.models import AbstractDeviceData, AbstractDeviceMonitoring
+from .utils import load_model_patched
+
+BaseDevice = load_model_patched('config', 'Device', require_ready=False)
 
 
 class DeviceData(AbstractDeviceData, BaseDevice):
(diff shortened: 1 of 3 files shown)


nepython:
Seems a good fix :smiley:. Can we add a regression test to be sure that extended models will work nicely if it's not too difficult.

nemesisdesign:
> Can we add a regression test to be sure that extended models will work nicely if it's not too difficult.

Any suggestion or idea on how to test this?

nepython:
> > Can we add a regression test to be sure that extended models will work nicely if it's not too difficult.
> 
> Any suggestion or idea on how to test this?

Well, yes a very rough idea on how I would have approached it. We can test it as how `django` tests it.
1. We create a new `test_app` with a `models.py`, containing the extended model but we disable auto loading for the model (https://github.com/django/django/blob/master/tests/migrations/models.py#L37)
2. Override settings for `INSTALLED_APPS` and `EXTENDED_APPS` for the test
3. Now migrate the changes within the test (https://github.com/django/django/blob/master/tests/migrations/test_commands.py#L32).
4. Reload the apps
5. Check if the extended model can be imported safely.
6. Run reverse migrations on completion / failure

:warning: **IMHO it is not worth the effort** as it seems now. :sweat_smile:

nemesisdesign:
>  **IMHO it is not worth the effort** as it seems now.

I agree, it's a lot of code for not much added benefit for us.
Let's merge the way it is, if in the future we feel the need to add another test we will.

Issue #173, opened 2020-07-27
[ux] Threshold value for resource metrics is weird


nemesisdesign:
We show the hardware resource monitoring charts (memory, cpu, storage) in percentages from 0% to 100% .

But the threshold values are 0.95, 0.8, 0.9 which is inconsistent.

Can we make the threshold values consistent? Can we use 95, 80 and 90?

nepython:
> But the threshold values are 0.95, 0.8, 0.9 which is inconsistent.
>
> Can we make the threshold values consistent? Can we use 95, 80 and 90?

That was what was being done originally (https://github.com/openwisp/openwisp-monitoring/pull/76#discussion_r426193740) :sweat_smile: .

In fact in order to have threshold values `0.95, 0.8, 0.9`, we had to change Threshold value field from Integer to Float.
This should be a doable, we would need to add one more migration for existing alertsettings. Let's do this after #185 which should be near complete hopefully and it would be more easier after it.

nepython:
I don't think we can do much about the old data. I read it at multiple places that `timeseries` data is not meant to be updated and thus tsdb are not optimized for the same. One such thread, https://community.influxdata.com/t/update-influxdb-measurements/8707

nemesisdesign:
> I don't think we can do much about the old data. I read it at multiple places that `timeseries` data is not meant to be updated and thus tsdb are not optimized for the same. One such thread, https://community.influxdata.com/t/update-influxdb-measurements/8707

I feared we'd end up deleting the old data, but it's fine, let's do it soon rather than later.

PR #199, opened 2020-08-18
[ux] Improve threshold values of reseource metrics from cent #173

nepython:
Checks:

:warning: This needs to be updated after #198 gets merged.


nepython:
>what happens to the old measurements collected? Do they have to be deleted?
>If yes, can you please write a data migration that does this (only for influxdb, you can use also the python-influxdb library directly if you need).
>This would help me to deploy it faster once ready.

We have a `delete_metric_data` method (it deletes only data not the metric object) in tsdb client that we use during tests which comes to the rescue :smile:.
```python
>>> from influxdb import InfluxDBClient
>>> client = InfluxDBClient(username='openwisp', password='openwisp', database='openwisp2')
>>> client.get_list_measurements()
[{'name': 'br_lan'}, {'name': 'cpu'}, {'name': 'disk'}, {'name': 'histogram'}, {'name': 'memory'}, {'name': 'ping'}, {'name': 'test-retry'}]  # before the migration
>>> client.get_list_measurements()
[{'name': 'br_lan'}, {'name': 'histogram'}, {'name': 'ping'}, {'name': 'test-retry'}]    # after the migration
```

Note: Please ignore the measurement `test-retry` (I use it to test things locally). Stating to avoid any confusion :sweat_smile:.

I also checked the Charts after running the migration and they seemed fine. Also, if you had any other influxdb specific refactoring in mind which would have required us to delete old data, now would be a good time to do it (I don't have any such refactoring to suggest right now).

openwisp_monitoring/monitoring/migrations/0021_resources_threshold_values.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 3.0.3 on 2020-08-17 10:18
+
+from django.db import migrations
+
+
+resources_metrics = {'memory': 95, 'disk': 80, 'cpu': 90}
```
nemesisdesign, line 6: take the value from the configuration please, avoid duplication or we'll face issues if those values are changed.

openwisp_monitoring/monitoring/migrations/0021_resources_threshold_values.py
```diff
@@ -0,0 +1,47 @@
...
+def standardize_alertsettings_threshold(apps, schema_editor):
+    AlertSettings = apps.get_model('monitoring', 'AlertSettings')
+    for alert_settings in AlertSettings.objects.all():
+        config_name = alert_settings.metric.configuration
+        if (
+            config_name not in resources_metrics
+            or alert_settings.custom_threshold is None
+        ):
+            continue
+        alert_settings.custom_threshold *= 100
```
nemesisdesign, line 18: this multiplication should be done only if the current value is not an integer (or beween 0 and 1), but I think we should do this by filtering in the query rather than with an if

openwisp_monitoring/monitoring/migrations/0021_resources_threshold_values.py
```diff
@@ -0,0 +1,47 @@
...
+
+from django.db import migrations
+
+
+resources_metrics = {'memory': 95, 'disk': 80, 'cpu': 90}
+
+
+def standardize_alertsettings_threshold(apps, schema_editor):
+    AlertSettings = apps.get_model('monitoring', 'AlertSettings')
+    for alert_settings in AlertSettings.objects.all():
```
nemesisdesign, line 11: filter only the metrics we're interested in, why looping over all?
filter also the values which are lower than 1

openwisp_monitoring/monitoring/migrations/0021_resources_threshold_values.py
```diff
@@ -0,0 +1,47 @@
...
+            continue
+        alert_settings.custom_threshold *= 100
+        if alert_settings.custom_threshold == resources_metrics[config_name]:
+            alert_settings.custom_threshold = None
+        alert_settings.save()
+
+
+def reverse_alertsettings_threshold(apps, schema_editor):
+    AlertSettings = apps.get_model('monitoring', 'AlertSettings')
+    for alert_settings in AlertSettings.objects.all():
```
nemesisdesign, line 26: same here, filter

openwisp_monitoring/monitoring/migrations/0021_resources_threshold_values.py
```diff
@@ -0,0 +1,47 @@
...
+        config_name = alert_settings.metric.configuration
+        if config_name not in resources_metrics:
+            continue
+        if alert_settings.custom_threshold is None:
+            alert_settings.custom_threshold = resources_metrics[config_name]
+        alert_settings.custom_threshold /= 100
```
nemesisdesign, line 32: this multiplication should be done only if the current value is not an integer (or beween 0 and 1), but I think we should do this by filtering in the query rather than with an if

openwisp_monitoring/monitoring/migrations/0022_delete_resources_metric_data.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.0.3 on 2020-08-20 14:10
+
+from django.db import migrations
+from openwisp_monitoring.db import timeseries_db
+
+
+def delete_resources_metric_data(apps, schema_editor):
+    Metric = apps.get_model('monitoring', 'Metric')
+    for metric in Metric.objects.all():
+        if metric.configuration in ['disk', 'memory', 'cpu']:
```
nemesisdesign, line 10: filter the metric by configuration rather than doing this if

PR #201, opened 2020-08-19
[ux] Help user to find out what's wrong #172

### LINKED ISSUE #172: [ux] Help user to find out what's wrong
Now that we have multiple checks and multiple metrics that generate alerts and change the health status to "problem", whenever a user accesses the device page and sees "problem" as health status, the first question that will arise to them is:

**What's the problem?**

We have to improve the UI so it becomes immediately evident where the problem lies.

Here's a possible solution (it's just a suggestion, if you think about a better solution let's discuss):

- we can add a new field like `health_status` in `openwisp_monitoring.device.admin.DeviceAdmin`
- this new field would give information about where the problem lies, this information can be retrieved by querying metrics related to the device which have the `is_healthy` field set to `False`, so the field could display the list of metrics which have problems
- this field could be shown only when the monitoring status is not ok (to save unneeded DB queries)

nepython:
Checks:

Screenshot from 2020-08-19 20-46-10

Closes #172

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -126,7 +126,7 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
             )
         return ctx
 
-    def metric_health(self, obj):
+    def health_checks(self, obj):
         metric_rows = []
         for metric in DeviceData(pk=obj.pk).metrics.all():
             # Skip metrics which don't affect device health
@@ -138,8 +138,10 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
                 f'alt="health"> {metric.name}</li>'
             )
         return format_html(
-            mark_safe(f'<ul class="metric_health">{"".join(metric_rows)}</ul>'),
+            mark_safe(f'<ul class="health_checks">{"".join(metric_rows)}</ul>'),
         )
+
+    health_checks.short_description = _('health checks')
 
     def health_status(self, obj):
         return format_html(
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/device/admin.py
```diff
@@ -126,6 +126,21 @@ def get_extra_context(self, pk=None):
             )
         return ctx
 
+    def metric_health(self, obj):
+        metric_rows = []
+        for metric in DeviceData.objects.get(pk=obj.pk).metrics.all():
```
nemesisdesign, line 131: I think there's no need to generate another query to retrieve DeviceData, which is just a proxy model BTW and hence does not have particular DB content not available to device.

You should be able to just do:

`DeviceData(pk=obj.pk).metrics.all()`

This will be enough for the proxy model instance to know what query to make.

```python
for metric in DeviceData.objects.get(pk=obj.pk).metrics.all():
```

openwisp_monitoring/device/admin.py
```diff
@@ -139,6 +154,12 @@ def get_form(self, request, obj=None, **kwargs):
         """
         Adds the help_text of DeviceMonitoring.status field
         """
+        health_field = 'metric_health' in self.fields
+        if obj and obj.monitoring.status != 'problem' and health_field:
+            self.fields.remove('metric_health')
+        elif obj and obj.monitoring.status == 'problem' and health_field is False:
+            self.fields.insert(self.fields.index('last_ip'), 'metric_health')
+            self.readonly_fields.append('metric_health')
```
nemesisdesign, line 162: the approach is not correct.

You have to understand that both `self.fields` and `self.readonly_fields` are class attributes which point to lists, which are mutable data structures. When you modify at runtime a list which is a class attribute (hence a global value), you are modifying it for the whole application.

This approach can lead to inconsistent behavior.

You should instead override these two methods:

- https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_fields
- https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_readonly_fields

nepython:
> Looks good, but there's some aspects which need improvement. See below.
> 
> Please also take a look at the number of SQL queries in the django debug toolbar, ensure there's not an exaggerated increment of queries after the patch is applied ok?

With this PR, **38** queries,
Screenshot from 2020-08-20 16-03-51

On `dev`, **37** queries,
Screenshot from 2020-08-20 15-52-48

I think addition of only one extra query is good enough :sweat_smile:

openwisp_monitoring/device/admin.py
```diff
@@ -145,6 +160,20 @@ def get_form(self, request, obj=None, **kwargs):
         )
         return super().get_form(request, obj, **kwargs)
 
+    def get_fields(self, request, obj=None):
+        if not obj or obj.monitoring.status != 'problem':
```
nemesisdesign, line 164: I think we should show this also when the status is critical, it's helpful, moreover the system can be configured to have other metrics which are critical and may trigger this status.

openwisp_monitoring/device/admin.py
```diff
@@ -145,6 +160,20 @@ def get_form(self, request, obj=None, **kwargs):
...
+            return self.fields
```
nemesisdesign, line 165: when working with the openwisp django admin, especially DeviceAdmin which is extended by different modules, never ever brutally override a method like you're doing here: **you're throwing away several customizations that have been done elsewhere**.

You have to always call `super()`, get the result, and do your operations on that result.

Here's what how the add device page looked before your changes:

Screenshot from 2020-08-20 21-15-07

After these changes, fields which should not be shown in the add page are instead shown (uuid, management ip, health status):

Screenshot from 2020-08-20 21-14-33

Please pay attention to this! If I didn't notice now we'd introduce another bug.

As you can see, even if the test coverage of each module is pretty high, passing tests + high coverage is not a guarantee of being bug free. Manual testing is **very** important.

PR #214, opened 2020-08-23
[chores] Made unregistering "default" notification type resilient to failure

pandafy:
Added object notification widget for DeviceAdmin which enables user
to disable notification for that device.

Checks:

### DIFF
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 from django.core.cache import cache
+from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/device/settings.py
```diff
@@ -58,3 +72,5 @@ def get_health_status_labels():
 MAC_VENDOR_DETECTION = getattr(
     settings, 'OPENWISP_MONITORING_MAC_VENDOR_DETECTION', True
 )
+
+update_object_notification_widget()
```
nemesisdesign, line 76: It would be better to do this in the app.ready() method, so we are sure it's executed only once.
I would do it in openwisp-controller.

nemesisdesign (approved):

PR #216, opened 2020-08-25
[fix] Fixed create_ping and create_config migrations #205

### LINKED ISSUE #205: [migrations] check 0003_create_ping.py is flawed
While migrating the system used at the University of Urbino by @cappe87 and @AlessiaVe, we incurred in a flaw in this data migration: 

https://github.com/openwisp/openwisp-monitoring/blob/8ff5b28f2d5db587c50f0680c0994fbce173b2d0/openwisp_monitoring/check/migrations/0003_create_ping.py#L6-L14

The problem is that `auto_create_ping` assumes we can use the concrete model, but this is not possible on older versions, because some fields are missing and trying to exeucte that will raise an exception.

A possible solution could be to add the possibility of passing the `Check` model as argument, if supplied, this model should be used and the concrete model should not be loaded:

https://github.com/openwisp/openwisp-monitoring/blob/8ff5b28f2d5db587c50f0680c0994fbce173b2d0/openwisp_monitoring/check/tasks.py#L37-L54

nepython:
Checks:

Closes #205

### DIFF
--- a/openwisp_monitoring/check/tasks.py
+++ b/openwisp_monitoring/check/tasks.py
@@ -10,6 +10,10 @@ from swapper import load_model
 logger = logging.getLogger(__name__)
 
 
+def get_check_model():
+    return load_model('check', 'Check')
+
+
 @shared_task
 def run_checks():
     """
@@ -19,8 +23,13 @@ def run_checks():
     This allows to enqueue all the checks that need to be performed
     and execute them in parallel with multiple workers if needed.
     """
-    Check = load_model('check', 'Check')
-    iterator = Check.objects.filter(is_active=True).only('id').values('id').iterator()
+    iterator = (
+        get_check_model()
+        .objects.filter(is_active=True)
+        .only('id')
+        .values('id')
+        .iterator()
+    )
     for check in iterator:
         perform_check.delay(check['id'])
 
@@ -31,9 +40,8 @@ def perform_check(uuid):
     Retrieves check according to the passed UUID
     and calls ``check.perform_check()``
     """
-    Check = load_model('check', 'Check')
     try:
-        check = Check.objects.get(pk=uuid)
+        check = get_check_model().objects.get(pk=uuid)
     except ObjectDoesNotExist:
         logger.warning(f'The check with uuid {uuid} has been deleted')
         return
@@ -43,12 +51,14 @@ def perform_check(uuid):
 
 
 @shared_task
-def auto_create_ping(model, app_label, object_id):
+def auto_create_ping(
+    model, app_label, object_id, check_model=None, content_type_model=None
+):
     """
     Called by django signal (dispatch_uid: auto_ping)
     registered in check app's apps.py file.
     """
-    Check = load_model('check', 'Check')
+    Check = check_model or get_check_model()
     ping_path = 'openwisp_monitoring.check.classes.Ping'
     has_check = Check.objects.filter(
         object_id=object_id, content_type__model='device', check=ping_path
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/check/tasks.py
```diff
@@ -57,17 +65,19 @@ def auto_create_ping(model, app_label, object_id):
     if has_check:
         return
     ct = ContentType.objects.get(app_label=app_label, model=model)
-    check = Check(name='Ping', check=ping_path, content_type=ct, object_id=object_id)
+    check = Check(
+        name='Ping', check=ping_path, content_type_id=ct.id, object_id=object_id
```
nepython, line 69: I am not sure why content_type didn't work with `apps.get_model('check', 'Check')` :/
I was getting this strange error, which I couldn't make understand. Anyways, content_type_id works and we achieve the same objective as prior.
```python
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/check/migrations/0003_create_ping.py", line 15, in create_device_ping
    check_model=Check,
  File "/home/hardik/.local/lib/python3.6/site-packages/celery/local.py", line 191, in __call__
    return self._get_current_object()(*a, **kw)
  File "/home/hardik/.local/lib/python3.6/site-packages/celery/app/task.py", line 392, in __call__
    return self.run(*args, **kwargs)
  File "/home/hardik/openwisp/openwisp-monitoring/openwisp_monitoring/check/tasks.py", line 66, in auto_create_ping
    check = Check(name='Ping', check=ping_path, content_type=ct, object_id=object_id)
  File "/home/hardik/.local/lib/python3.6/site-packages/django/db/models/base.py", line 482, in __init__
    _setattr(self, field.name, rel_obj)
  File "/home/hardik/.local/lib/python3.6/site-packages/django/db/models/fields/related_descriptors.py", line 219, in __set__
    self.field.remote_field.model._meta.object_name,
ValueError: Cannot assign "<ContentType: config | device>": "Check.content_type" must be a "ContentType" instance.
```

openwisp_monitoring/check/tasks.py
```diff
@@ -57,17 +65,19 @@ def auto_create_ping(model, app_label, object_id):
     if has_check:
         return
     ct = ContentType.objects.get(app_label=app_label, model=model)
```
nemesisdesign, line 67: can you load ContentType with the migration framework as well please?
And test it again.

nepython: It worked, thanks :+1:

nepython:
> How did you test this?

I started with a clean db, created admin and then ran migrations. After all the migrations were run, I simply ran the below command to unmigrate,
```shell
./tests/manage.py migrate check 0002
```
Now I opened django shell, created four new devices.
Next step, ran the below command to test the fix,
```shell
./tests/manage.py migrate check
```
So everything forward and reverse migrations worked out fine for me. I am using `postgresql` with `postgis` as the relational database.

Also I once unmigrated and migrated back to apply your suggestion related to passing `ContentType` model, then I was using this very database so I think it is works fine even without a clean db for me.

nemesisdesign:
> > How did you test this?
> 
> I started with a clean db, created admin and then ran migrations. After all the migrations were run, I simply ran the below command to unmigrate,
> 
> ```shell
> ./tests/manage.py migrate check 0002
> ```
> 
> Now I opened django shell, created four new devices.
> Next step, ran the below command to test the fix,
> 
> ```shell
> ./tests/manage.py migrate check
> ```
> 
> So everything forward and reverse migrations worked out fine for me. I am using `postgresql` with `postgis` as the relational database.
> 
> Also I once unmigrated and migrated back to apply your suggestion related to passing `ContentType` model, then I was using this very database so I think it is works fine even without a clean db for me.

Thanks for the explanation.

PR #217, opened 2020-08-26
[chores] Correct use of swappable in migrations and include migrations in qa-checks

### LINKED ISSUE #210: [swappable] Migrations not following swappable docs
Some parts of the migrations are not following the indications of the [swappable documentation regarding migrations](https://github.com/wq/django-swappable-models#migration-scripts).

eg:

- to=settings.MONITORING_METRIC_MODEL
- options={'abstract': False, 'swappable': 'MONITORING_CHART_MODEL',},

There are more.

Can we also take advantage to remove the exclusion of the QA checks from the migrations as did here?
https://github.com/openwisp/openwisp-notifications/pull/115/files#diff-380c6a8ebbbce17d55d50ef17d3cf906L7-L21

nepython:
Checks:

Closes #210


nepython:
Right now, running the command `openwisp-qa-format` does not take care of line length so I had to do it manually. Unsure but I think this should be configurable.

nepython:
> Can you fix the conflict that appeared after merging another PR please?

Done :smiley:

PR #227, opened 2020-09-01
[deps] Upgrade InfluxDB python library to 5.3 #191

### LINKED ISSUE #191: [deps] Upgrade influxdb python library
Upgrade influxdb python library to latest available version.

nepython:
Closes #191

Checks:

- [ ] ~~I have written new test cases to avoid regressions (if necessary)~~
- [ ] ~~I have updated the documentation (e.g. README.rst)~~

Read the change log at https://github.com/influxdata/influxdb-python/blob/master/CHANGELOG.md#v530---2020-04-10.
There seem no breaking change and did a manual testing of the influxdb methods we are using in the tsdb client.

Also, there seems to be an improvement where we can pass in the time-precision of the timestamp we want in the latest version, noting here for future refs.

:warning: To be merged after #228.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 openwisp-controller~=0.7.post1
-influxdb>=5.2,<5.3
+influxdb~=5.3
 django-notifications-hq>=1.6,<1.7
 django-celery-email>=3.0.0,<3.1
 djangorestframework>=3.11,<3.12
--- a/openwisp_monitoring/device/tests/test_api.py
+++ b/openwisp_monitoring/device/tests/test_api.py
@@ -284,17 +284,17 @@ class TestDeviceApi(DeviceMonitoringTestCase):
                 '2',
                 '0.4',
                 '0.1',
-                '2',
-                '1',
+                '2.0',
+                '1.0',
                 '9.73',
-                '0',
+                '0.0',
                 '8.27',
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 openwisp-controller~=0.7.post1
-influxdb>=5.2,<5.3
+influxdb>=5.3,<5.4
```
nemesisdesign, line 2: let's switch to the form `~=5.3`, which is better when using a tight range like this.

If you could also convert the other requirements but in a separate (and possibly previous) commit it would be great!

nepython: > If you could also convert the other requirements but in a separate (and possibly previous) commit it would be great!

Created #228 and there is a query there.

.travis.yml
```diff
@@ -7,8 +7,9 @@ python:
   - "3.7"
 
 env:
-  - DJANGO="django>=2.2,<3.0"
-  - DJANGO="django>=3.0,<3.1"
+  - DJANGO="django~=2.2"
+  - DJANGO="django~=3.0"
+  - DJANGO="django~=3.1"
```
nemesisdesign, line 12: this change shouldn't be done here

nemesisdesign (changes requested):

PR #229, opened 2020-09-02
[fix] Ensure background tasks are executed on transaction commit

nemesisdesign:
Many tasks were being called without transaction.on_commit,
which does not guarantee the tasks will be executed after
changes have been committed to the database, which in turn
leads to inconsistent results.


nepython:
Thanks for letting me know, read more about it in django docs https://docs.djangoproject.com/en/3.1/topics/db/transactions/#performing-actions-after-commit

PR #235, opened 2020-09-23
[fix] Config Applied checks would be skipped when device config status is errored #195

### LINKED ISSUE #195: [bug] config applied should ignore error status
To be done only after https://github.com/openwisp/openwisp-controller/issues/191.

The config applied check should not run if the config status is `errored`.

nepython:
Closes #195
This should be now taken care off by openwisp-controller (https://github.com/openwisp/openwisp-controller/pull/265).

Checks:

- [ ] ~I have updated the documentation (e.g. README.rst)~

### DIFF
--- a/openwisp_monitoring/check/tests/test_models.py
+++ b/openwisp_monitoring/check/tests/test_models.py
@@ -142,6 +142,39 @@ class TestModels(TestDeviceMonitoringMixin, TransactionTestCase):
         self.assertEqual(dm.status, 'problem')
         self.assertEqual(Notification.objects.count(), 1)
 
+    @patch('openwisp_monitoring.check.settings.AUTO_PING', False)
+    def test_config_error(self):
+        """
+        Test that ConfigApplied checks are skipped when device config status is errored
+        """
+        self._create_admin()
+        self.assertEqual(Check.objects.count(), 0)
+        self._create_config(status='error', organization=self._create_org())
+        dm = Device.objects.first().monitoring
+        dm.update_status('ok')
+        self.assertEqual(Check.objects.count(), 2)
+        self.assertEqual(Metric.objects.count(), 0)
+        self.assertEqual(AlertSettings.objects.count(), 0)
+        check = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        with freeze_time(now() - timedelta(minutes=10)):
+            check.perform_check()
+        # Check needs to be run again without mocking time for threshold crossed
+        self.assertEqual(check.perform_check(), 0)
+        self.assertEqual(Metric.objects.count(), 1)
+        m = Metric.objects.first()
+        self.assertEqual(AlertSettings.objects.count(), 1)
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'problem')
+        self.assertEqual(Notification.objects.filter(actor_object_id=m.id).count(), 0)
+        # Check config recovery
+        dm.device.config.set_status_applied()
+        # We are once again querying for the check to override the cached property check_instance
+        check = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        check.perform_check()
+        dm.refresh_from_db()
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/check/tests/test_models.py
```diff
@@ -142,6 +142,24 @@ def test_config_modified_device_problem(self):
...
+        self._create_config(status='error', organization=self._create_org())
+        self.assertEqual(Check.objects.count(), 2)
+        self.assertEqual(Metric.objects.count(), 0)
+        self.assertEqual(AlertSettings.objects.count(), 0)
+        check = Check.objects.filter(check=self._CONFIG_APPLIED).first()
+        with freeze_time(now() - timedelta(minutes=10)):
+            check.perform_check()
+        self.assertEqual(Metric.objects.count(), 0)
+        self.assertEqual(AlertSettings.objects.count(), 0)
+        self.assertIsNone(check.perform_check())
```
nemesisdesign, line 161: This test passes also without the fix above. Try it.
I just tried it and it's not failing.

You must always ensure the test fails without your patch.
The best thing is to add the test before the fix, alternatively comment out the fix temporarily.

nemesisdesign (changes requested):
Welcome back! See my comment below.

nemesisdesign (commented):
I realized something. We'll end up with the health status being OK when what we want in the truth is to avoid the redundant alert.

I'm thinking if we can get the status change without the alert.

That would be the ideal solution. This way we would have the notification from openwisp-controller, but the health status would change to PROBLEM instead of remaining OK.

A simple approach which doesn't require changing a lot of other models would be something like:

```python
# if the device is down or does not have a config
# do not run config applied check
if (
    self.related_object.monitoring.status in ['critical', 'unknown']
    or not hasattr(self.related_object, 'config')
):
    return

metric = self._get_metric()
result = int(self.related_object.config.status == 'applied')

# if the device config is in error status we don't need to notify
# the user (because that's already done by openwisp-controller)
# but we need to ensure health status will be changed
if self.related_object.config.status == 'error':
    store = False
    metric.check_threshold(notify=False)

if store:
    metric.write(
        result, retention_policy=SHORT_RP,
    )
return result
```

This would require adding a `notify` argument to the `check_threshold`.

alternatively, we could add the `notify` argument also to metric.write() which wouold in turn pass it to `check_threshold`, the resulting code would then be simpler:

```python
# if the device is down or does not have a config
# do not run config applied check
if (
    self.related_object.monitoring.status in ['critical', 'unknown']
    or not hasattr(self.related_object, 'config')
):
    return

metric = self._get_metric()
result = int(self.related_object.config.status == 'applied')

# if the device config is in error status we don't need to notify
# the user (because that's already done by openwisp-controller)
# but we need to ensure health status will be changed
notify = self.related_object.config.status != 'error':

if store:
    metric.write(
        result, retention_policy=SHORT_RP,
    )

return result
```

This may also have the additional benefit that we will be notified when the configuration goes back to normal! Which is really nice, we don't have this in openwisp-controller, so the two would complement each other.

I'm in favour of the second approach, what do you think?

nepython:
Yes, the second approach seems more convenient as it would provide us with more flexibility with (when to and when not to send) alerts (neglecting the threshold if need be like here) during a metric write in the longer run. I will work upon it :)

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -145,7 +145,7 @@ def content_type_key(self):
         except AttributeError:
             return None
 
-    def check_threshold(self, value, time=None, retention_policy=None):
+    def check_threshold(self, value, time=None, retention_policy=None, send_alert=True):
```
nepython, line 148: named this parameter as `send_alert` instead of `notify` to avoid any confusion with a signal of the same name.

nemesisdesign (approved):
Tested it and it works, great work @nepython :+1:

PR #236, opened 2020-09-24
[fix] Support for tolerance in both directions #234

### LINKED ISSUE #234: [monitoring] Support for tolerance in both directions
Right now, in the following situation this is what happens:

- set tolerance of a device to 15 minutes
- the device goes down, after 15 minutes an alert is emitted
- the device goes up: a notification is immediately generated - **but it would be better to respect the tolerance here as well**

I suggest to proceed with writing the tests first.
Look at the existing tests:

- https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py#L194-L208
- probably this test will need to be modified: https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py#L243-L256 - if I remember correctly it expects the alert to be generated straightaway, but once we implement this change the alert should not be generated unless the number of minutes specified in tolerance has passed

PabloCastellano:
Fixes #234

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -154,25 +154,81 @@ class AbstractMetric(TimeStampedEditableModel):
             alert_settings = self.alertsettings
         except ObjectDoesNotExist:
             return
-        crossed = alert_settings._is_crossed_by(value, time, retention_policy)
+        # crossed = alert_settings._is_crossed_by(value, time, retention_policy)
+        crossed = alert_settings._value_crossed(value)
+        tolerance = alert_settings.tolerance
         first_time = False
-        # situation has not changed
-        if (not crossed and self.is_healthy) or (crossed and self.is_healthy is False):
-            return
-        # problem: not within threshold limit
-        elif crossed and self.is_healthy in [True, None]:
-            if self.is_healthy is None:
-                first_time = True
-            self.is_healthy = False
-            notification_type = f'{self.configuration}_problem'
-        # ok: returned within threshold limit
-        elif not crossed and self.is_healthy is False:
-            self.is_healthy = True
-            notification_type = f'{self.configuration}_recovery'
-        # First metric write within threshold
-        elif not crossed and self.is_healthy is None:
-            self.is_healthy = True
-            first_time = True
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,25 +154,81 @@ def check_threshold(self, value, time=None, retention_policy=None):
...
-            self.is_healthy = False
-            notification_type = f'{self.configuration}_problem'
-        # ok: returned within threshold limit
-        elif not crossed and self.is_healthy is False:
-            self.is_healthy = True
-            notification_type = f'{self.configuration}_recovery'
-        # First metric write within threshold
-        elif not crossed and self.is_healthy is None:
-            self.is_healthy = True
-            first_time = True
```
nemesisdesign, line 175: Are you sure is necessary to change this part?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -550,32 +607,25 @@ def _time_crossed(self, time):
         threshold_time = timezone.now() - timedelta(minutes=self.tolerance)
         return time < threshold_time
 
-    def _is_crossed_by(self, current_value, time=None, retention_policy=None):
-        """ do current_value and time cross the threshold? """
-        value_crossed = self._value_crossed(current_value)
-        if value_crossed is NotImplemented:
-            raise ValueError('Supplied value type not suppported')
-        if not value_crossed:
```
nemesisdesign, line 558: maybe it's enough to just add ` and self.tolerance == 0` and make sure the rest of the code of this function works also when the value does not cross the threshold, that is:

- if tolerance is not zero
- return False only if the all the recent values do not cross the threshold

Right now instead, the code returns False as soon as hits `not value_crossed`

PabloCastellano: I tried that approach in the beginning and it didn't seem to work but I will consider reviewing it again

nemesisdesign:
Closed by #240, thanks @PabloCastellano.

PR #237, opened 2020-09-26
[admin] Removing monkey patching in device admin #202

### LINKED ISSUE #202: [admin] Conflicting approach in extending DeviceAdmin
I noticed we use 2 different approaches in extending DeviceAdmin at the same time:

we create a new class, unregister the admin from openwisp-controller and register the extended class in openwisp-monitoring, but then we also monkeypatch the class we just defined.
https://github.com/openwisp/openwisp-monitoring/blob/8da5b200baef91954564516c9549683c7ffe912d/openwisp_monitoring/device/admin.py#L170-L203

I didn't notice this before, @nepython can you help me to fix this? We should be able to just rewrite all the monkey patching in the class definition above those lines linked.

nepython:
Closes #202

Checks:

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -110,6 +110,18 @@ class MetricInline(NestedGenericStackedInline):
 
 class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
     change_form_template = 'admin/config/device/change_form.html'
+    list_filter = ['monitoring__status'] + BaseDeviceAdmin.list_filter
+    list_select_related = ['monitoring'] + list(BaseDeviceAdmin.list_select_related)
+    list_display = list(BaseDeviceAdmin.list_display)
+    list_display.insert(list_display.index('config_status'), 'health_status')
+    readonly_fields = ['health_status'] + BaseDeviceAdmin.readonly_fields
+
+    class Media:
+        js = MetricAdmin.Media.js + (
+            'monitoring/js/percircle.js',
+            'monitoring/js/alert-settings.js',
+        )
+        css = {'all': ('monitoring/css/percircle.css',) + MetricAdmin.Media.css['all']}
 
     def get_extra_context(self, pk=None):
         ctx = super().get_extra_context(pk)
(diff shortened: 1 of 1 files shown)


nepython:
Tests are failing on travis but passing locally for me. Will need to check why so.

openwisp_monitoring/device/admin.py
```diff
@@ -160,56 +167,68 @@ def get_form(self, request, obj=None, **kwargs):
...
+        list_select_related.append('monitoring')
+        return list_select_related
 
-DeviceAdmin.get_inlines = device_admin_get_inlines
+    def get_list_filter(self, request):
+        list_filter = list(super().get_list_filter(request))
+        list_filter.insert(
+            0, 'monitoring__status',
+        )
+        return list_filter
```
nemesisdesign, line 200: why do you override the method instead of just doing:

```python
list_filter = ['monitoring__status'] + BaseDeviceAdmin.list_filter
```

Let's aim for less code.

openwisp_monitoring/device/admin.py
```diff
@@ -160,56 +167,68 @@ def get_form(self, request, obj=None, **kwargs):
...
         if not obj or obj.monitoring.status in ['ok', 'unknown']:
             return readonly_fields
-        readonly_fields = list(readonly_fields)
         readonly_fields.append('health_checks')
         return readonly_fields
 
+    def get_list_display(self, request):
+        list_display = list(super().get_list_display(request))
+        list_display.insert(list_display.index('config_status'), 'health_status')
+        return list_display
```
nemesisdesign, line 188: same here

openwisp_monitoring/device/admin.py
```diff
@@ -160,56 +167,68 @@ def get_form(self, request, obj=None, **kwargs):
         return super().get_form(request, obj, **kwargs)
 
     def get_fields(self, request, obj=None):
-        fields = super().get_fields(request, obj)
+        fields = list(super().get_fields(request, obj))
         if not obj or obj.monitoring.status in ['ok', 'unknown']:
             return fields
-        fields = list(fields)
+        fields.insert(fields.index('last_ip'), 'health_status')
```
nemesisdesign, line 173: this won't be executed if the if above is triggered, which was not how it worked before (always added), please rever this

openwisp_monitoring/device/admin.py
```diff
@@ -160,56 +167,68 @@ def get_form(self, request, obj=None, **kwargs):
...
         fields.insert(fields.index('health_status') + 1, 'health_checks')
         return fields
 
     def get_readonly_fields(self, request, obj=None):
-        readonly_fields = super().get_readonly_fields(request, obj)
+        readonly_fields = list(super().get_readonly_fields(request, obj))
+        readonly_fields.append('health_status')
```
nemesisdesign, line 179: add this as a class attribute as for `list_filter`, no need to do it here,

nemesisdesign (changes requested):
On the right track but you went a bit too far with the changes. 

When you do these kind of changes, try to avoid introducing code changes that can change the behavior of the application. See below.

nemesisdesign (commented):
Thanks @nepython :+1:

nemesisdesign (approved):

PR #246, opened 2020-10-24
[monitoring] Rethink when check threshold #223

### LINKED ISSUE #223: [monitoring] Rethink when it's best to check threshold (eg: after metric is written)
We recently changed the metric write operation to be run in the background #190, but the threshold is checked right after the write operation is sent to be performed to the background worker:

https://github.com/openwisp/openwisp-monitoring/blob/f6c75913ecadf8d17e4ba3d1a9c248cf11ff5689/openwisp_monitoring/monitoring/base/models.py#L207-L220

The `check_threshold` method handles a case in which past data is used for the check, but the code has not been updated after the changes to #190, so several inconsistencies may appear:

- the check threshold operation may be performed before the value is written to the tsdb, which can cause inconsistencies in behavior since the code has been written to be executed after the values have been written
- the check threshold may fail if in that moment the tsdb is experiencing issues

Taking this into account we should think about a solution.
I think we should find a way to ensure the check threshold operation is done after the metric has been written successfully to the tsdb.

nepython:
Closes #223

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -2,6 +2,7 @@ import json
 import logging
 from datetime import date, datetime, timedelta
 
+from dateutil.parser import parse as parse_date
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.fields import GenericForeignKey
@@ -146,6 +147,14 @@ class AbstractMetric(TimeStampedEditableModel):
         except AttributeError:
             return None
 
+    def _get_time(self, time):
+        """
+        If time is a string, convert it to a datetime
+        """
+        if isinstance(time, str):
+            return parse_date(time)
+        return time
+
     def check_threshold(self, value, time=None, retention_policy=None, send_alert=True):
         """
         checks if the threshold is crossed
@@ -155,8 +164,7 @@ class AbstractMetric(TimeStampedEditableModel):
             alert_settings = self.alertsettings
         except ObjectDoesNotExist:
             return
-        if isinstance(time, str):
-            time = utc.localize(datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%fZ'))
+        time = self._get_time(time)
         crossed = alert_settings._is_crossed_by(value, time, retention_policy)
         first_time = False
         # situation has not changed
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -29,7 +29,7 @@ from ..configuration import (
     get_metric_configuration,
 )
 from ..exceptions import InvalidChartConfigException, InvalidMetricConfigException
-from ..signals import post_metric_write, pre_metric_write, threshold_crossed
+from ..signals import pre_metric_write, threshold_crossed
 from ..tasks import timeseries_write
 
 User = get_user_model()
@@ -155,6 +155,8 @@ class AbstractMetric(TimeStampedEditableModel):
             alert_settings = self.alertsettings
         except ObjectDoesNotExist:
             return
+        if isinstance(time, str):
+            time = utc.localize(datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%fZ'))
         crossed = alert_settings._is_crossed_by(value, time, retention_policy)
         first_time = False
         # situation has not changed
@@ -213,13 +215,17 @@ class AbstractMetric(TimeStampedEditableModel):
             database=database,
             retention_policy=retention_policy,
         )
-        timeseries_write.delay(name=self.key, values=values, **options)
-        post_metric_write.send(**signal_kwargs)
         # check can be disabled,
         # mostly for automated testing and debugging purposes
-        if not check:
-            return
-        self.check_threshold(value, time, retention_policy, send_alert)
+        if check:
+            options['check_threshold_kwargs'] = {
+                'value': value,
+                'time': time,
+                'retention_policy': retention_policy,
+                'send_alert': send_alert,
+            }
+            options['metric_pk'] = self.pk
+        timeseries_write.delay(name=self.key, values=values, **options)
 
     def read(self, **kwargs):
         """ reads timeseries data """
--- a/openwisp_monitoring/monitoring/tasks.py
+++ b/openwisp_monitoring/monitoring/tasks.py
@@ -1,13 +1,30 @@
 from celery import shared_task
+from django.core.exceptions import ObjectDoesNotExist
+from swapper import load_model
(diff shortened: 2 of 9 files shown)


openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
```diff
@@ -148,18 +156,24 @@ def test_general_check_threshold_crossed_for_long_time(self):
...
+            m.refresh_from_db()
             self.assertFalse(m.is_healthy)
             self.assertEqual(Notification.objects.count(), 1)
 
         with self.subTest('Test recovery notification after tolerance is passed'):
             with freeze_time(ten_minutes_after):
                 m.write(50)
+            # TODO: This sleep needs to be removed, currently it has been added
+            # to avoid this test being flaky due to async task.
+            sleep(1)
```
nemesisdesign, line 175: during tests celery tasks are executed synchronously: https://github.com/openwisp/openwisp-monitoring/blob/3d55db0a94933515ce410989c9e8c2ca24b61088/tests/openwisp2/settings.py#L165-L170

Therefore this sleep should not be needed.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -213,13 +213,18 @@ def write(
...
+        if check:
+            options['threshold_dict'] = {
+                'value': value,
+                'time': time,
+                'retention_policy': retention_policy,
+                'send_alert': send_alert,
+            }
+            options['metric_pk'] = self.pk
+        timeseries_write.delay(name=self.key, values=values, **options)
+        post_metric_write.send(**signal_kwargs)
```
nemesisdesign, line 227: I think now is a good time to move this signal to the background task and execute it after the metric has been written.
We should also update the related docs to explain it will fire only after a successful metric write.
The way it's implemented now is really not very useful.

openwisp_monitoring/monitoring/tasks.py
```diff
@@ -1,13 +1,35 @@
...
-def timeseries_write(self, name, values, **kwargs):
+def timeseries_write(self, name, values, metric_pk=None, threshold_dict=None, **kwargs):
     """
     write with exponential backoff on a failure
     """
     timeseries_db.write(name, values, **kwargs)
+    if not metric_pk or not threshold_dict:
+        return
+    try:
+        metric = load_model('monitoring', 'Metric').objects.get(pk=metric_pk)
```
nemesisdesign, line 22: the try/except block should include only this line:

```python
try:
    DB query
except ObjectDoesNotExist:
    return
```

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -213,13 +213,18 @@ def write(
...
         )
-        timeseries_write.delay(name=self.key, values=values, **options)
-        post_metric_write.send(**signal_kwargs)
         # check can be disabled,
         # mostly for automated testing and debugging purposes
-        if not check:
-            return
-        self.check_threshold(value, time, retention_policy, send_alert)
+        if check:
+            options['threshold_dict'] = {
```
nemesisdesign, line 219: `threshold_dict` should be named `check_threshold_kwargs`, which indicates these are the keyword arguments for the `check_threshold()` method.

openwisp_monitoring/monitoring/tasks.py
```diff
@@ -1,13 +1,35 @@
...
+        time = threshold_dict.pop('time')
+        if isinstance(time, str):
+            time = (
+                datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%fZ')
+                if 'Z' in time
+                else datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%f')
+            )
+            time = pytz.utc.localize(time)
```
nemesisdesign, line 30: is there a simpler way to do this? Isn't the time we get always supposed to be in UTC?

Also, I think it would be better to move it to `check_threshold`
I wold simply call `metric.check_threshold(**check_threshold_kwargs)` and let `check_threshold` figure out what it has to do.

nemesisdesign (changes requested):
Good progress, see my comments below. I have not tested this yet.

openwisp_monitoring/monitoring/base/models.py
nepython: Agreed :+1:

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -155,6 +155,8 @@ def check_threshold(self, value, time=None, retention_policy=None, send_alert=Tr
             alert_settings = self.alertsettings
         except ObjectDoesNotExist:
             return
+        if isinstance(time, str):
+            time = utc.localize(datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%fZ'))
```
nemesisdesign, line 159: the only potentially dangerous line I see is this one.

nemesisdesign (commented):
Looks good, I need to find some time to test it manually to ensure everything keeps working normally.

openwisp_monitoring/monitoring/base/models.py
nepython: To be honest, I agree with you. I feel slightly uncomfortable with timezones and their implications (as to where UTC is to be used and where the local timezone) right now. What I am sure of is that this datetime object needed to be time aware, `Z` implied UTC timezone hence I went with it. Please let me know if I am misunderstanding this anywhere or if you meant dangerous in a different sense, thanks.

requirements.txt
```diff
@@ -7,3 +7,4 @@ django-cache-memoize~=0.1
 django-nested-admin~=3.3.2
 swapper~=1.1
 netaddr~=0.8
+python-dateutil>=2.7,<3.0
```
nemesisdesign, line 10: It's possible that the library is already being pulled because it's a dependency of some other package, but I added it anyway with a relaxed version range to follow the **"explicit is better than implicit"** approach.
The version range is relaxed because it's a very stable library, at the same time we won't accept an eventual 3.0 automatically without manual testing.

nemesisdesign (commented):
I changed the date parsing code, see the commit message for an explanation:
https://github.com/openwisp/openwisp-monitoring/pull/246/commits/c05ac7fd40af27cdba9e2576ab70494c33ff1e4a

Tested manually, haven't found issues :+1:.

Let's wait for the build to finish, once it's successful we can merge.

requirements.txt
nepython: Understood the approach from the commit message and it seems much more readable and simple. I didn't use this library earlier, thanks for introducing an application for the same :)

nemesisdesign (approved):

PR #253, opened 2020-11-18
[migrations] Squash migrations #169

### LINKED ISSUE #169: [migrations] Squash migrations
Once we're ready for the first release and once all the existing deployed systems using the development version have been updated, we should squash the migrations to make our life easier.

Once the squashing is done, ensure the migrations can be ran backward without errors, see https://github.com/openwisp/openwisp-monitoring/pull/216#pullrequestreview-476335066 for more info.

nepython:
Closes #169
* All 23 migrations in `monitoring` app squashed to 1
* Squashed the first two migrations in `check` app. I think we can also replace migration 0004 in the `check` app by modifying this squashed initial migration. I wasn't very sure about it so didn't go ahead with the same.
* Squashed the two migrations in `device_monitoring` app.

Checks:

- [ ]  ~~I have written new test cases to avoid regressions (if necessary)~~
- [ ] ~~I have updated the documentation (e.g. README.rst)~~


nemesisdesign (approved):
I think it's good enough, even if the migrations in check are not all squashed as you pointed out, but since we have some data migrations in the middle and since these are not many migration files, I think it's good enough, let's move on! :blush: :+1: 

PS: I tested it by creating new migrations and rolling back and it's working, so we can proceed.

PR #254, opened 2020-12-11
[deps] Removed django-celery-email from requirements

pandafy:
Checks:

- [ ] ~I have manually tested the proposed changes~
- [ ] ~I have written new test cases to avoid regressions (if necessary)~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -223,9 +223,6 @@ Configure celery (you may use a different broker if you want):
         },
     }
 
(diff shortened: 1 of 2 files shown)


nemesisdesign (commented):
Let's wait for this since there are systems deployed in prod which rely on it.
I think we'll have to move it to ansible-openwisp2, it's good to send emails in the background anyway.

nemesisdesign (approved):

PR #257, opened 2020-12-18
[fix] Fix random SynchronousOnlyOperation in celery tasks #249

### LINKED ISSUE #249: [bug] SynchronousOnlyOperation in perform_check
Stack trace:

This exception comes up periodically in production:

```
SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
  File "celery/app/trace.py", line 412, in trace_task
    R = retval = fun(*args, **kwargs)
  File "celery/app/trace.py", line 704, in __protected_call__
    return self.run(*args, **kwargs)
  File "openwisp_monitoring/check/tasks.py", line 44, in perform_check
    check = get_check_model().objects.get(pk=uuid)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 425, in get
    num = len(clone)
  File "django/db/models/query.py", line 269, in __len__
    self._fetch_all()
  File "django/db/models/query.py", line 1308, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "django/db/models/query.py", line 53, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "django/db/models/sql/compiler.py", line 1154, in execute_sql
    cursor = self.connection.cursor()
  File "django/utils/asyncio.py", line 24, in inner
    raise SynchronousOnlyOperation(message)
```

My guess is that this happens because [ansible-openwisp2 deploys celery with gevent](https://github.com/openwisp/ansible-openwisp2/blob/master/templates/supervisor/celery.j2#L4), which runs an eventloop, but I don't understand why the [SynchronousOnlyOperation](https://docs.djangoproject.com/en/3.0/ref/exceptions/#synchronousonlyoperation) exception is raised only some times.

More investigation:

- [StackOverflow](https://stackoverflow.com/questions/65326908/synchronousonlyoperation-from-celery-task-using-gevent-execution-pool)
- [Django Developers](https://groups.google.com/g/django-developers/c/nMiAvLm_Pvk/m/3ri8I-L3AAAJ)

PS: it happens also in `timeseries_write`.

nemesisdesign:
When using gevent as a concurrency pool for celery,
SynchronousOnlyOperation exception may be run,
but not always, only in a minority of cases,
so this patch tries to run the code normally and
only if the exception is raised it will re-run the
code through asgiref.sync.sync_to_async.

Fixes #249

Checks:

- [ ] I have manually tested the proposed changes - I am testing this
- [ ] ~~I have written new test cases to avoid regressions (if necessary)~~ N/A
- [ ] ~~I have updated the documentation (e.g. README.rst)~~ N/A

### DIFF
--- a/openwisp_monitoring/check/tasks.py
+++ b/openwisp_monitoring/check/tasks.py
@@ -7,6 +7,8 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ObjectDoesNotExist
 from swapper import load_model
 
+from ..utils import fix_async
+
 logger = logging.getLogger(__name__)
 
 
@@ -41,11 +43,11 @@ def perform_check(uuid):
     and calls ``check.perform_check()``
     """
     try:
-        check = get_check_model().objects.get(pk=uuid)
+        check = fix_async(lambda: get_check_model().objects.get(pk=uuid))
     except ObjectDoesNotExist:
         logger.warning(f'The check with uuid {uuid} has been deleted')
         return
-    result = check.perform_check()
+    result = fix_async(lambda: check.perform_check())
     if settings.DEBUG:  # pragma: nocover
         print(json.dumps(result, indent=4, sort_keys=True))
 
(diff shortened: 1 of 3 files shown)


nemesisdesign:
Solved by openwisp/ansible-openwisp2#254.

Issue #203, opened 2020-08-20
Prepare 0.1 release


nemesisdesign:
Prerequisites:

- [x] #207 
- [x] #206
- [x] #205
- [x] #204 
- [x] #201 
- [x] #202 
- [x] #199 
- [x] #182
- [x] #169 
- [x] #167
- [x] first release of openwisp-notifications

After this list is completed, we can go ahead and issue the first release! :clap:

PabloCastellano:
@nemesisdesign Nice to see that the release date is so close now! I think you already know it, but GitHub milestones are useful to track the pending issues for a release

nemesisdesign:
@PabloCastellano thanks for the suggestion, I created the milestone in this repo: https://github.com/openwisp/openwisp-monitoring/milestone/1
For the issues in the other repos we'll rely on the link here I guess.

nemesisdesign:
Only 4 issues left for 0.1 release!

nepython:
> Only 4 issues left for 0.1 release!

This is so close for an initial release now, I will try to work upon lua-monitoring issue right now :smile:

nemesisdesign:
Finally :+1:

PR #258, opened 2020-12-20
[feature] Auto install monitoring scripts #204

### LINKED ISSUE #204: [feature] Auto install the monitoring scripts
We need to automatically install the monitoring scripts, at least until we [convert lua-monitoring into an openwrt package](https://github.com/openwisp/lua-monitoring/issues/6), but I'm not sure when we'll be able to do that, I'd love to do it, but I just don't have the time now.

Here's how I would do it:

- define some internal variables which describe each of the aspects of the monitoring template: https://gist.github.com/nemesisdesign/cb005dc80a75536cf8537105c3038a71
- we should name the variable something like `TEMPLATE_OPENWISP_MONITORING_01`, `TEMPLATE_NETJSON_MONITORING_01`, `TEMPLATE_CRONTAB_MONITORING_01` (01 gives us the possibility to introduce new versions if needed and update the template)
- we should then create a data migration which looks for a template with some of the aspects which shouldn't change, file names for example, if this template already exists, the data migration should do nothing, if it doesn't exist, the data migration should create it (`organization=None` to make it available to all organizations) and flag it as default

We should then document this aspect, we should state that:

- the monitoring script is required to make some metrics and checks work (point to the metrics and checks with links)
- explain very briefly how the script works, just to give an idea to our users
- list the OpenWRT dependencies needed for the monitoring scripts to work (check `/usr/sbin/update-openwisp-packages` in the linked gist to find out)

### LINKED ISSUE #223: [monitoring] Rethink when it's best to check threshold (eg: after metric is written)

nepython:
Closes #204

Checks:


nepython:
I tested the migration by both applying and unapplying it in the following scenarios:
1) When the migration was applied on a database not containing any default template and without the terms `/usr/sbin/openwisp-monitoring`, `/usr/sbin/netjson-monitoring` mentioned in any template's config, a new default template is created.
2) When there was a default template with the two terms `/usr/sbin/openwisp-monitoring`, `/usr/sbin/netjson-monitoring` mentioned in that template's config, no new template is created after running the migration as required.
3) When we run the migration backwards, it won't delete the created template (if any).

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,101 @@
...
+from django.db.models import Q
+
+
+def migrate_data(apps, schema_editor):
+    Template = apps.get_model('config', 'Template')
+    if Template.objects.filter(
+        Q(config__contains='/usr/sbin/openwisp-monitoring')
+        & Q(config__contains='/usr/sbin/netjson-monitoring'),
+        default=True,
+    ).count():
```
nemesisdesign, line 15: I suggest using `exists()` instead of `count()`, it's more efficient.

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,101 @@
...
+        return
+    default_template = Template(
```
nemesisdesign, line 17: can you check if it's possible somehow to force a specific UUID?
If we can force the UUID, then we can automatically upgrade the template as well and document this behaviour suggesting that if they want to do customizations they should create a new template.

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+TEMPLATE_OPENWISP_MONITORING_01 = OrderedDict(
+    {
+        "path": "/usr/sbin/openwisp-monitoring",
+        "mode": "0744",
+        "contents": "uuid=$(uci get openwisp.http.uuid)\nkey=$(uci get openwisp.http.key)\nbase_url=$(uci get openwisp.http.url)\nverify_ssl=$(uci get openwisp.http.verify_ssl)\nincluded_interfaces=$(uci get openwisp.monitoring.included_interfaces)\nurl=\"$base_url/api/v1/monitoring/device/$uuid/?key=$key\"\ndata=$(/usr/sbin/netjson-monitoring \"$included_interfaces\")\nif [ \"$verify_ssl\" = 0 ]; then\n    curl_command='curl -k'\nelse\n    curl_command='curl'\nfi\n# send data via POST\n$curl_command -H \"Content-Type: application/json\" \\\n              -X POST \\\n              -d \"$data\" \\\n              -v $url\n",  # noqa
+    }
+)
+
```
nemesisdesign, line 66: I'd remove this blank line

nemesisdesign (commented):
Look on the right track, I left some comments below.

openwisp_monitoring/device/migrations/0002_create_template.py
nepython: We can force a specific UUID while creating this template object but I didn't understood the use case.
It can be done in the following manner,
```python
default_template = Template(
    pk=uuid.UUID('66b1a809-2d83-4e33-8b2d-64aa5d4036a9'),  # This UUID would then be common for monitoring templates across openwisp instances
    ...
```
Let me know if the above is what you were suggesting and I remain unsure on how we can use the UUID of templates to upgrade them (so if possible a brief explanation for it would be helpful).

nemesisdesign: @nepython the lua scripts are evolving rapidly and will need to be updated over time so it would be good to be able to update the lua code.

But users could customize that code so we should let them know that if they want to do that, they should create another template (so we do not overwrite their changes).

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,103 @@
...
+    }
+)
+TEMPLATE_POST_RELOAD_HOOK_01 = OrderedDict(
+    {
+        "path": "/etc/openwisp/post-reload-hook",
+        "mode": "0744",
+        "contents": "#!/bin/sh\ntouch /etc/crontabs/root\n/etc/init.d/cron start\n/usr/sbin/update-openwisp-packages\n/usr/sbin/openwisp-monitoring\n",  # noqa
+    }
+)
```
nemesisdesign, line 103: store all these variables in `device/migrations/__init__.py`

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,103 @@
...
+    Template = apps.get_model('config', 'Template')
+    if Template.objects.filter(
+        Q(config__contains='/usr/sbin/openwisp-monitoring')
+        & Q(config__contains='/usr/sbin/netjson-monitoring'),
+        default=True,
+    ).exists():
+        return
+    default_template = Template(
+        # The UUID has been fixed so that the template could be auto-updated
+        pk=uuid.UUID('66b1a809-2d83-4e33-8b2d-64aa5d4036a9'),
```
nemesisdesign, line 20: store this UUID in `device/migrations/__init__.py` so it can be called from future migrations.

Let's make it also obviously artificial, so it will be easily recognizable by us when debugging, eg: `uuid.UUID('00000000-defa-defa-defa-000000000000')`

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,103 @@
...
+from django.db import migrations
+from django.db.models import Q
+
+
+def migrate_data(apps, schema_editor):
+    Template = apps.get_model('config', 'Template')
+    if Template.objects.filter(
+        Q(config__contains='/usr/sbin/openwisp-monitoring')
+        & Q(config__contains='/usr/sbin/netjson-monitoring'),
+        default=True,
```
nemesisdesign, line 15: let's remove this, the monitoring template may not be enabled by default

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,103 @@
...
+    ).exists():
+        return
+    default_template = Template(
+        # The UUID has been fixed so that the template could be auto-updated
+        pk=uuid.UUID('66b1a809-2d83-4e33-8b2d-64aa5d4036a9'),
+        name='default template',
```
nemesisdesign, line 21: Let's change this to `Monitoring (default)`

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -0,0 +1,103 @@
...
...\n            elseif vpn_interfaces[name] then\n                netjson_interface.type = 'virtual'\n            else\n                netjson_interface.type = 'other'\n            end\n        end\n        if include_stats[name] then\n            netjson_interface.statistics = interface.statistics\n        end\n        addresses = get_addresses(name)\n        if next(addresses) then\n            netjson_interface.addresses = addresses\n        end\n        info = get_interface_info(name, netjson_interface)\n        if info.stp ~= nil then\n            netjson_interface.stp = info.stp\n        end\n        table.insert(interfaces, netjson_interface)\n        -- DNS info is independent from interface\n        if info.dns_servers then\n            array_concat(info.dns_servers, dns_servers)\n        end\n        if info.dns_search then\n            array_concat(info.dns_search, dns_search)\n        end\n    end\nend\n\nif next(interfaces) ~= nil then\n    netjson.interfaces = interfaces\nend\nif next(dns_servers) ~= nil then\n    netjson.dns_servers = dns_servers\nend\nif next(dns_search) ~= nil then\n    netjson.dns_search = dns_search\nend\n\nprint(cjson.encode(netjson))\n",  # noqa
```
nemesisdesign, line 73: please update this with the contents from this file: https://github.com/openwisp/lua-monitoring/blob/master/netjson-monitoring.lua
I found out this version I passed you was not up to date.

nemesisdesign (changes requested):
Great progress @nepython, see my comments below.

nemesisdesign (commented):
Looks good! Added some minor improvements.
I will be testing it a bit more before merging it.

nemesisdesign (commented):
Found another issue which I corrected in https://github.com/openwisp/openwisp-monitoring/pull/258/commits/88709b4216c4cf0b30784b3e01b043e8459cf1b2, pay attention Hardik.. that missing s caused the traffic data collection to not work.

nepython:
> Found another issue which I corrected in [88709b4](https://github.com/openwisp/openwisp-monitoring/commit/88709b4216c4cf0b30784b3e01b043e8459cf1b2), pay attention Hardik.. that missing s caused the traffic data collection to not work.

Thanks. I am not sure though how the s got missed by me in this PR since it definitely is there in your template. Will pay more attention.

nemesisdesign (changes requested):
An important change was just merged in the monitoring scripts:

- https://github.com/openwisp/lua-monitoring/pull/23
After pulling the latest changes in this branch,  can you update the code of `/usr/bin/netjson-monitoring` to reflect these changes please?

nepython:
Updated the PR. If I am correct then this PR should lead to the closure of issue https://github.com/openwisp/openwisp-monitoring/issues/203 and milestone *0.1 release* should be achieved :blush:

nemesisdesign:
Now only waiting for https://github.com/openwisp/lua-monitoring/pull/24 to get merged.

PR #269, opened 2021-02-16
[monitoring] Move check into monitoring #13

### LINKED ISSUE #13: [monitoring] Move check into monitoring app
We should move check in `openwisp_monitoring.monitoring` since it makes sense that it's part of the main monitoring solution. It's still good to keep the code separated, but it should be configured in a way that it shows up under the monitoring app and not in its own app.

We should make sure any DB migration generated by this change performs the migration in a backward compatible way.

Purhan:
Closes #13

Checks:


nemesisdesign (commented):
We have to move the whole check app code but also make sure the DB content is ported.

Purhan:
Closing this for now, I will reopen later if required.

PR #275, opened 2021-03-11
[change] ConfigApplied triggered on config_status_changed instead of config_modified

nemesisdesign:
It does not make sense to trigger this check when the config is modified,
because the result will be always that config.status == 'modified'.
Moreover, due to the recent changes in openwisp-controller,
this action can now be executed also when a config has been just created
and m2m templates are added to it.

Rather than changing the code to check whether the signal is being
sent from m2m changes or not, it is a lot more useful to just
switch to the config_status_changed signal, which will trigger
the check also when the configuration is flagged as applied by
the system, making the check a lot more proactive.

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -431,6 +431,11 @@ is running and applying configuration changes in a timely manner.
 You may choose to disable auto creation of this check by using the
 setting `OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK <#OPENWISP_MONITORING_AUTO_DEVICE_CONFIG_CHECK>`_.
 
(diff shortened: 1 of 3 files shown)


nemesisdesign:
My analysis was slightly wrong, in the sense that the check should not be executed when a device is created, unless there's a mistake in the code, which was exactly what triggered the issue which prompted me to change this, however, I think the change makes sense anyway, apart from checking periodically, we will trigger this check when a change of the config status is detected, rather than when the config is modified. Probably that was the initial plan but due to misunderstanding it was implemented using config_modified instead of config_status_changed.

PR #284, opened 2021-03-30
[change] Upated monitoring template

R9295:
I will test later today with a device, opened a PR in case anyone can test earlier. 
Checks:


nemesisdesign (commented):
Great, can you please update it so we have this latest version in?
https://github.com/openwisp/lua-monitoring/commit/c7eda03aa49bb683feabdc88878c1eafe835590e

pandafy (commented):
Found these differences in the script added in monitoring template and  [lua-monitoring](https://github.com/openwisp/lua-monitoring/blob/c7eda03aa49bb683feabdc88878c1eafe835590e/netjson-monitoring.lua): https://www.diffchecker.com/YGX9dFIB

The differences are just in quotes(`'` instead of `"`) but that does not make any logical difference.

nemesisdesign (changes requested):
Since we already released a migration file with this version in the 0.1.0 version, we have to define a new version, eg: `TEMPLATE_OPENWISP_MONITORING_02`  and create a new data migraton.

nemesisdesign (changes requested):
The migration is failing because of a style issue.

I just had to change another thing:
https://github.com/openwisp/lua-monitoring/commit/2ae2def791fbbeaead26116e5a3a805310098517

After testing on different systems I come to the conclusion that the best default setting is to monitor all interfaces, because we do not know how users will name interfaces so they may expect to create interfaces named in any way and have these monitored automatically, but with the current implementation that would not happen and they get disappointed.

So, do you have time to update `TEMPLATE_NETJSON_MONITORING_02` again but also update this line: https://github.com/openwisp/openwisp-monitoring/blob/b2b3e18346cc01e0f046cc51c74bd1771e07b673/openwisp_monitoring/device/migrations/0002_create_template.py#L48 to have `*` instead of the interface names?

BTW I noticed only now that line is not easily versionable but right now for me it's fine to just change that one since this module is not being widely used yet.

devkapilbansal:
@nemesisdesign If I am not wrong, this PR is no longer needed. Since we are going to package the scripts on openwrt

devkapilbansal:
Closing this as no longer needed. 
https://github.com/openwisp/openwrt-openwisp-monitoring/tree/master is already updated.

PR #292, opened 2021-04-08
[feature] Clear management_ip when device goes offline #286

### LINKED ISSUE #286: [change] Remove management_ip from device when status changes to offline
We need to introduce a setting, set to `True` by default, called `OPENWISP_MONITORING_AUTO_CLEAR_MANAGEMENT_IP`, which when `True` automatically clears the `management_ip` of a device when its monitoring status passes to `critical`.
We can do this directly in `AbstractDeviceMonitoring.update_status()`.

Reason: whenever a single OpenVPN instance or pool of instances using the same DHCP server is shared among different organizations, the same IP can be assigned first to the device of org A, then to the device of org B, but the management IP saved in device of org A is not cleared so we end up with two devices having the same management IP.
OpenWISP controller already does this cleanup, but it only works within the same organization, different organizations can have the same management IP because they could use different VPN instances.

devkapilbansal:
Closes #286

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -804,6 +804,18 @@ In case you just want to change the colors used in a chart here's how to do it:
         }
     }
 
+``OPENWISP_MONITORING_AUTO_CLEAR_MANAGEMENT_IP``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+This setting allows you to automatically clear management_ip of a device
+when it goes offline. It is enabled by default.
+
 Registering / Unregistering Metric Configuration
 ------------------------------------------------
 
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -253,6 +253,11 @@ class AbstractDeviceMonitoring(TimeStampedEditableModel):
         self.status = value
         self.full_clean()
         self.save()
+        # clear device management_ip when device is offline
+        if self.status == 'critical' and app_settings.AUTO_CLEAR_MANAGEMENT_IP:
+            self.device.management_ip = None
+            self.device.save(update_fields=['management_ip'])
+
         health_status_changed.send(sender=self.__class__, instance=self, status=value)
 
     @property
--- a/openwisp_monitoring/device/tests/test_models.py
+++ b/openwisp_monitoring/device/tests/test_models.py
@@ -600,6 +600,17 @@ class TestDeviceMonitoring(CreateConnectionsMixin, BaseTestCase):
         self.assertEqual(dm.status, 'critical')
         self.assertIsNone(dm.device.management_ip)
 
+    @patch('openwisp_monitoring.device.settings.AUTO_CLEAR_MANAGEMENT_IP', False)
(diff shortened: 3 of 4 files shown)


openwisp_monitoring/device/base/models.py
```diff
@@ -253,6 +253,11 @@ def update_status(self, value):
         self.status = value
         self.full_clean()
         self.save()
+        # clear device management_ip when device is offline
+        if self.status == 'critical' and app_settings.AUTO_CLEAR_MANAGEMENT_IP:
+            self.device.management_ip = None
+            self.device.save()
```
nemesisdesign, line 259: Try:

`self.device.save(update_fields=['management_ip'])`

It's more efficient.

tests/openwisp2/settings.py
```diff
@@ -144,6 +144,7 @@
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 OPENWISP_MONITORING_MANAGEMENT_IP_ONLY = False
+OPENWISP_MONITORING_AUTO_CLEAR_MANAGEMENT_IP = True
```
nemesisdesign, line 147: This line should not be needed because it's the default value, or what am I missing?

README.rst
```diff
@@ -804,6 +804,19 @@ In case you just want to change the colors used in a chart here's how to do it:
...
++--------------+-------------+
+| **type**:    | ``bool``    |
++--------------+-------------+
+| **default**: | ``True``    |
++--------------+-------------+
+
+This setting allows you to automatically clear management_ip of a device
+when it goes offline. It is enabled by default.
+
+
```
nemesisdesign, line 819: please avoid this extra blank line

nemesisdesign (changes requested):
Looks good @devkapilbansal, see my comments below.

PS: also change the prefix of your commit to `[feature]` please.

tests/openwisp2/settings.py
devkapilbansal: You're right. I should remove this

openwisp_monitoring/device/base/models.py
devkapilbansal: I was unaware of this. Thanks :smile:

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -590,6 +590,16 @@ def test_ok_problem_problem_problem_ok(self):
         load.check_threshold(80)
         self.assertEqual(dm.status, 'ok')
 
+    def test_management_ip_clear_device_offline(self):
```
devkapilbansal, line 593: Should there be a test case too to ensure management_ip is not cleared when `AUTO_CLEAR_MANAGEMENT_IP` is false?

pandafy: Yes, that would be great to have.

nemesisdesign (approved):
Very good work @devkapilbansal!

I changed a minor detail in the tests: https://github.com/openwisp/openwisp-monitoring/pull/292/commits/7f7b25f8e393dcccdee70e3e97d1986b30024d86
Please always ensure the data in the test is somewhat realistic, `0.0.0.0` does not make sense as a `management_ip` and even if for the sake of the test it doesn't change whether the value is realistic or not, it does affect the readability of the code, moreover, in the eye of a potential user who may be a network engineer who is inspecting our code, it can make a difference in the way they judge our work.

PR #294, opened 2021-05-02
[feature] Chart mobile (LTE/5G/UMTS/GSM) signal strength #270

### LINKED ISSUE #270: [feature] Chart signal strength (LTE/UMTS/etc.)
Modem-manager interfaces are used for 3G/LTE routers and provide information regarding the protocols used in the mobile industry.

[OpenWISP Monitoring is already are able to collect information in the device status data](https://github.com/openwisp/openwisp-monitoring/commit/fb50d3dabbaf12fdc5b47f89a50507c2c0d25554) (please take an indepth look at this link), which is then displayed in the device status tab.

Now we need to add logic to the device metric view which is also able to log the signal strength (RSSI) information in a chart, as we already do for
the device charts.

The only difficult thing is that the RSSI value depends on the mobile protocol being used, because the router could be using
LTE, UMTS or other protocols depending on different factors.

**Key points:**

- only 3 charts
- each chart will have max 2 different lines shown
- the access technology chart does not use lines but bar color to indicate access technology used

**Signal strength chart:**

- RSRP (5G/4G) or RSCP (UMTS) can be labeled as Signal Power
- RSSI, which is available for 2G (GPRS), 3G (UMTS, EVDO, CDMA1X), 5G/4G can be labeled as Signal Strength

**Signal quality chart:**

- 4G and 5G SNR can be grouped under Signal to Noise Ratio (4G/5G SNR)
- 3G RSCP and 4G RSRQ can be grouped under Signal Quality (3G RSCP, 4G/5G RSRQ)

**Access technology chart:**

A fixed length bar chart (like reachability) indicating the access technology (GPRS, 3G, 4G, 5G, etc).

A color is associated to each number, but the user will be able to see the exact access technology when hovering with the mouse on the bar (this will help to clarify which specific access tech is used when there can be different ones for the same generation of mobile access tech, like in the  case of 3G)

We could use a color scale like the one below:

Screenshot from 2021-04-16 20-06-52

Useful Resources:

- https://www.telcoantennas.com.au/blog/guide-to-mobile-networks/4g-lte-signal-strength-reference-guide/
- https://www.telcoantennas.com.au/blog/guide-to-mobile-networks/3g-umts-signal-strength-reference-guide/​

purhan:
Checks:

Closes #270

### DIFF
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -193,6 +193,52 @@ class DeviceMetricView(GenericAPIView):
         data = self.instance.data
         ct = ContentType.objects.get_for_model(Device)
         for interface in data.get('interfaces', []):
+            if 'mobile' in interface:
+                for iftype in interface['mobile']['signal'].keys():
+                    # create signal strength chart
+                    signal_data = interface['mobile']['signal'][iftype]
+                    signal_power = signal_strength = None
+                    if iftype in ['lte', '5g']:
+                        signal_power = signal_data['rsrp']
+                    elif iftype == 'umts':
+                        signal_power = signal_data['rscp']
+                    if iftype in ['cdma1x', 'evdo', 'gsm', 'lte', 'umts']:
+                        signal_strength = signal_data['rssi']
+                    if signal_strength is not None and signal_power is not None:
+                        extra_values = {'signal_power': signal_power}
+                        name = f'{iftype} signal strength'
+                        metric, created = Metric._get_or_create(
+                            object_id=pk,
+                            content_type=ct,
+                            configuration='signal_strength',
+                            name=name,
+                            key=iftype,
+                        )
+                        metric.write(signal_strength, extra_values=extra_values)
+                        if created:
+                            self._create_signal_strength_chart(metric)
+                    # create signal quality chart
+                    snr = signal_quality = None
+                    if iftype in ['lte', '5g']:
+                        snr = signal_data['snr']
+                        signal_quality = signal_data['rsrq']
+                    elif iftype == 'evdo':
+                        snr = signal_data['sinr']
+                    if iftype == 'umts':
+                        signal_quality = signal_data['rscp']
+                    if snr is not None and signal_quality is not None:
+                        extra_values = {'snr': snr}
+                        name = f'{iftype} signal quality'
+                        metric, created = Metric._get_or_create(
+                            object_id=pk,
+                            content_type=ct,
+                            configuration='signal_quality',
+                            name=name,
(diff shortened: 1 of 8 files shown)


openwisp_monitoring/monitoring/configuration.py
```diff
@@ -174,6 +174,56 @@
...
+        'name': '{name}',
+        'key': '{key}',
+        'field_name': 'signal_strength',
+        'related_fields': ['signal_power'],
+        'charts': {
+            'signal_strength': {
+                'type': 'scatter',
+                'title': _('Signal Strength: {metric.key}'),
+                'label': _('Signal Strength'),
+                'colors': DEFAULT_COLORS,
```
Purhan, line 188: Wanted some help here. Without this line, the chart shows the same color for both lines. I observed that other charts (like RTT) show different colors for each line without adding this explicitly. What could possibly be wrong?
<details open>
<summary>Screenshot</summary>

</details>

Purhan:
@nemesisdesign This is how these charts look currently (populated with arbitrary values):
<details>
<summary>Signal Strength</summary>

</details>
<details>
<summary>Signal Quality</summary>

</details>
<details>
<summary>Access Technology</summary>

</details>

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -97,7 +97,8 @@
                 options.marker = {
                     cmax: config.max,
                     cmin: config.min,
-                    colorbar: {title: config.label},
+                    colorbar:
+                        config.hide_colorbar === true ? null : {title: config.label},
```
Purhan, line 101: I introduced this new parameter, because I think it doesn't make much sense to keep the color bar in the `access technology` chart. Tell me if this is not desirable.

openwisp_monitoring/device/api/views.py
```diff
@@ -194,6 +195,61 @@ def _write(self, request, pk):
         ct = ContentType.objects.get_for_model(Device)
         for interface in data.get('interfaces', []):
             ifname = interface['name']
+            if 'mobile' in interface:
```
nemesisdesign, line 198: can you try to move this new block of code to its own method like we do with `_write_cpu`, `_write_disk` and `_write_memory`?

nemesisdesign (changes requested):
great progress @Purhan! See my comment below.

openwisp_monitoring/device/api/views.py
Purhan: Done

nemesisdesign (approved):

PR #295, opened 2021-05-11
[Fix] map tooltip and popup not working

pniaps:
In my environment i have a first location of type Polygon and two more of type Point.
In the map, the tooltip and popup only works only in markers of type Point, not in Polygon, also only works if the point is not inside the polygon.

Before:

https://user-images.githubusercontent.com/7120526/117864427-13fab200-b295-11eb-8cdd-6aed2c00e258.mp4

After:

https://user-images.githubusercontent.com/7120526/117864453-1b21c000-b295-11eb-9933-3af0986c1ec4.mp4

Now both are working.

I've also formatted the js code.

Checks:


nemesisdesign (commented):
I still need to test it but looks good at first glance, :+1: .

nemesisdesign (approved):
Thank you! :+1:

PR #302, opened 2021-05-27
[change] Add support for new OpenWRT Packages in Backwards Compatible Manner

### LINKED ISSUE #296: [cleanup] Remove auto-install script
Remove migration to add scripts on backend and document how to use OpenWRT package

devkapilbansal:
Closes #296

Checks:


README.rst
```diff
@@ -1233,21 +1233,16 @@ Example usage:
...
-are required to make the `checks <https://github.com/openwisp/openwisp-monitoring#available-checks>`_ and
-`metrics <#openwisp_monitoring_metrics>`_ work.
+You can install ``openwisp_monitoring`` and ``netjson_monitoring`` package from `openwrt-openwisp-monitoring <https://github.com/openwisp/openwrt-openwisp-monitoring/tree/master#compiling-openwisp-netjson-monitoring>`_
 
-The ``netjson-monitoring`` script collects the required data from the openwrt device in realtime. This
-data is then sent by the ``openwisp-monitoring`` script to the server in the form of JSON data via SSL.
-All the dependencies are updated and installed (if needed) by ``update-openwisp-packages`` script.
-The OpenWRT dependencies needed for the monitoring scripts to work are ``libubus-lua``, ``lua-cjson`` and
-``rpcd-mod-iwinfo``.
+These packages are required to make the `checks <https://github.com/openwisp/openwisp-monitoring#available-checks>`_ and
```
pandafy, line 1241:
```suggestion
These packages are required to make the `checks <#available-checks>`_ and
```

openwisp_monitoring/device/migrations/0003_remove_migration_template.py
```diff
@@ -0,0 +1,26 @@
...
+
+def delete_migration(apps, schema_editor):
+    Template = apps.get_model('config', 'Template')
+    templates = Template.objects.filter(
+        Q(config__contains='/usr/sbin/openwisp-monitoring')
+        & Q(config__contains='/usr/sbin/netjson-monitoring'),
+    )
+
+    if templates.exists():
+        templates.delete()
```
pandafy, line 15: You can directly execute delete query without checking if the queryset exists. It should not raise an error.

openwisp_monitoring/device/migrations/0003_remove_migration_template.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.1.2 on 2020-12-20 06:15
+
+from django.db import migrations
+from django.db.models import Q
+
+
+def delete_migration(apps, schema_editor):
```
pandafy, line 7: Let's rename use a more descriptive name like `delete_monitoring_templates`.

openwisp_monitoring/device/migrations/0003_remove_migration_template.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.1.2 on 2020-12-20 06:15
```
pandafy, line 1: Rename the file to `0003_remove_monitoring_templates.py`

pandafy (changes requested):

openwisp_monitoring/device/migrations/0003_remove_monitoring_templates.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('device_monitoring', '0002_create_template'),
+    ]
+
+    operations = [
+        migrations.RunPython(delete_monitoring_templates, reverse_code=migrations.RunPython.noop)
+    ]
```
nemesisdesign, line 24: let's not do this. Let's instead change the data migrations which introduce the template so they do nothing. Otherwise we risk that some people will upgrade the system and not read this, get their monitoring template removed without having installed the new package.

We should advise users using previous versions to remove this template manually and install the new openwrt package in the README.

nemesisdesign (changes requested):

openwisp_monitoring/device/migrations/0003_remove_monitoring_templates.py
pandafy: > We should advise users using previous versions to remove this template manually and install the new openwrt package in the README.

Sounds good!

devkapilbansal: >Let's instead change the data migrations which introduce the template so they do nothing.

What will be the best way to do so? Should I remove that function or just empty the operations list?

pandafy: Replacing the function with `migrations.RunPython.noop` will be good I guess. Don't forget to leave a comment explaining why migration file is like that.

> Should I remove that function or just empty the operations list?

Won't Django raise error on this?

devkapilbansal: Wasn't sure about that and replaced the function with `migrations.RunPython.noop` as suggested

pandafy (commented):
@devkapilbansal the tests are failing because of lack of time resolution. See below

```
./tests/manage.py test openwisp_monitoring.device.tests.test_api.TestDeviceApi.test_200_multiple_measurements
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
wlan0 2021-08-03 14:27:40
wlan1 2021-08-03 14:27:40
wlan0 2021-08-03 14:27:40
wlan1 2021-08-03 14:27:40
wlan0 2021-08-03 14:27:41
wlan1 2021-08-03 14:27:41
wlan0 2021-08-03 14:27:41
wlan1 2021-08-03 14:27:41
F
Summary of slow tests (>0.3s)


Total slow tests detected: 0

======================================================================
FAIL: test_200_multiple_measurements (openwisp_monitoring.device.tests.test_api.TestDeviceApi)
----------------------------------------------------------------------
  File "/home/pandafy/openwisp/openwisp-monitoring/openwisp_monitoring/device/tests/test_api.py", line 138, in test_200_multiple_measurements
    self.assertEqual(len(points), 4)
AssertionError: 2 != 4

----------------------------------------------------------------------
Ran 1 test in 0.723s

FAILED (failures=1)
Destroying test database for alias 'default'...
```

I believe the data points are getting overwritten.

devkapilbansal:
Hi @pandafy openwrt doesn't support time resolution in seconds. One possible solution however can be to install additional dependencies that I don't think should be good. :thinking:

pandafy:
@devkapilbansal I don't think this problem will occur with the monitoring module on the device since the daemon collects data at different intervals of time.

devkapilbansal:
> @devkapilbansal I don't think this problem will occur with the monitoring module on the device since the daemon collects data at different intervals of time.

Actually the problem was to maintain the precision with datetime. If I don't send the time in nanoseconds precision from device then it will not be parsed on this end. 

The best solution I can think of is:
https://github.com/openwisp/openwrt-openwisp-monitoring/pull/34/files#r682056808

README.rst
```diff
@@ -1320,21 +1320,18 @@ Example usage:
     cd tests/
     ./manage.py run_checks
 
-Monitoring scripts
+Monitoring Package
```
nemesisdesign, line 1323: I believe this information should come much earlier in the README, we should probably move it up (eg: what about moving it before https://github.com/openwisp/openwisp-monitoring/tree/issues/298-receive-data-when-device-was-offline#device-health-status?

And what if we move https://github.com/openwisp/openwisp-monitoring#setup-integrate-in-an-existing-django-project) just after the development install instructions?

openwisp_monitoring/device/migrations/__init__.py
```diff
@@ -1,46 +0,0 @@
...
-    }
-)
-TEMPLATE_POST_RELOAD_HOOK_01 = OrderedDict(
-    {
-        "path": "/etc/openwisp/post-reload-hook",
-        "mode": "0744",
-        "contents": "#!/bin/sh\ntouch /etc/crontabs/root\n/etc/init.d/cron start\n/usr/sbin/update-openwisp-packages\n/usr/sbin/openwisp-monitoring\n",  # noqa
-    }
-)
```
nemesisdesign, line 46: as said, I would not remove this code but add a new migration which introduces the following changes:

- ensure the name of the files do not collide with the openwrt openwisp monitoring package
- change the code of the monitoring template so that if it detects the openwrt monitoring package agent it stops and doesn't collect/send data
- remove the automatic install of openwisp-config since it's painful to maintain and can cause issues

nemesisdesign (changes requested):
I took some time to think about how we can transition to the new agent in a backward compatible way.

**Important points:**

1. we cannot remove the monitoring template indiscriminately because we risk turning off monitoring to users which will upgrade openwisp-monitoring without having started to use the new openwrt package yet
2. we should instead create a new data migration which edits the monitoring template in this PR, so that the name of the files do not collide with the openwrt package (because I believe that with the latest renamings we did now the names do collide and hence the template here could overwrite those files) AND if it detects that the new openwrt openwisp-monitoring package is available on the filesystem it shall not collect/send data, so that if it happens that both monitoring package and monitoring templates are present, only monitoring package is used
3. we should publish the new openwrt package on downloads.openwisp.io (this is ongoing thanks to https://github.com/openwisp/openwisp-config/pull/140)
4. we should point to the package here in the README (let's improve what we have here in this PR)

devkapilbansal:
Updated the PR with current master

nemesisdesign (commented):
Here's what I think we should do here:

- ensure we don't mess with updating openwisp config anymore
- rename the scripts installed by the template so they begin with `legacy-`
- the script legacy-openwisp-monitoring should check whether `openwisp-monitoring` exists, and in that case it shall just exit without doing anything
- then we have to make sure that on new installations, the monitoring template is not installed at all, hence we have to make the migrations smart so that they change the monitoring template only if installed, but the old migrations that install the monitoring template in the first place shall not execute anymore
- then we have to document this clearly: users should be warned that if they are using custom monitoring templates, they must disable the monitoring template from their devices before installing the openwrt-openwisp-monitoring package, **we must also add a reference to this important point in the change log** 

This should be it @devkapilbansal @pandafy.

README.rst
```diff
@@ -1432,18 +1341,47 @@ Example usage:
 Monitoring scripts
 ------------------
 
-The monitoring scripts which are automatically installed by a `migration file of device-monitoring app <https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/migrations/0002_create_template.py>`_
-are required to make the `checks <https://github.com/openwisp/openwisp-monitoring#available-checks>`_ and
-`metrics <#openwisp_monitoring_metrics>`_ work.
+Monitoring scripts are now deprecated in favour of `monitoring packages <https://github.com/openwisp/openwrt-openwisp-monitoring#openwrt-openwisp-monitoring>`_.
+Follow the migration guide in `Migrating from monitoring scripts to monitoring packages <#migrating-from-monitoring-scripts-to-monitoring-packages>`_
+section of this documentation.
+
```
pandafy, line 1347: I have kept this section in the documentation for returning users. I think this has better DX than completely removing this section.

README.rst
```diff
@@ -1432,18 +1341,47 @@ Example usage:
...
-``rpcd-mod-iwinfo``.
+After the proper configuration of the `openwisp-monitoring package <https://github.com/openwisp/openwrt-openwisp-monitoring#openwrt-openwisp-monitoring>`_
+on your device, you can remove the monitoring template from your devices.
 
-**WARNING**: Please create a new template if you wish to implement customizations. If you modify the
-default template to create your custom template then your code can get overwritten post an update.
+We suggest removing the monitoring template from the devices one at a time instead
+of deleting the template. This ensures the correctness of
+*openwisp monitoring package* configuration and you'll not miss out on
+any monitoring data.
```
pandafy, line 1379: I can remove this if needed. I added this only as a precaution to users.

openwisp_monitoring/device/migrations/0002_create_template.py
```diff
@@ -63,5 +11,10 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(migrate_data, reverse_code=migrations.RunPython.noop)
+        # NOTE: Usage of "Monitoring Template" is deprecated in favour of
+        # "openwisp-openwisp-monitoring" package. See
+        # https://github.com/openwisp/openwrt-openwisp-monitoring#openwrt-openwisp-monitoring
+        migrations.RunPython(
+            migrations.RunPython.noop, reverse_code=migrations.RunPython.noop
+        )
```
pandafy, line 19: I have removed the code that was responsible for creating the default monitoring template. 
On new systems, no monitoring template will be created because of this. 

I also thought of commenting the code that creates the template for reference. But, we can always use git to check the older version of the code.

openwisp_monitoring/device/migrations/__init__.py
pandafy: @nemesisdesign I have updated the `/usr/sbin/update-openwisp-packages"` script and only removed the part that installed openwisp-config.

openwisp_monitoring/device/migrations/0002_create_template.py
devkapilbansal: 🤝

README.rst
```diff
@@ -103,24 +103,161 @@ Available Features
...
+    docker-compose build
+
+Run the docker container:
+
+.. code-block:: shell
+
+    docker-compose up
 
 Setup (integrate in an existing Django project)
 -----------------------------------------------
```
nemesisdesign, line 263: I think we should move this section under the installation instructions as we have in OpenWISP Controller

README.rst
```diff
@@ -103,24 +103,161 @@ Available Features
 
 ------------
 
-Install Dependencies
---------------------
+Installation instructions
+-------------------------
+
```
nemesisdesign, line 108: I think we need a "Deploy it in production" section like the one we have in OpenWISP Controller: https://github.com/openwisp/openwisp-controller#deploy-it-in-production.

README.rst
```diff
@@ -223,11 +359,97 @@ Configure celery (you may use a different broker if you want):
...
+  from OpenWISP to the device.
+- `Executing commands remotely on the device <https://github.com/openwisp/openwisp-controller#sending-commands-to-devices>`_.
+- `Running checks on the device <https://github.com/openwisp/openwisp-monitoring#available-checks>`_.
+- `Performing firmware upgrades on the device <https://github.com/openwisp/openwisp-firmware-upgrader#perform-a-firmware-upgrade-to-a-specific-device>`_
+
+It is **required** to set up the *management interface* for functioning
+of above-listed features.
+
+You can only get away from setting up a management interface
+if OpenWISP is deployed in a Layer 2 network, i.e. OpenWISP can reach the devices
```
nemesisdesign, line 413: in the same Layer 2 network, any computer has both L2 and L3 networks working in its stack, saying that  something is deployed in an L2 network doesn't have much sense.

README.rst
```diff
@@ -223,11 +359,97 @@ Configure celery (you may use a different broker if you want):
...
+
+Creating checks for a device
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, *openwisp-monitoring* creates `active checks <#available-checks>`_
+for all devices. These checks are created and executed asynchronously.
+
+If you don't see checks created immediately after a device registers, don't
+be alarmed. It may take some time for the asynchronous worker(s) to create
+checks for the device. Unless you have disabled the automatic creation of a check.
```
nemesisdesign, line 451: I think last sentence is not entirely correct, it shouldn't take long, a few seconds max, the user may be misled by this and waste time if there's any issue.

README.rst
```diff
@@ -223,11 +359,97 @@ Configure celery (you may use a different broker if you want):
...
+
+You can only get away from setting up a management interface
+if OpenWISP is deployed in a Layer 2 network, i.e. OpenWISP can reach the devices
+on their **Last IP**. In this scenario, set the `"OPENWISP_MONITORING_MANAGEMENT_IP_ONLY" <#openwisp-monitoring-management-ip-only>`_
+setting in OpenWISP to ``False``.
+
+Install monitoring packages on the device
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Follow the `instructions from openwrt-openwisp-monitoring documentation <https://github.com/openwisp/openwrt-openwisp-monitoring/tree/master#install-pre-compiled-packages>`_
```
nemesisdesign, line 420: please try to break the links to avoid having lines that are too long, eg:

```
`instructions ....
<link>`_
```

This helps with the maintenance of the document.

nemesisdesign (commented):
I was leaving feedback to the README here but I will instead take over and do some changes directly, I think it will be faster.

README.rst
```diff
@@ -223,11 +372,173 @@ Configure celery (you may use a different broker if you want):
...
+    EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
+
+If you decide to use Redis (as shown in these examples),
+install the following python packages.
+
+.. code-block:: shell
 
     pip install redis django-redis
 
+Quickstart Guide
```
pandafy, line 385: This section looks really good right now. 

Maybe, we can also copy this section to the openwisp2-docs in the future.

openwisp_monitoring/device/migrations/__init__.py
```diff
@@ -23,24 +37,52 @@
...
+        "mode": "0744",
+        "contents": "#!/bin/sh\nreboot=false\n\n# updates opkg lists only if necessary\nopkg_update(){\n    (\n        test -d /tmp/opkg-lists/ && \\\n        test -f /tmp/opkg-lists/openwrt_base && \\\n        test -f /tmp/opkg-lists/openwrt_packages && \\\n        test -f /tmp/opkg-lists/openwrt_core\n    ) || opkg update;\n}\n\n# installs libubus-lua if necessary\nlibubus_lua_installed=$(opkg list-installed | grep libubus-lua -c)\nif [ \"$libubus_lua_installed\" == \"0\" ]; then\n    opkg_update\n    opkg install libubus-lua\nfi\n\n# installs lua-cjson if necessary\nluacjson_installed=$(opkg list-installed | grep lua-cjson -c)\nif [ \"$luacjson_installed\" == \"0\" ]; then\n    opkg_update\n    opkg install lua-cjson\nfi\n\n# installs rpcd-mod-iwinfo if necessary\nrpcd_mod_iwinfo_installed=$(opkg list-installed | grep rpcd-mod-iwinfo -c)\nif [ \"$rpcd_mod_iwinfo_installed\" == \"0\" ]; then\n    opkg_update\n    opkg install rpcd-mod-iwinfo\n    reboot=true \nfi\n\n# reboots if rpcd-mod-iwinfo has been installed\nif [ \"$reboot\" == \"true\" ]; then\n    sleep 5\n    reboot && exit\nfi\n",  # noqa
+    }
+)
```
nemesisdesign, line 74: this must be removed: https://github.com/openwisp/openwisp-monitoring/issues/306.

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Here's a problem I found during testing:

```
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/base.py", line 89, in wrapped
    res = handle_func(*args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/core/management/commands/migrate.py", line 244, in handle
    post_migrate_state = executor.migrate(
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/db/migrations/executor.py", line 117, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/db/migrations/executor.py", line 227, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/db/migrations/migration.py", line 126, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/django/db/migrations/operations/special.py", line 190, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/opt/openwisp2/env/lib/python3.8/site-packages/openwisp_monitoring/device/migrations/0003_update_template.py", line 25, in migrate_data
    template.config['files'][5] = TEMPLATE_POST_RELOAD_HOOK_02
IndexError: list assignment index out of range
```

The code is making wrong assumptions.
I think we should simply overwrite the template altogether, we cannot know how users modified this template.
We must advise users to back up the monitoring template if they did any modifications to it before upgrading (we can do so in the change log).

nemesisdesign (approved):

PR #315, opened 2021-08-09
[fix] Use REDIS_HOST in CHANNEL_LAYERS in test app

pniaps:
CHANNEL_LAYERS was not using REDIS_HOST env var


nemesisdesign (approved):
Thanks

PR #316, opened 2021-08-10
[feature] Allow specify time of received timeseries data #298

### LINKED ISSUE #298: [feature] Allow specify time of received timeseries data
- [ ] Patching device metric API to collect device data when server was down
- [ ] Allow DateTime to be passed explicitly
- [ ] Add documentation about it in README

devkapilbansal:
Closes #298

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1226,13 +1226,19 @@ Collect device metrics and status
 
     POST /v1/monitoring/device/{pk}/?key={key}&time={time}
 
+If data is latest then an additional parameter current can also be passed. For e.g.:
+
+.. code-block:: text
+
+    POST /v1/monitoring/device/{pk}/?key={key}&time={time}&current=true
+
 The format used for Device Status is inspired by `NetJSON DeviceMonitoring <http://netjson.org/docs/what.html#devicemonitoring>`_.
 
 **Note**: Device data will be saved with time passed in timeseries database.
 It should be in `%d-%m-%Y_%H:%M:%S.%f` format, else Bad Response(400) will be returned. 
 If request is made without passing time, then server time will be used.
 
-``time`` parameter is used to [send the data that is collected](https://github.com/openwisp/openwrt-openwisp-monitoring#collecting-vs-sending) when device was not reachable.
+``time`` parameter is used to [send the data collected](https://github.com/openwisp/openwrt-openwisp-monitoring#collecting-vs-sending) from device. 
 
 Signals
 -------
@@ -1248,8 +1254,6 @@ Signals
 - ``request``: the HTTP request object
 - ``time``: time with which metrics will be saved. If none, then server time will be used
 - ``is_latest``: whether the data is latest or not
-
-If none, then server time will be used.
 
 This signal is emitted when device metrics are received to the ``DeviceMetric``
 view (only when using HTTP POST).
--- a/README.rst
+++ b/README.rst
@@ -1254,7 +1254,7 @@ Signals
 - ``instance``: instance of ``Device`` whose metrics have been received
 - ``request``: the HTTP request object
 - ``time``: time with which metrics will be saved. If none, then server time will be used
-- ``is_latest``: whether the data is latest or not
+- ``current``: whether the data has just been collected or was collected previously and sent now due to network connectivity issues
 
 This signal is emitted when device metrics are received to the ``DeviceMetric``
 view (only when using HTTP POST).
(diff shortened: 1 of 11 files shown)


openwisp_monitoring/device/api/views.py
```diff
@@ -165,6 +167,12 @@ def post(self, request, pk):
         except ValidationError as e:
             logger.info(e.message)
             return Response(e.message, status=status.HTTP_400_BAD_REQUEST)
+        time_obj = request.query_params.get('time')
+        if time_obj:
+            time = datetime.strptime(time_obj, '%d-%m-%Y_%H:%M:%S.%f')
```
nemesisdesign, line 172: what if a bogus time is supplied?
Please make sure 400 bad request is returned with a proper error message + a test.

nemesisdesign (changes requested):
I think the README should be updated: https://github.com/openwisp/openwisp-monitoring#collect-device-metrics-and-status, we should mention the time parameter and how we use it (to collect data stored previously, we should link the relevant section of the agent README too).

I don't see any new test being added, but I think at least new test should be added with an assertion which checks the time parameter is respected (data in timeseries DB is inserted with the specified time), don't you think?

openwisp_monitoring/device/api/views.py
```diff
@@ -165,6 +167,12 @@ def post(self, request, pk):
...
+            self.time = time.replace(tzinfo=UTC)
+        else:
+            self.time = None
```
nemesisdesign, line 175: I think we should pass this around as argument instead of using an attribute, it's an argument specified by the sender, it's really weird to set it as an attribute.

nemesisdesign (changes requested):
There's no info in the README about this, please add a mention of the time parameter here: https://github.com/openwisp/openwisp-monitoring#retrieve-device-charts-and-device-status-data.

I think we have to update some signals (and their relevant documentation) to add the passed time, because anyone who will want to use these signals will want to know if the metric data is old or recent:

- https://github.com/openwisp/openwisp-monitoring/blob/issues/298-receive-data-when-device-was-offline/openwisp_monitoring/device/api/views.py#L183
- https://github.com/openwisp/openwisp-monitoring#pre-metric-write
- https://github.com/openwisp/openwisp-monitoring#post-metric-write

Question: does the agent send latest (current) data in a way that clearly distinguishes from old data right? So that in the code of this module we know what is old and what is new right?

I ask this because one of the problems I'm pretty sure that we'll run into is that whenever old data is collected, a bunch of alerts will be generated all in one go, at that point we'll have to figure out how to avoid this if possible.

Please make sure the same approach is used also in signals, so that also signal receivers can now when the data is current and when it's old data.

Regarding the WiFi clients, here's a possible solution which fixes the tests: https://github.com/openwisp/openwisp-monitoring/issues/317#issuecomment-902346162. Please try it out.

devkapilbansal:
> Question: does the agent send latest (current) data in a way that clearly distinguishes from old data right? So that in the code of this module we know what is old and what is new right?
> 
Yes, an additional request query parameter `current=True` is received for latest data. This can be used to distinguish if data is old or not.
> 
> Regarding the WiFi clients, here's a possible solution which fixes the tests: [#317 (comment)](https://github.com/openwisp/openwisp-monitoring/issues/317#issuecomment-902346162). Please try it out.
:ok:

devkapilbansal:
> I don't see any new test being added, but I think at least new test should be added with an assertion which checks the time parameter is respected (data in timeseries DB is inserted with the specified time), don't you think?

Tests are added to ensure time passed in timeseries database is the time used to create the metric (by testing the signals). I don't think we need any additional test case. :thinking:

openwisp_monitoring/device/api/views.py
```diff
@@ -165,14 +168,26 @@ def post(self, request, pk):
...
             logger.info(e.message)
             return Response(e.message, status=status.HTTP_400_BAD_REQUEST)
+        time_obj = request.query_params.get('time', now())
+        is_latest = request.query_params.get('current', False)
+        try:
+            time = datetime.strptime(time_obj, '%d-%m-%Y_%H:%M:%S.%f').replace(
+                tzinfo=UTC
+            )
+        except ValueError:
+            return Response({'detail': 'Incorrect time format'}, status=400)
```
nemesisdesign, line 178: please flag 'Incorrect time format' as translatable

README.rst
```diff
@@ -1240,6 +1252,8 @@ Signals
 
 - ``instance``: instance of ``Device`` whose metrics have been received
 - ``request``: the HTTP request object
+- ``time``: time with which metrics will be saved. If none, then server time will be used
+- ``is_latest``: whether the data is latest or not
```
nemesisdesign, line 1257: The fact that use use `is_latest` in one place and `current` in the other does not convince me 100%.

Did you think about this?

Also, please let's clarify what this line means, a user may not understand what "whether the data is latest or not", I think we should write something like "whether the data has just been collected (latest data) or was collected previously and is being sent now due to network connectivity issues on the sender side".

nemesisdesign (changes requested):

nemesisdesign (approved):
Great work, thanks! :+1:

Issue #323, opened 2021-10-12
[ci] Add python 3.8 and python 3.9, remove python 3.6


codesankalp:
Add python 3.8 and python 3.9, remove python 3.6 from GitHub actions testing.

Fix any bug that may appear.

harsh-9in:
Hi @codesankalp, I am new to the project and tried to implement the solution at pr #324 please look into it and if any changes are needed, let me know.

PR #324, opened 2021-10-15
[change] Changed python version from 3.6 to 3.8 and 3.9 #323

harsh7in:
Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
+++ b/openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
@@ -4,7 +4,10 @@ from django.utils import timezone
 from freezegun import freeze_time
 from swapper import load_model
 
-from ...device.tests import DeviceMonitoringTestCase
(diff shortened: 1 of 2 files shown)


pandafy:
@nemesisdesign you will need to update list of required checks in the repository.

atb00ker:
@pandafy Is there is something else that needs to be updated for the 3.6 test to go away? 
This problem is there in multiple such PRs! Are you aware of it?

devkapilbansal:
> @pandafy Is there is something else that needs to be updated for the 3.6 test to go away? This problem is there in multiple such PRs! Are you aware of it?

Maybe the branch protection rules that @nemesisdesign commented about in one of the PR

atb00ker:
> Maybe the branch protection rules that @nemesisdesign commented about in one of the PR

Oh, I thought that was only true for travis. 
This is not ideal frankly if we have to do everytime, need to come up with a more permanent solution!

PR #335, opened 2022-01-05
[change] Made metric health status independent of AlertSetting tolerance #293

### LINKED ISSUE #293: [bug/change] Decouple status change from alert
I describe the problem we have currently with an example.

- we set a tolerance of 5 minutes on the a specific check, say the ping check
- a device is online, then for any reason it's not reachable by the ping check
- the system will wait 5 minutes before sending the alert and changing its status

However, it would be a lot better if we could change the status immediately, but just ensure it does not trigger the alert.

**This is one of those issues which sounds trivial but it's not, so beware.**

pandafy:
Closes #293

Checks:


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,48 +156,97 @@ def _get_time(self, time):
...
+    def _set_is_tolerance_healthy(
+        self, alert_settings, value, time, retention_policy, send_alert
+    ):
+        """
+        Sets the value of "is_tolerance_healthy" if "value"
+        crosses threshold for tolerance defined in alert_settings.
+        It also sends the notification if required.
+        Returns "None" if value of "is_tolerance_healthy" is unchanged.
+        Returns "True" if it is the first metric write within threshold.
+        Returns "False" in other cases.
```
pandafy, line 186: Using three different values for depicting different things might seems complex. Any suggestion on how it can be improved?

openwisp_monitoring/device/base/models.py
```diff
@@ -273,7 +273,7 @@ def related_metrics(self):
     @receiver(threshold_crossed, dispatch_uid='threshold_crossed_receiver')
     def threshold_crossed(sender, metric, alert_settings, target, first_time, **kwargs):
         """
-        Changes the health status of a device when a alert_settings is crossed.
+        Changes the health status of a device when a metric threshold is crossed.
```
nemesisdesign, line 276: the threshold is not set at metric level though, but at alert settings level. What about: "when a threshold defined in the alert settings related to the metric is crossed"? Or something along this line.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # TODO: put back the standard pypi release when
 # next point version of openwisp-controller is released
-openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/master
+openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/controller-views-urls
```
nemesisdesign, line 3: this should be merged

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -52,6 +52,8 @@ class AbstractMetric(TimeStampedEditableModel):
     content_object = GenericForeignKey('content_type', 'object_id')
     # NULL means the health has yet to be assessed
     is_healthy = models.BooleanField(default=None, null=True, blank=True, db_index=True)
+    # Like "is_healthy", but respects tolerance of alert settings
+    is_tolerance_healthy = models.BooleanField(default=None, null=True, blank=True)
```
nemesisdesign, line 56: What about `is_healthy_tolerant`?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,48 +156,97 @@ def _get_time(self, time):
...
+        elif not crossed and self.is_healthy in [False, None]:
+            self.is_healthy = True
+        return True
+
+    def _set_is_tolerance_healthy(
+        self, alert_settings, value, time, retention_policy, send_alert
+    ):
+        """
+        Sets the value of "is_tolerance_healthy" if "value"
+        crosses threshold for tolerance defined in alert_settings.
```
nemesisdesign, line 182: I would try to make this dosctring a bit more readable:

Sets the value of "is_tolerance_healthy" if "value"
crosses the threshold for more than the amount of seconds defined in the alert_settings tolerance field.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,48 +156,97 @@ def _get_time(self, time):
...
             notification_type = f'{self.configuration}_recovery'
         # First metric write within threshold
-        elif not crossed and self.is_healthy is None:
-            self.is_healthy = True
+        elif not crossed and self.is_tolerance_healthy is None:
+            self.is_tolerance_healthy = True
             first_time = True
-        self.save()
+
+        # Don't raise alert if it is the first metric within threshold.
```
nemesisdesign, line 211: I would write something like:

```
If we got to this point, it means we have to send an alert, because the metric
has been crossed for more than the tolerated amount of time.
There's one exception though: if the device is new, its status will be unknown
and the metric will become healthy for the first time, in this case we do not need
to send an alert.
```

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -154,48 +156,97 @@ def _get_time(self, time):
             return parse_date(time)
         return time
 
-    def check_threshold(self, value, time=None, retention_policy=None, send_alert=True):
+    def _set_is_healthy(self, alert_settings, value):
         """
-        checks if the threshold is crossed
-        and notifies users accordingly
+        Sets the value of "is_healthy" field if "value"
+        crosses threshold defined in "alert_settings".
```
nemesisdesign, line 162: We could add that this method does not take into account the alert settings tolerance, which is done instead by the other method (pointing out which one).

openwisp_monitoring/monitoring/base/models.py
nemesisdesign: We could also write that this method is very similar to `_set_is_healthy` but it takes into account the alert settings tolerance so it's slightly different and more complex.

nemesisdesign (changes requested):
Looks good! See my comments below.

nemesisdesign (changes requested):
I found an issue while testing.
After deploying it to a test instance I received one alert for each device that was present in the system even though these alerts were not necessary.

nemesisdesign (approved):
This is good, I will merge until the build gets fixed :+1:.

PR #336, opened 2022-01-06
[fix] Restoring a deleted device restores related objects #334

### LINKED ISSUE #334: [bug] Recovering deleted device does not recover related objects
Device status, device chart, checks and alert settings are not recovered after restoring a deleted device. These models should be tracked by django-reversion.

**Blockers**
- [x]  https://github.com/openwisp/openwisp-controller/pull/581

pandafy:
Restorig a deleted device restores related DeviceMonitorig,
Metric and AlertSettings objects.

Related Check objects are not restored.

Closes #334

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/admin.py
+++ b/openwisp_monitoring/monitoring/admin.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
 from django.forms import ModelForm
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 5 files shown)


nemesisdesign (changes requested):
Can you please add a test which fails without the fix?

nemesisdesign (approved):
It works, thanks! :+1:

PR #337, opened 2022-01-10
[deps] Added support for Django 3.2 and Django 4.0 #322  

### LINKED ISSUE #322: [deps] Add support for django 3.2 and django 4.0 alpha
Let's make sure this module can run with Django 3.2 and Django 4.0 alpha.

pandafy:
- Bumped django-nested-admin~=3.4.0 - Pinned test dependencies

Closes #322 

Checks:


README.rst
```diff
@@ -1853,8 +1853,8 @@ Step 2: remove the following line from your root ``urls.py`` file:
 
 .. code-block:: python
 
-    url(
-        r'^api/v1/monitoring/device/(?P<pk>[^/]+)/$',
+    path(
```
nemesisdesign, line 1856: wouldn't this need `re_path`?

openwisp_monitoring/check/base/models.py
```diff
@@ -65,7 +65,7 @@ def check_class(self):
         """
         returns check class
         """
-        return import_string(self.check)
+        return import_string(self.check_type)
```
nemesisdesign, line 68: because of this, I think we should call this `check_class` instead of `check_type`, because the content of this field is a python path leading to a python class, it makes the code more explicit and hence more readable.

nemesisdesign (changes requested):

openwisp_monitoring/check/base/models.py
pandafy: There is already a property of Check model name check_class. 

https://github.com/openwisp/openwisp-monitoring/blob/5b8169707783e0078e6541044635a48826a918f9/openwisp_monitoring/check/base/models.py#L63-L68

nemesisdesign (approved):

PR #339, opened 2022-01-13
[tests] Added test for triggering device checks on recovery #338

### LINKED ISSUE #338: [bug] Task trigger_device_checks triggering a lot of unnecessary network checks
The task `trigger_device_checks` is called when checksum and monitoring data is received, which happens all the time.
Then this code is executed in the background:

https://github.com/openwisp/openwisp-monitoring/blob/bdcfff2b17468278ec161b4b31b64ea67a6869fb/openwisp_monitoring/device/tasks.py#L12-L33

The point of this feature is to quickly find out if a device has come back online after it wasn't.
But the code doesn't take into account that if a device is already ok it should not repeat the checks.

So if there's 1000 devices online, not only the monitoring worker will do 1000 checks every 5 minutes, but it will do a lot more because every 5 minutes there's at least 1 metric sending and 2 checksum requests, so every 5 minutes the system will do 4000 checks.

If my analysis is correct, this is a problem.

So first of all we should check whether the analysis is correct, then I think we can join the device data with DeviceMonitoring (so we keep doing one query to retrieve the device, but will fetch the health status as well), then check whether the health status is ok and in that case I think we can just return.

pandafy:
Closes #338

### DIFF
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -105,7 +105,8 @@ class DeviceMonitoringConfig(AppConfig):
     def trigger_device_recovery_checks(cls, instance, **kwargs):
         from .tasks import trigger_device_checks
 
-        if cache.get(get_device_cache_key(device=instance)):
+        # Cache is managed by "manage_device_recovery_cache_key".
+        if cache.get(get_device_cache_key(device=instance), False):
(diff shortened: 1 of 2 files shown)


nemesisdesign (approved):
Thanks! :+1:

PR #342, opened 2022-01-27
[refactoring] Migrate from Popen() to run() #300

### LINKED ISSUE #300: [refactoring] Migrate from Popen() to run()
From the [Python docs](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module)
> The recommended approach to invoking subprocesses is to use the `run()` function for all use cases it can handle. For more advanced use cases, the underlying Popen interface can be used directly.
> 
> The run() function was added in Python 3.5; if you need to retain compatibility with older versions, see the Older high-level API section.

Is this an "advanced use case"?  Should we consider migrating to `run()`?

https://github.com/openwisp/openwisp-monitoring/blob/1b78b37b9e511db9029007111d1ddc2b97ba1a2b/openwisp_monitoring/check/classes/ping.py#L140

Aryamanz29:
Fixes #300

Checks:

### DIFF
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -148,8 +148,8 @@ class Ping(BaseCheck):
         """
         Executes command (easier to mock)
         """
-        p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


nemesisdesign (approved):
Thanks @Aryamanz29.

PR #346, opened 2022-02-18
[feature] Make requests to timeseries DB resilient

### LINKED ISSUE #345: [feature] Make requests to timeseries DB resilient
On a busy system, requests to influxdb can fail for different reasons.

Now, for the write operation, we already have a resilience mechanism in place, but for all other operations, we don't.

I think we could write a solution which implements resiliency for any the other call which talks to the timeseries DB in the influxdb backend.

The resiliency in this case can be something simple, a configurable retry mechanism, eg: a loop like for `attempt in range(0, MAX_ATTEMPTS)`. There should be a sleep between each retry but only if the third attempt fails, the sleep can be a fixed amount of seconds and should be configurable, defaulting to 2 seconds. The number of attempts can default to 6.

The exceptions catched shall be defined as a class attribute, we should start with the ones which we are collecting from the error logs.

Every time an attempt fails we should log some useful info with the `info` level.

If the max number of attempts is reached the code shall fail as it's failing now.

I would also take advantage to rename the `get_db` property (which was a method and was converted to a property but not renamed) to something different, maybe just `db` or `_db`.

codesankalp:
Closes #345 

Checks:


README.rst
```diff
@@ -949,6 +949,28 @@ domain, you can use this option to change the base of the url, this will
 enable you to point all the API urls to your openwisp-monitoring API server's
 domain, example: ``https://mymonitoring.myapp.com``.
 
+``OPENWISP_MONITORING_INFLUXDB_MAX_RETRIES``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 953: let's rename these settings to be generic, so that can be reused with any other future timeseries backend we may have, eg: ``OPENWISP_MONITORING_TIMESERIES_MAX_RETRIES``

README.rst
```diff
@@ -949,6 +949,28 @@ domain, you can use this option to change the base of the url, this will
...
++--------------+-------------+
+| **type**:    |   ``int``   |
++--------------+-------------+
+| **default**: |    ``6``    |
++--------------+-------------+
+
+This settings allow you to configure the max retries count on failure in InfluxDB operations.
+
+``OPENWISP_MONITORING_INFLUXDB_RETRY_DELAY``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 964: ``OPENWISP_MONITORING_TIMESERIES_RETRY_DELAY``

README.rst
```diff
@@ -949,6 +949,28 @@ domain, you can use this option to change the base of the url, this will
...
+``OPENWISP_MONITORING_INFLUXDB_MAX_RETRIES``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    |   ``int``   |
++--------------+-------------+
+| **default**: |    ``6``    |
++--------------+-------------+
+
+This settings allow you to configure the max retries count on failure in InfluxDB operations.
```
nemesisdesign, line 961: Let's not mention InfluxDB but use the generic "Timeseries Database".

Please also make sure these new settings are listed after `OPENWISP_MONITORING_WRITE_RETRY_OPTIONS` and make sure to explain the difference between these new settings and the old one.

README.rst
```diff
@@ -949,6 +949,28 @@ domain, you can use this option to change the base of the url, this will
...
+``OPENWISP_MONITORING_INFLUXDB_RETRY_DELAY``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    |   ``int``   |
++--------------+-------------+
+| **default**: |    ``2``    |
++--------------+-------------+
+
+This settings allow you to configure the retry delay time (in seconds) after 3 failed attempt in InfluxDB.
```
nemesisdesign, line 972: let's not mention InfluxDB but use the generic "Timeseries Database".

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Seems you forgot this point:

Please also make sure these new settings are listed after OPENWISP_MONITORING_WRITE_RETRY_OPTIONS and make sure to explain the difference between these new settings and the old one.

nemesisdesign (changes requested):
Thank you Sankalp, it looks very good and it seems to be working well.

After reviewing the docs again, I think we should do as follows for simplicity and consistency with the other setting which allows to control the retry mechanism for write operations.

I think it's also easier to explain, look at my docs changes regarding `OPENWISP_MONITORING_TIMESERIES_RETRY_OPTIONS`, can you please update the code to reflect those docs changes?

Thanks.

nemesisdesign (approved):
Thanks :+1:

Issue #274, opened 2021-03-10
[CI] Migrate CI build to influxdb 2.0


nemesisdesign:
I had to pin the inflxudb docker image used for testing to version 1.8.4 because the new version 2.0 breaks the tests:
https://github.com/openwisp/openwisp-monitoring/pull/273

However, this is not great: we want to test with the latest influxdb version and therefore it would be great to upgrade our code so that the test environment is compatible with the last influxdb version. It should be just a matter of setting the authentication credentials correctly.

devkapilbansal:
Hi @nemesisdesign the current library used here supports `InfluxDB 1.x` only. 
Another library ([influxdb-client](https://github.com/influxdata/influxdb-client-python)) only support `InfluxDB 1.8+`.

Would it be good to drop support for InfluxDB 1.7 or earlier ? :thinking:

nemesisdesign:
> Hi @nemesisdesign the current library used here supports `InfluxDB 1.x` only.
> Another library ([influxdb-client](https://github.com/influxdata/influxdb-client-python)) only support `InfluxDB 1.8+`.
> 
> Would it be good to drop support for InfluxDB 1.7 or earlier ?

@devkapilbansal I guess we have no other viable option, influxdb 1.8 works well so if we can support from 1.8 onwards that's good enough, what worries me is the change of library, do you know if the syntax is similar?

devkapilbansal:
Yes, as both libraries are maintained by `InfluxData`, syntax is quite similar

AbhigyaShridhar:
@nemesisdesign I cannot reproduce this issue. I changed the version of influxdb image to 2.0-alpine in both docker-compose.yml and workflows/ci.yml. All the tests run perfectly.

devkapilbansal:
@AbhigyaShridhar how did you checked it? I just checked and issue is reproducible:
https://github.com/devkapilbansal/openwisp-monitoring/runs/5649166266?check_suite_focus=true#step:10:17

Had it not been reproducible, the version have been updated a long time ago

nemesisdesign:
> @nemesisdesign I cannot reproduce this issue. I changed the version of influxdb image to 2.0-alpine in both docker-compose.yml and workflows/ci.yml. All the tests run perfectly.

[...]

PR #354, opened 2022-04-12
[monitoring] Adding backend for influxdb2.x #274

AbhigyaShridhar:
I realized that this is a big issue and will take some time to be resolved completely so this is the initial code for it

Changes made:

 - added another backend in /home/abhigya/Documents/OpenSource/openwisp/monitoring/openwisp-monitoring/openwisp_monitoring/db/backends
 - made changes to the __init__ file of backends to now deal with two possible backends and their different requirements
 - added initial code for the influxb2.x backend client
 - wrote initial code for a method to 'standardize' query results. This will have to be implemented for all the backends in future

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     volumes:
       - influxdb-data:/var/lib/influxdb
     ports:
-      - "8086:8086"
+      - "8085:8085"
     environment:
       INFLUXDB_DB: openwisp2
       INFLUXDB_USER: openwisp
--- a/openwisp_monitoring/db/backends/__init__.py
+++ b/openwisp_monitoring/db/backends/__init__.py
@@ -9,15 +9,8 @@ logger = logging.getLogger(__name__)
 
 TIMESERIES_DB = getattr(settings, 'TIMESERIES_DATABASE', None)
 if not TIMESERIES_DB:
-    TIMESERIES_DB = {
-        'BACKEND': 'openwisp_monitoring.db.backends.influxdb',
-        'USER': getattr(settings, 'INFLUXDB_USER', 'openwisp'),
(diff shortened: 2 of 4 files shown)


AbhigyaShridhar:
@nemesisdesign @pandafy 

In order to provide the users with an option to choose among multiple backends, the return type of all the db queries must be standardized. In this way, we won't need to make much changes to either views or tests. 

I have attempted to write one such method for influxb2 and I'll keep updating that as I work with more queries. Please let me know your thoughts on this.

Also, as of now I am testing everything I write in a python command line shell. In order to run actual tests and set up a default user, we have to run a docker container setting up the database on a localhost. I am having trouble running the container for influxdb2 (inside the docker-compose file for this project, I ran a container outside the project directory to test my code)

The docker file should look something like [this](https://github.com/AbhigyaShridhar/InfluxDB-Demo/blob/main/docker-compose.yml)

There are two options to run a docker container for influxdb2, in 'setup' mode and in 'upgrade' mode. As per my understanding, when a user chooses influxdb2 as a backend, they have already migrated to the db, we don't need the functionality to allow users to migrate. So, we should run the container in 'setup' mode. Please let me know your thoughts on this as well. (however as of now, when I start influxdb2 container in the project directory, it doesn't work as intended and we need to figure out why)

pushpitkamboj:
This PR felt like the most close implementation for adding influxDB2.0 but was left abandoned. Was there any reason for this?

AbhigyaShridhar:
Hi,
The reason was pure bandwidth constraints.

I picked this up back when I was in college. However, my proposal couldn't be selected and I got busy with other stuff. Now that I have more experience, I would still like to pick this up independently without constraining myself with any open source programe. If the timeseries project is not picked for this edition of GSOC I will be picking this up in the coming months.

@pushpitkamboj Are you planning to work on this?

PR #368, opened 2022-04-27
[!change] Changed structure of the timeseries data on InfluxDB #363

### LINKED ISSUE #363: [change] Change structure of timeseries data on influxdb
Right now, the traffic and wifi clients information is stored in a way that does not allow efficient filtering.

For example, we cannot find out all the traffic consumed by an organization, or all the WiFi clients of a device, or all the WiFi clients of an organization.

The interface traffic and wifi client info is stored in a measurement which is named after the interface name.

I believe we should change this mechanism as follows:

- store traffic info in a measurement called "traffic"
- store wifi client info in a measurement called "wifi_clients"
- add the organization UUID of the device as tag called `organization_id`, this will allow filtering by organization
- add the UUID of the location of the device as a tag called `location_id`, this will allow filtering by location

We should look for a way to migrate the data, I think we can create a data migration which rewrites this data to the new format and when done deletes the old measurements.

pandafy:
- [x] Updated structure of traffic data stored in InfluxDB:
    - The data in stored under `traffic` measurement instead of using the interface name as measurement key. 
    - The values in `traffic` measurements are tagged with `organization_id` of the device and `ifname` (interface name).
- [x] Updated structure for storing WiFi clients in InfluxDB:
    - The values in `wifi_clients` measurements are tagged with `organization_id` of the device and `ifname` (interface name).
- [x]  Added migration that migrates data from old structure to the new structure.

 Closes #363


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -95,11 +117,30 @@ def _get_or_create(cls, **kwargs):
...
+            # organization of a device is changed, etc.)
+            # which results in the creation of new Metric objects.
+            # Hence, only ifname (if present) is used for the
+            # database lookup.
+            if extra_tags.get('ifname'):
+                condition &= models.Q(
+                    extra_tags__contains='"ifname": "{ifname}"'.format(
+                        ifname=extra_tags.get('ifname')
+                    )
+                )
```
nemesisdesign, line 137: This is quite hacky.
It would be better to have 2 fields:

- main_tags: we store things like ifname there
- extra_tags: we store things like org id, location id and floorplan id

We could query data only using main tags, object id and content type, so even if some other detail changes, the charts should not end up empty.
I think we still have to store the other details in extra_tags to allow writing the tags properly.

If a new metric is created when the org or location changes, I don't think it's a big deal, especially if we use only the main_tags to query the data.

nemesisdesign (changes requested):

openwisp_monitoring/monitoring/migrations/0005_migrate_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def forward_migrate_metric(metric_model, configuration, new_key):
+    metric_qs = metric_model.objects.filter(configuration=configuration).exclude(
+        key__in=EXCLUDED_MEASUREMENTS
+    )
+    updated_metrics = []
+    for metric in metric_qs.iterator(chunk_size=CHUNK_SIZE):
+        extra_tags = {
+            'ifname': metric.key,
```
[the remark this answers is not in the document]
pandafy: @nemesisdesign did you check this comment? 
https://github.com/openwisp/openwisp-monitoring/pull/368#discussion_r860538208

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -73,7 +81,7 @@ class Meta:
             'field_name',
             'content_type',
             'object_id',
-            'extra_tags',
+            'main_tags',
```
nemesisdesign, line 84: won't this cause issues if there are multiple metrics with different extra_tags but same main_tags?

CHANGES.rst
```diff
@@ -18,6 +18,17 @@ Backward incompatible changes
...
   create a backup of your template before running migrations. Running
   migrations will make changes to the default *Monitoring Template*.
+- The time-series database schema for storing
+  `interface traffic <https://github.com/openwisp/openwisp-monitoring#traffic>`_
+  and `associated WiFi clients <https://github.com/openwisp/openwisp-monitoring#wifi-clients>`_
+  has been updated. The data for *interface traffic* and *associated WiFi clients*
+  is stored in ``traffic`` and ``wifi_clients`` measurements respectively.
+  The Django migrations will perform the necessary operations in the time-series
+  database aysnchronously. It is recommended that you backup the time-series
+  database before running the migrations.
```
nemesisdesign, line 28: What about the fact that when the migrations are run via ansible-openwisp2, most likely the migration will be executed before supervisor is reloaded, therefore the celery task will not be available yet? What can we do about this?
Can you add a way to call the influxdb migration idempotently so we can add an explicit task in ansible-openwisp2 for this?

nemesisdesign (changes requested):
Looks good, see my comment below.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -50,14 +53,36 @@ class AbstractMetric(TimeStampedEditableModel):
...
+        dump_kwargs={'indent': 4},
+        db_index=True,
+    )
+    extra_tags = JSONField(
+        _('extra tags'),
+        default=dict,
+        blank=True,
+        load_kwargs={'object_pairs_hook': OrderedDict},
+        dump_kwargs={'indent': 4},
+        db_index=True,
```
nemesisdesign, line 70: Are we querying by extra_tags? If not, please remove this.

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #372, opened 2022-05-10
[feature] Added monitoring of WiFi Client and Sessions #360

### LINKED ISSUE #360: [feature] Store WiFi sessions
## `WifiClient`

- create WiFi client model which holds the data of WiFi clients which is not present in the device status
  - include mac address, vendor, HT (boolean), VHT (boolean), WMM (boolean), WDS (boolean), WPS (booelan) info, make sure this info is updated if changed
    mac address shall be the primary key (no need of uuid)
    add also created and modified fields (from openwisp-utils)
    no need to have an organization field because the WiFi client session model below is the one which will be displayed

## `WifiSession`

- create WiFi client session model which holds:
  - relation to Device
  - relation to WiFi client
  - SSID
  - interface name
  - start time
  - stop time (null=True, db_index=True)
  - make sure it has the modified fields (from openwisp-utils) but not the created field (because would be redundant with start_time)
  - shall have uuid

## Populate wifi clients and wifi sessions

- if wifi clients are detected in the device status data, pass the data to a new background task
- the new background task shall save/update the WiFi client info if not already present/updated
  but it should write only if needed (no unneeded writes);
  then it shall also create new WiFi client session entries if needed (again no unneeded writes);
  as last step, it should look whether there's any open WiFi client session (start_time=None)
  for other wifi clients with mac addresses different than the ones currently active,
  if there is any it shall retrieve the wifi client sessions, create a list with the primary keys of the
  sessions retrieved and build an update query
  to flag those sessions as closed, eg: ``WifiSession.objects.filter(pk__in=[pk_list]).update(stop_time=now)``
- add management task to delete WiFi client sessions older than X and then it shall delete any WiFI client
  which doesn't have any related session, with X defaulting to 6 months
- add a new admin class for WiFi sessions:
  - in the list show mac address, vendor, organization, device name with link to its page (as done in openwisp-controller https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/subnet_division/admin.py#L54-L105), SSID, HT, VHT, start time, stop time
  - in the details show the same fields plus the interface name, WMM, WDS, WPS; also make sure the device name links to the device page (as done in openwisp-controller https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/subnet_division/admin.py#L54-L105) and the modified date
  - implement multi-tenancy based on the device relation
  - allow filtering by device name or pk (see DeviceFilter here for reference on how to implement this custom filter https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/subnet_division/filters.py#L29-L44), groups (device group), start and stop time, stop_time=None, organization;
  - allow searching by wifi client mac address, device name, device mac address
- add a new navigation menu in the monitoring group called "WiFi Sessions"
- add a setting which allows to disable this feature, if this feature is disabled, these features shall be disabled:
  - the background tasks which writes the data to the SQL primary store
  - dashboard charts
  - admin section
  - navigation menu item shouldn't be shown

## Mock ups

wifi-client-session-list

wifi-client-session-detail

pandafy:
Closes #360

TODO

- [ ]  Add documentation 

Screenshot from 2022-05-10 22-49-12
Screenshot from 2022-05-10 22-45-38


nemesisdesign (changes requested):
First round of testing, looks very good! :+1: 

Let's add a wifi icon to openwisp-utils for the menu, I think you can move the menu up before metrics and charts (which aren't really useful for every day users).

I think we also need to add permissions to operators and administratrors to view wifi client and session objects.

See also my comments below.

openwisp_monitoring/monitoring/tasks.py
```diff
@@ -50,3 +51,60 @@ def migrate_timeseries_database():
...
+            active_wireless_sessions.append(session_object.pk)
+
+    # Close open WifiSession
+    WifiSession.objects.filter(device_id=device_pk, stop_time=None,).exclude(
+        pk__in=active_wireless_sessions
+    ).update(stop_time=now())
+
+
+@shared_task
+def delete_wifi_clients_and_session(days=6 * 30):
```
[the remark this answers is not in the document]
pandafy: I planned to document it like https://github.com/openwisp/openwisp-notifications#scheduled-deletion-of-notifications

README.rst
```diff
@@ -88,6 +88,9 @@ Available Features
   mobile signal (LTE/UMTS/GSM `signal strength <#mobile-signal-strength>`_,
   `signal quality <#mobile-signal-quality>`_,
   `access technology in use <#mobile-access-technology-in-use>`_)
+* Maintains a record of `WiFi sessions <#monitoring-wifi-sessions>`_ with clients'
+  MAC address and vendor, session start and stop time and connected device
+  along with other information.
```
nemesisdesign, line 93: The other points don't have a final dot so let's maintain consistency.

Let's make clear that both currently open sessions and past sessions can be viewed and filtered.

nemesisdesign (changes requested):
Great progress Gagan!

Let's not forget the WiFi icon in the menu please.

See my comments below.

nemesisdesign (commented):
The open wifi sessions button looks a bit too far, I tried giving it:

```
    position: relative;
    top: -20px;
```

And it's better, I think we should add the possibility of providing a CSS class here.

nemesisdesign (commented):
I found a flaw in my design! If the de device is offline, the wifi clients remain online.
I think we should trigger the celery task when the device status moves into critical to close all the sessions

nemesisdesign (changes requested):

openwisp_monitoring/device/migrations/0004_wificlient_wifisession.py
```diff
@@ -0,0 +1,141 @@
...
+            ],
+            options={
+                'verbose_name': 'WiFi Session',
+                'ordering': ('-start_time',),
+                'abstract': False,
+                'swappable': 'DEVICE_MONITORING_WIFISESSION_MODEL',
+            },
+        ),
+        migrations.RunPython(
+            assign_permissions_to_groups, reverse_code=migrations.RunPython.noop
```
nemesisdesign, line 139: please run this migration in a separate data migration step (file), do you remember on postgres we get errors when we try to do changes to the data in the same transaction in which the schema is changed? Please don't forget.. we already ran into this issue.

nemesisdesign (changes requested):

nemesisdesign (changes requested):
we need to switch the openwisp-utils version to point to the current master for the wifi icon to work

nemesisdesign (approved):
Great :+1:

PR #375, opened 2022-05-18
[feature] Added inline for WifiSession in DeviceAdmin #362

### LINKED ISSUE #362: [feature] Add WiFi clients tab to device page
Depends on https://github.com/openwisp/openwisp-monitoring/issues/360.

Add a new tab to the device page as a Tabular inline which shows the currently open wifi sessions.

Ensure the wifi client data is fetched with select_related to avoid unnecessary queries.

Make sure only the currently open sessions are shown.

The list of sessions should look similar to the mock up below, except that organization, device and stop time are not shown, and only the "online" sessions are shown, we should add a button to allow opening the full list of sessions (also the closed ones), it can be a link to the wifi sessions page with the device filter set.

wifi-client-session-list

The available filters shall be:

- start time
- stop time
- org
- group
- device name or id (using [openwisp-utils filter](https://github.com/openwisp/openwisp-utils#openwisp_utilsadmin_themefiltersinputfilter))

## Detail page

Clicking on a WiFi client entry will open its details as shown below (read only).

wifi-client-session-detail

pandafy:
Closes #362

Checks:

----
image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -820,6 +820,12 @@ You can disable this feature by configuring
 `OPENWISP_MONITORING_WIFI_SESSIONS_ENABLED <#openwisp_monitoring_wifi_sessions_enabled>`_
 setting.
 
+You can also view open WiFi sessions of a device directly from the device's change admin
+under the "WiFi Sessions" tab.
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/device-wifi-session-inline.png
+  :align: center
+
 Scheduled deletion of WiFi sessions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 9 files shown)


pandafy:
@nemesisdesign I added the button inside the "WiFi Sessions" tab instead of adding it at the top along with other buttons ("Silence Notification", "Send Command", etc.) because those buttons perform operations directly related to the device. Also, it will become crowded if we keep adding buttons on the top. 

If you prefer, we can move this button to the top.

nemesisdesign (changes requested):
I merged the latest master and fixed the tests, I did also some minor improvements.

Other problems found during testing pending solution:

- the see wifi sessions button is shown also in the other tabular inlines

I also found out https://github.com/openwisp/openwisp-monitoring/issues/378, but it's not directly caused by this PR but was introduced in the previous one.

nemesisdesign (approved):

nemesisdesign (approved):

PR #376, opened 2022-05-18
[feature] Added view for fetching dashboard timeseries data #364

### LINKED ISSUE #364: [feature] Add REST API view which allows to fetch dashboard timeseries data
The goal of this issue is to allow retrieving timeseries data and showing charts for all the devices of an organization or multiple organizations, these charts will be shown in the dashboard (index page of OpenWISP) or could be shown in third party apps.

We need to add a view which allows to fetch any timeseries data, not just for a specific device as we have now.
The view shall ensure the user has the permissions to query the data for the specified `organization_slug` which is passed as a filter parameter.
If the `organization_slug` param is not specified, the view shall automatically add all the orgs the user has access to in the query.
If the user is a superuser and the `organization_slug` param is not speicified, the results from any org will be fetched.

It shall be possible to also supply a `location_id` param.

It shall be possible to group by time as in the device metric view (1 day, 3 days, 1 week, 1 month, 1 year).

It shall be possible to export the data as in the device metrics view.

Let's keep in mind the traffic chart should be configurable via setting, the setting should allow to configure which interfaces are shown, eg:

```python
OPENWISP_MONITORING_DASHBOARD_TRAFFIC_CHART = {
    '': ['wan'],  # default value applying to all orgs
    'org1': ['eth1', 'eth2']  # this will override the chart for org1
}
```

In the example above the org org1 would count both interfaces eth1 and eth2.

Instead of `org1` (slug) we could also use the UUID of the org, let's look for similar features and settings in controller, radius and monitoring itself to maintain consistency.

To keep it simple for this iteration I think we should handle the following cases:

- user has access to only 1 org and there's no specific config for that org: use deafult
- user has access to only 1 org and there is a specific config for that org: use that conf
- superadmin or administrator with access to multiple orgs see only default conf

We can retrieve the charts by looking for metrics which do not have any related object (we consider them as general metrics), I think we should create new configurations for these charts and I think we should generate the default general metrics and charts with a data migration.
We have to change the timeseries read code too to make it handle reading tags which can have mutliple values (ifname = eth1 OR ifname = eth2).

pandafy:
Closes #364

Checks:


openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,121 @@
...
+
+logger = logging.getLogger(__name__)
+Chart = load_model('monitoring', 'Chart')
+Metric = load_model('monitoring', 'Metric')
+Organization = load_model('openwisp_users', 'Organization')
+
+
+class DashboardTimeseriesView(APIView):
+    def _get_organization_lookup(self):
+        query_orgs = self.request.query_params.get('organization_slug', '')
```
pandafy, line 23: TODO: Simplify this method

openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if not orgs:
+            return ''
+        org_lookup = []
+        for org in orgs:
+            org_lookup.append(f"organization_id = '{org}'")
+        return 'AND ({org_lookup})'.format(org_lookup=' OR '.join(org_lookup))
+
+    def get(self, request):
+        org_lookup = self._get_organization_lookup()
```
pandafy, line 53: This method is similar to the method in device_monitoring API views. We can make something reusable.

openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,121 @@
...
+        data = self._get_charts_data(charts, time, timezone, org_lookup)
+        # csv export has a different response
+        if request.query_params.get('csv'):
+            response = HttpResponse(self._get_csv(data), content_type='text/csv')
+            response['Content-Disposition'] = 'attachment; filename=data.csv'
+            return response
+        return Response(data)
+
+    def _get_charts_data(self, charts, time, timezone, org_lookup):
+        chart_map = {}
```
pandafy, line 76: This method is similar to the method in device_monitoring API views. We can make something reusable.

openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,121 @@
...
+                orgs = self.request.user.organizations_managed
+            else:
+                raise PermissionDenied
+
+        if not orgs:
+            return ''
+        org_lookup = []
+        for org in orgs:
+            org_lookup.append(f"organization_id = '{org}'")
+        return 'AND ({org_lookup})'.format(org_lookup=' OR '.join(org_lookup))
```
pandafy, line 50: I have added the logic for generating "WHERE" clause for organization lookup here.  The `Chart` model uses `timeseries_db.get_query` method to generate database query: 

https://github.com/openwisp/openwisp-monitoring/blob/d49715490be5195d29091ec767a76c2eafac53cf/openwisp_monitoring/monitoring/base/models.py#L561-L564

https://github.com/openwisp/openwisp-monitoring/blob/d49715490be5195d29091ec767a76c2eafac53cf/openwisp_monitoring/monitoring/base/models.py#L484-L496

`timeseries_db.get_query` does not build the database query but only substitutes the placeholder values 

https://github.com/openwisp/openwisp-monitoring/blob/d49715490be5195d29091ec767a76c2eafac53cf/openwisp_monitoring/db/backends/influxdb/client.py#L214-L230

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -224,6 +224,27 @@ def _get_access_tech():
...
+                'label': _('WiFi clients'),
+                'title': _('WiFi clients'),
+                'description': _('WiFi clients associated to the wireless interface.'),
+                'summary_labels': [_('Total Unique WiFi clients')],
+                'unit': '',
+                'order': 230,
+                'query': chart_query['general_wifi_clients'],
+                'query_default_param': {
+                    'organization_lookup': '',
+                },
```
pandafy, line 244: The `Chart.clean` method validates timeseries DB query. While creating Chart in database migration, the organization_lookup is absent. Therefore, I have added default query params.

openwisp_monitoring/urls.py
```diff
@@ -1,5 +1,11 @@
 from django.urls import include, path
 
 urlpatterns = [
-    path('', include('openwisp_monitoring.device.api.urls', namespace='monitoring'))
+    path(
+        '',
+        include('openwisp_monitoring.device.api.urls', namespace='device_monitoring'),
```
pandafy, line 6: I have changed the namespace here.  Is this okay? Or, should we give a different namespace to the other API?

nemesisdesign: other API for backward compatibility, we are 1.0.0 now, keep in mind

openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,148 @@
...
+            else:
+                raise PermissionDenied
+
+    def _get_query_clause(self, field, items):
+        if not items:
+            return ''
+        lookup = []
+        for item in items:
+            lookup.append(f"{field} = '{item}'")
+        return 'AND ({lookup})'.format(lookup=' OR '.join(lookup))
```
nemesisdesign, line 52: any logic related to the specific timeseries DB must be moved to the specific timeseries class which serve as an abstraction layer

nemesisdesign (changes requested):

openwisp_monitoring/device/apps.py
```diff
@@ -261,8 +263,42 @@ def register_dashboard_items(self):
...
+                'chart_quick_links': {
+                    'WiFi clients': {
+                        'url': reverse_lazy(
+                            'admin:{app_label}_{model_name}_changelist'.format(
+                                app_label=WifiSession._meta.app_label,
+                                model_name=WifiSession._meta.model_name,
+                            )
+                        ),
+                        'label': 'Open WiFi Session',
+                        'title': 'View full history of WiFi Sessions',
```
nemesisdesign, line 294: flag these as translatable please

openwisp_monitoring/device/apps.py
```diff
@@ -261,8 +263,42 @@ def register_dashboard_items(self):
...
+                'default_time': Chart.DEFAULT_TIME,
+                'chart_quick_links': {
+                    'WiFi clients': {
+                        'url': reverse_lazy(
+                            'admin:{app_label}_{model_name}_changelist'.format(
+                                app_label=WifiSession._meta.app_label,
+                                model_name=WifiSession._meta.model_name,
+                            )
+                        ),
+                        'label': 'Open WiFi Session',
```
nemesisdesign, line 293: this is inconsistent with the other button

openwisp_monitoring/monitoring/api/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import path
+
+from . import views
+
+app_name = 'monitoring_general'
+
+urlpatterns = [
+    path(
+        'api/v1/monitoring/dashboard/',
```
nemesisdesign, line 9: is this API endpoint added to the REST API endpoint list in the README?

openwisp_monitoring/monitoring/api/views.py
```diff
@@ -0,0 +1,183 @@
...
+        fileobj = StringIO()
+        csv.writer(fileobj).writerows(rows)
+        return fileobj.getvalue()
+
+    def _get_csv_header(self, chart, trace):
+        header = trace[0]
+        return f'{header} - {chart["title"]}'
+
+
+dashboard_timeseries = DashboardTimeseriesView.as_view()
```
nemesisdesign, line 183: I think as last step we shall try to remove the duplication between this view and the device monitoring view

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -224,6 +252,27 @@ def _get_access_tech():
...
+        'label': _('General Clients'),
+        'name': 'General Clients',
+        'key': 'wifi_clients',
+        'field_name': 'clients',
+        'charts': {
+            'gen_wifi_clients': {
+                'type': 'bar',
+                'label': _('General Clients'),
+                'title': _('WiFi clients'),
+                'description': _('WiFi clients associated to the wireless interface.'),
```
nemesisdesign, line 265: this will have to be updated to be more meaningful

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -204,6 +204,34 @@ def _get_access_tech():
...
+        'key': 'traffic',
+        'field_name': 'rx_bytes',
+        'related_fields': ['tx_bytes'],
+        'charts': {
+            'general_traffic': {
+                'type': 'scatter',
+                'title': _('General Traffic'),
+                'label': _('General Traffic'),
+                'description': _(
+                    'Network traffic, download and upload, ' 'measured in GB.'
```
nemesisdesign, line 219: this will have to be updated to be more meaningful

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -153,3 +156,63 @@ def test_restoring_deleted_device(self):
...
+
+class TestDashboardCharts(
+    SeleniumTestMixin, TestDeviceMonitoringMixin, StaticLiveServerTestCase
+):
+    def test_dashboard_timeseries_charts(self):
+        self.login()
+        try:
+            WebDriverWait(self.web_driver, 5).until(
+                EC.visibility_of_element_located(
+                    (By.XPATH, '//*[@id="ow-chart-inner-container"]')
```
nemesisdesign, line 169: can we use CSS instead of xpath when possible please? It's more familiar to most people.

README.rst
```diff
@@ -96,6 +96,7 @@ Available Features
 * CSV Export of monitoring data
 * An overview of the status of the network is shown in the admin dashboard,
   a chart shows the percentages of devices which are online, offline or having issues;
+  timeseries charts are shown for WiFi sessions and Traffic of all managed devices,
```
nemesisdesign, line 99: there are also two timeseries charts which show respectively the total unique wifi clients and the the traffic flowing to the network.

I think we need to add a new brief section that gives a bit more info about this feature, shows a screenshot and explains that the traffic chart is configurable and needs to be adapted according to the name of the main interface the traffic flows through.

README.rst
```diff
@@ -1196,6 +1197,44 @@ being set to ``True`` (which is the default).
 You can turn this off if you do not use the geographic features
 of OpenWISP.
 
+``OPENWISP_MONITORING_DASHBOARD_TRAFFIC_CHART``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------------------+
+| **type**:    | ``dict``                 |
++--------------+--------------------------+
+| **default**: | ``{'__all__': ['wan']}`` |
```
nemesisdesign, line 1206: I realized this default value won't work on most systems because openwrt doens't call the interfaces wan but eth1 or eth0.1, so maybe a better default value is a list containing all these values (wan, eth1, eth0.1).

nemesisdesign (changes requested):
It's looking good! See my comments below.

I will need to do more testing and reviewing.

openwisp_monitoring/monitoring/api/views.py
pandafy: Yes, I have already planned for this. https://github.com/openwisp/openwisp-monitoring/pull/376#discussion_r878154238

README.rst
nemesisdesign: The default value here needs to be updated.

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -468,6 +521,8 @@ def _get_access_tech():
 
 DEFAULT_CHARTS = {}
 
+DEFAULT_DASHBOARD_TRAFFIC_CHART = {'__all__': ['wan', 'eth1', 'eth0_2']}
```
nemesisdesign, line 524: I think we should set `eth0.2` here and make the necessary changes so the system can handle this properly (it shall convert it automatically to `eth0_2` and we need a test for this.

nemesisdesign (changes requested):

openwisp_monitoring/monitoring/configuration.py
pandafy: > I think we should set `eth0.2` here and make the necessary changes so the system can handle this properly (it shall convert it automatically to `eth0_2` and we need a test for this.

There's already a test interfaces added by user. I will update the code to also include the default interface in the conversion.

README.rst
```diff
@@ -800,6 +802,22 @@ Mobile Access Technology in use
 .. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/access-technology.png
   :align: center
 
+Dashboard Monitoring Charts
+---------------------------
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/dashboard-charts.png
```
nemesisdesign, line 808: please create a `1.1/` directory for these new charts

requirements.txt
```diff
@@ -4,4 +4,4 @@ django-cache-memoize~=0.1.0
 django-nested-admin~=3.4.0
 netaddr~=0.8
 python-dateutil>=2.7,<3.0
-openwisp-utils[rest] @ https://github.com/openwisp/openwisp-utils/tarball/master
+openwisp-utils[rest] @ https://github.com/openwisp/openwisp-utils/tarball/dashboard-templates-after-charts
```
nemesisdesign, line 7: this is merged, you can update it

nemesisdesign (changes requested):
Great work @pandafy, please see my comments below, rebase on the latest master and take care of this comment in the other PR: https://github.com/openwisp/openwisp-monitoring/pull/381#discussion_r885052992 .

Let's not forget https://github.com/openwisp/openwisp-monitoring/issues/384.

nemesisdesign (changes requested):
Please make sure the spacing of the charts in the dashboard is consistent, the red rectangle should not overlap:

spacing1

spacing2

README.rst
```diff
@@ -1196,6 +1214,44 @@ being set to ``True`` (which is the default).
...
+
+This settings allows to configure the interfaces which should
+be included in the **General Traffic** chart in the admin dashboard.
+
+This setting should be defined in the following format:
+
+.. code-block::python
+
+    OPENWISP_MONITORING_DASHBOARD_TRAFFIC_CHART = {
+        '<organization-slug>': ['<list-of-interfaces>']
```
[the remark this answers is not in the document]
pandafy: > Did you see any similar setting in other OW modules?

Yes, https://openwisp-radius.readthedocs.io/en/stable/user/settings.html#openwisp-radius-called-station-ids

nemesisdesign (changes requested):
Almost there, see my comments.

nemesisdesign (approved):

PR #379, opened 2022-05-19
[fix] Fixed bug in deleting device with WiFiSessions #378

### LINKED ISSUE #378: [bug] Deleting devices from the admin is not possible after merging WiFi Sessions PR
Happening on master, deleting a device from the admin is not possible anymore.

Screenshot from 2022-05-18 17-08-11

pandafy:
Closes #378

Checks:

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -377,6 +377,9 @@ class WifiSessionAdmin(MultitenantAdminMixin, ReadOnlyAdmin):
 
     related_organization.short_description = _('organization')
 
+    def has_delete_permission(self, request, obj=None):
+        return super(admin.ModelAdmin, self).has_delete_permission(request, obj)
+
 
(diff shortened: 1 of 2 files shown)


openwisp_monitoring/device/admin.py
```diff
@@ -377,6 +377,9 @@ def related_organization(self, obj):
 
     related_organization.short_description = _('organization')
 
+    def has_delete_permission(self, request, obj=None):
+        return super(admin.ModelAdmin, self).has_delete_permission(request, obj)
```
nemesisdesign, line 381: Do you think this is a bug in ReadOnlyAdmin? May it make sense to solve this problem at the root?

nemesisdesign (approved):

openwisp_monitoring/device/admin.py
pandafy: If this is not an intentional feature, then yes.

PR #381, opened 2022-05-26
[change] Change look of traffic charts #366

### LINKED ISSUE #366: [change] Change look of traffic charts
I want to change the look of the current traffic charts to look like the following:

traffic-complete

With this new configuration, we can show at glance the upload, download and total.

A prototype can be seen in action here, it can be also exported as HTML, which will contain the JS code for plotly:
https://chart-studio.plotly.com/~nemesisdev/1

yashikajotwani12:
The look of traffic charts is to be
modified as per the given description
in the issue.
**_Need to add the query to get the total of the download and upload of the traffic charts._** 

Screenshot (43)

**UPDATE**: 
* Added total traffic query, need to change total traffic type from _stackedbar_ to _scatter_.

Screenshot (45)

Fixes #366

Checks:


README.rst
```diff
@@ -643,7 +643,7 @@ Traffic
 +--------------------+--------------------------------------------------------------------------+
 | **type**:          | ``int``                                                                  |
 +--------------------+--------------------------------------------------------------------------+
-| **fields**:        | ``rx_bytes``, ``tx_bytes``                                               |
+| **fields**:        | ``rx_bytes``, ``tx_bytes``, ``total_bytes``                                               |
```
Aryamanz29, line 646: This line in the table should be properly aligned with the above line.

Aryamanz29 (commented):
QA check fails because :

openwisp_monitoring/device/api/views.py
```diff
@@ -235,7 +235,10 @@ def _write(self, request, pk, time=None, current=False):
                 extra_values = {
                     'tx_bytes': self._calculate_increment(
                         ifname, 'tx_bytes', ifstats['tx_bytes']
-                    )
+                    ),
+                    'total_bytes': self._calculate_increment(
+                        ifname, 'total_bytes', ifstats['rx_bytes'] + ifstats['tx_bytes']
+                    ),
```
pandafy, line 241: We don't need to store the `total_bytes` in the influxdb database. We can calculate it when we retrieve the data from the database.

yashikajotwani12: Ohh Okay! I will remove it from here. Thanks 👍🏻

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -184,10 +184,22 @@ def _get_access_tech():
         'name': '{name}',
         'key': '{key}',
         'field_name': 'rx_bytes',
-        'related_fields': ['tx_bytes'],
+        'related_fields': ['tx_bytes', 'total_bytes'],
```
pandafy, line 187: This change is also not required. We are not making any change to the metric. We are only changing the way we show the collected data to the user. 

```suggestion
        'related_fields': ['tx_bytes'],
```

pandafy (changes requested):

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -57,10 +72,6 @@
         if (type === 'histogram') {
             layout.hovermode = 'closest';
         }
-        if (type === 'stackedbar') {
-            type = 'bar';
-            layout.barmode = 'stack';
-        }
```
nemesisdesign, line 63: why is this being removed?

nemesisdesign (changes requested):
Great progress @pandafy, see my comments below, then I think we can merge.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
pandafy: this is being handled here 
https://github.com/openwisp/openwisp-monitoring/blob/8536dd63a8e4db1c6df47d0a553706218a4e3fe3/openwisp_monitoring/monitoring/static/monitoring/js/chart.js#L123-L134

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -187,20 +187,32 @@ def _get_access_tech():
...
                 'summary_labels': [
+                    _('Total traffic'),
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
                 'unit': _(' GB'),
                 'order': 240,
                 'query': chart_query['traffic'],
+                'colors': [
+                    '#6F6F6F',
```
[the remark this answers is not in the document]
pandafy: I have used the same colour in the code.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -84,15 +95,21 @@
...
             }
             var options = {
                     name: label,
                     type: type,
                     mode: mode,
                     fill: data.fill || 'tozeroy',
                     hovertemplate: [],
-                    y: []
+                    y: [],
+                    _key: key,
```
nemesisdesign, line 112: This part is a bit obscure, the options object is used to control plotly features, but this line doesn't seem related to plotly but more like an internal thing, can you please clarify this with a comment?

nemesisdesign (approved):
@pandafy I am going to merge it so you can continue with https://github.com/openwisp/openwisp-monitoring/pull/376, but please take care of the comment below.

Thank you @yashikajotwani12 for your initial effort!

openwisp_monitoring/monitoring/configuration.py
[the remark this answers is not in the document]
nemesisdesign: I may have mistyped this, I will change it later.

PR #382, opened 2022-05-27
[feature] Added REST API for wifisession and wificlients #374

### LINKED ISSUE #374: [feature] Add REST API for wifi sessions and wifi clients
Now that #372 is merged, we should add a way to fetch  / manipulate this data via REST API too. 

`/api/v1/monitoring/wifi-session/`

filters (similar to admin):

- device ID
- organization ID
- group ID
- start time
- stop time 

`/api/v1/monitoring/wifi-client/`

- device ID
- organization ID
- mac_address
- vendor

It shall give the possibility to add/edit/create items.

Both endpoints must respect multi-tenancy, docs and tests need to be updated.

The code shall go in `openwisp_monitoring.device.api`.

Aryamanz29:
**Blockers**

- [x]  https://github.com/openwisp/openwisp-monitoring/pull/517

Closes #374

https://user-images.githubusercontent.com/56113566/170691869-66b05a37-df95-43c4-bd78-a19c1d3a8e4e.mp4

## GIF/Demo

Checks:


Aryamanz29:
I need your review (when you're free) @pandafy @nemesisdesign. This will let me know that I'm going on the right direction with this PR or not 😄

nemesisdesign (changes requested):
Looks good, please add at least one test for each endpoint.

README.rst
```diff
@@ -1741,6 +1741,72 @@ If the request is made without passing the ``time`` argument, the server local t
 
 The ``time`` parameter was added to support `resilient collection and sending of data by the OpenWISP Monitoring Agent <https://github.com/openwisp/openwrt-openwisp-monitoring#collecting-vs-sending>`_.
 
+List wifi session
+#################
```
nemesisdesign, line 1745: mention available filters in the GET endpoints please

nemesisdesign (changes requested):

README.rst
Aryamanz29: Sure, I'll mention :+1:

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +22,26 @@ class Meta:
...
+            'start_time',
+            'stop_time',
+        ]
+        model = WifiSession
+
+
+class WifiClientSerializer(serializers.ModelSerializer):
+    class Meta:
+        fields = ['mac_address', 'vendor', 'ht', 'vht', 'wmm', 'wds', 'wps']
+        model = WifiClient
```
nemesisdesign, line 42: added and created fields are missing

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +22,26 @@ class Meta:
         model = DeviceMonitoring
 
 
+class WifiSessionSerializer(serializers.ModelSerializer):
+    class Meta:
+        fields = [
+            'id',
+            'device',
+            'wifi_client',
```
nemesisdesign, line 30: in order to make this API endpoint useful as the admin list of the wifi sessions, we must show the wifi client info here, I'd do that with a nested serializer which includes all the info of the wifi client here, eg:

```
{
  "id": "5db58413-548d-494b-a68a-a20ab63415be",
  "device": "be6bc1ba-4215-4512-8a1f-743b69fbfe4b",
  "client": {
      "mac_address": "b0:e1:7e:00:00:00",
      "vendor": "HUAWEI TECHNOLOGIES CO.,LTD",
      "ht": true,
      "vht": false,
      "wmm": true,
      "wds": false,
      "wps": false
  },
  "ssid": "openwisp-test",
  "interface_name": "wlan1",
  "start_time": "2022-05-30T18:54:07.225321+02:00",
  "stop_time": "2022-05-30T18:54:34.082815+02:00"
}
``` 

I would call it `client` instead of `wifi_client` for brevity and also because we know we're dealing with wifi sessions.

Make sure the query used in the view to fetch the wifi session info uses `select_related` to join wifi clients and wifi sessions to avoid generating too many queries.

For the sake of completeness I would add the following fields too:

- device_name
- organization_id
- organization_name

Still making sure all this info is fetched with `select_related`.

nemesisdesign (changes requested):
I found some problems during testing.

## Multitenancy test gave error

I tried logging in with a non superuser to test multitenancy but when I tried opening the wifi session list API endpoint I got this exception:

```
FieldError at /api/v1/monitoring/wifi-session/
Cannot resolve keyword 'organization' into field. Choices are: device, device_id, id, interface_name, modified, ssid, start_time, stop_time, wifi_client, wifi_client_id
``` 

The same happens for the wifi-client endpoint. Please test this on your end and look at the implementation of the multitenancy in other API endpoints, refer to https://github.com/openwisp/openwisp-users#organization_field and the other DRF utilities provided in OpenWISP Users.

## Filters invalid name

Do you know why there's this "[invalid name]" here?

Screenshot from 2022-05-30 16-28-24

## Browsable API selects for foreign keys

The wifi session API endpoint shows a select for "device" and "wifi client", however, these selects will be huge on production systems, slowing down the response time in the browsable API or even causing it to fail with a timeout error. For this reason, can you find a way to display this field using only the primary key?

See https://www.django-rest-framework.org/topics/browsable-api/#handling-choicefield-with-large-numbers-of-items for more info.

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +25,72 @@ class Meta:
...
+):
+    # When a relationship or ChoiceField has too many items,
+    # rendering the widget containing all the options can become very slow,
+    # and cause the browsable API rendering to perform poorly, Changed select field to input.
+    device = serializers.PrimaryKeyRelatedField(
+        queryset=Device.objects.all(), style={'base_template': 'input.html'}
+    )
+    wifi_client = serializers.PrimaryKeyRelatedField(
+        queryset=WifiClient.objects.all(), style={'base_template': 'input.html'}
+    )
```
pandafy, line 59: @Aryamanz29  can you double check if the same functionality can be added using `Meta.extra_kwargs`? I tried the following code and it seemed to be working. 

```python
class WifiSessionCreateUpdateSerializer(
    FilterByOrganizationManaged, ValidatedModelSerializer
):
    class Meta:
        model = WifiSession
        fields = ['device', 'wifi_client', 'ssid', 'interface_name']
        extra_kwargs = {
            'device': {'style': {'base_template': 'input.html'}},
            'wifi_client': {'style': {'base_template': 'input.html'}},
        }
```

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -1009,3 +1012,247 @@ def test_bearer_authentication(self):
...
+        response = self.client.get(url)
+        data = response.data
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(data), 11)
+        self.assertEqual(data['device_name'], 'default.test.device')
+        self.assertEqual(data['interface_name'], 'wlan0')
+        self.assertEqual(len(data['client']), 9)
+        self.assertEqual(data['client']['mac_address'], '22:33:44:55:66:77')
+        self.assertContains(response, 'created')
+        self.assertContains(response, 'modified')
```
pandafy, line 1070: I would add test for all the fields. It will help us to prevent accidental chaanges the serializer fields.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -1009,3 +1012,247 @@ def test_bearer_authentication(self):
...
+
+    def test_wifisession_detail_update(self):
+        device = self._create_device()
+        wifi_session = self._create_wifi_session(device=device)
+        self._login_admin()
+        url = reverse('monitoring:api_wifi_session_detail', args=(wifi_session.id,))
+        response = self.client.patch(
+            url,
+            data=json.dumps({'ssid': 'Free Public Wifi Updated'}),
```
pandafy, line 1079: Tip: You don't need to `json.dumps` here. The test client should take care of it.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -1009,3 +1012,247 @@ def test_bearer_authentication(self):
...
+            'interface_name': 'wlan0',
+        }
+        response = self.client.post(
+            url,
+            data=json.dumps(wifi_session_post_data),
+            content_type='application/json',
+        )
+        self.assertEqual(len(response.data), 4)
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(response.data['ssid'], 'Free Public Wifi Created')
```
pandafy, line 1130: I would add following assertions: 
```python
self.assertEqual(WifiSession.objects.count(), 1)
self.assertEqual(response.data['device'], device.pk)
self.assertEqual(response.data['client'], wifi_client.pk)
```

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +25,72 @@ class Meta:
...
+
+    class Meta:
+        model = WifiSession
+        fields = ['device', 'wifi_client', 'ssid', 'interface_name']
+
+
+class WifiSessionReadSerializer(FilterByOrganizationManaged, ValidatedModelSerializer):
```
pandafy, line 66: We should explore the possibility of using just one serializer. What problems did you face that made you split up the serializer?

pandafy: On creating a WifiSession, we only get following fields in the response. 

Screenshot from 2022-06-07 16-45-35

pandafy: Let's wait for @nemesisdesign opinion before proceeding with this change.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -1009,3 +1012,247 @@ def test_bearer_authentication(self):
...
+class TestWifiClientSessionApi(
+    AuthenticationMixin, TestWifiClientSessionMixin, TestCase
+):
+    def _login_admin(self):
+        User = get_user_model()
+        admin = User.objects.create_superuser('admin', 'admin', 'test@test.com')
+        self.client.force_login(admin)
+
+    # Wifi sessions
+    def test_wifisession_unauthorized_api_access(self):
```
pandafy, line 1026: Please remove these comments.

openwisp_monitoring/device/api/views.py
```diff
@@ -520,6 +540,84 @@ def _create_access_tech_chart(self, metric):
...
+            'mac_address',
+            'vendor',
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super(WifiClientFilter, self).__init__(*args, **kwargs)
+        # [invalid_name] displaying when not filtering on a model field.
+        # Need to explicitly provide the label.
+        self.filters['wifisession__device'].label = "Device"
+        self.filters['wifisession__device__organization'].label = "Device organization"
```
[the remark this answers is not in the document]
pandafy: @Aryamanz29 this will not work because one WifiClient can have multiple WifiSession.

README.rst
```diff
@@ -1741,6 +1741,130 @@ If the request is made without passing the ``time`` argument, the server local t
...
+
+.. code-block:: text
+   
+   GET /api/v1/monitoring/wifi-session/?device={device_id}
+
+You can filter using ``organization_id`` to get list of wifi sessions belongs to an organization.
+
+.. code-block:: text
+   
+   GET /api/v1/monitoring/wifi-session/?device__organization={organization_id}
```
pandafy, line 1763: In OpenWISP, we use `organization_slug` for filtering. See https://github.com/openwisp/openwisp-controller#list-locations. It will be better to maintain this consistency, 

cc @nemesisdesign

README.rst
```diff
@@ -1741,6 +1741,130 @@ If the request is made without passing the ``time`` argument, the server local t
...
+
+You can filter using ``group_id`` to get list of wifi sessions belongs to a group.
```
pandafy, line 1765:
```suggestion
You can filter using ``group_id`` to get list of wifi sessions belongs to a device group.
```

README.rst
```diff
@@ -1741,6 +1741,130 @@ If the request is made without passing the ``time`` argument, the server local t
...
+
+.. code-block:: text
+   
+   GET /api/v1/monitoring/wifi-session/?start_time={start_time}
+
+You can filter to get list of wifi sessions using ``stop_time``.
+
+.. code-block:: text
+   
+   GET /api/v1/monitoring/wifi-session/?stop_time={stop_time}
```
pandafy, line 1781: Please double check that it is possible to use comparison (`gt`, `gte`, `lt`, `lte`) for filtering with time.

pandafy (changes requested):
Good work @Aryamanz29! This PR is coming along well. 

Other requested changes
- [ ] It is possible to create WifiSession with device_id from another organization. 
- [ ] Add tests for REST API filters

README.rst
nemesisdesign: yes I think we should maintain consitency and add an equivalent filter

openwisp_monitoring/device/api/serializers.py
nemesisdesign: I asked to have nested serializer because I think it would work well here, see https://github.com/openwisp/openwisp-monitoring/pull/382#discussion_r885068006.

openwisp_monitoring/device/api/views.py
```diff
@@ -520,6 +540,84 @@ def _create_access_tech_chart(self, metric):
...
+
+
+class WifiClientFilter(FilterSet):
+    class Meta:
+        model = WifiClient
+        fields = [
+            'wifisession__device',
+            'wifisession__device__organization',
+            'mac_address',
+            'vendor',
```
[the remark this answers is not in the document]
Aryamanz29: You're right, filtering with primary key is redundant.

openwisp_monitoring/device/api/serializers.py
Aryamanz29: Yes, It works :)

openwisp_monitoring/device/tests/test_api.py
Aryamanz29: Sure

README.rst
Aryamanz29: Yes, it is possible :+1:

pandafy (changes requested):
I would add dedicated tests for multi-tenancy for each model: 
- test_wifi_clients_multi_tenancy
- test_wifi_session_multi_tenancy 
And test each operation (GET, POST, PUT, PATCH) within a subtest in these test cases.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -1076,43 +1086,143 @@ def test_wifisession_detail_update(self):
...
-        self.assertEqual(data['results'][0]['client']['vendor'], '')
-        self.assertEqual(data['results'][0]['client']['ht'], True)
-        self.assertEqual(data['results'][0]['client']['wmm'], False)
-        self.assertContains(response, 'created')
-        self.assertContains(response, 'modified')
-        self.assertIsNone(response.data['next'])
-        self.assertIsNone(response.data['previous'])
+        self.assertIn('start_time', data['results'][0].keys())
+        self.assertIn('stop_time', data['results'][0].keys())
+        self.assertIn('modified', data['results'][0].keys())
```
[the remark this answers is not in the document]
Aryamanz29: I'll update :+1:

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,77 @@ class Meta:
         model = DeviceMonitoring
 
 
+class WifiClientSerializer(FilterSerializerByOrgManaged, ValidatedModelSerializer):
+    class Meta:
+        fields = [
+            "mac_address",
```
nemesisdesign, line 32: maintain consistency of the quoting style with the rest of the codebase, we use single quotes

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,77 @@ class Meta:
...
+        # rendering the widget containing all the options can become very slow,
+        # and cause the browsable API rendering to perform poorly, Changed select field to input.
+        extra_kwargs = {
+            'device': {'style': {'base_template': 'input.html'}},
+            'wifi_client': {'style': {'base_template': 'input.html'}},
+        }
+
+    def validate_device(self, device):
+        user = self.context['request'].user
+        if user and not user.is_manager(device.organization.pk):
```
nemesisdesign, line 65: check if we can use `device.organization_id` which is less DB intensive

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,77 @@ class Meta:
...
+            "vendor",
+            "ht",
+            "vht",
+            "wmm",
+            "wds",
+            "wps",
+            "modified",
+            "created",
+        ]
+        model = WifiClient
```
nemesisdesign, line 42: move the model line as first please

openwisp_monitoring/device/api/views.py
```diff
@@ -423,6 +444,75 @@ def _create_access_tech_chart(self, metric):
...
+    queryset = WifiSession.objects.select_related(
+        'device', 'wifi_client', 'device__organization', 'device__group'
+    )
+    organization_field = 'device__organization'
+    filter_backends = [DjangoFilterBackend]
+    pagination_class = ListViewPagination
+    filterset_class = WifiSessionFilter
+
+    def get_serializer_class(self):
+        if self.request.method in ['GET']:
```
nemesisdesign, line 472: I think here we should put OPTIONS too

nemesisdesign, line 472: I think here we should put OPTIONS too

openwisp_monitoring/device/api/views.py
```diff
@@ -423,6 +444,75 @@ def _create_access_tech_chart(self, metric):
...
+
+
+class WifiSessionDetailView(ProtectedAPIMixin, RetrieveUpdateAPIView):
+    queryset = WifiSession.objects.select_related(
+        'device', 'wifi_client', 'device__organization'
+    )
+    organization_field = 'device__organization'
+
+    def get_serializer_class(self):
+        if self.request.method in ['GET']:
```
nemesisdesign, line 487: I think here we should put OPTIONS too

nemesisdesign, line 487: I think here we should put OPTIONS too

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,77 @@ class Meta:
...
+        read_only_fields = (
+            'created',
+            'modified',
+        )
+
+
+class WifiSessionCreateUpdateSerializer(
+    FilterSerializerByOrgManaged, ValidatedModelSerializer
+):
```
nemesisdesign, line 51: add here:

```python
organization_lookup = 'wifisession__device__organization__in'
```

Please add a test for this error (which shall fail without the fix suggested above), you need to do the request to the API specifying the `HTTP_ACCEPT='text/html'`:

```python
response = self.client.get(url, HTTP_ACCEPT='text/html')
```

nemesisdesign, line 51: add here:

```python
organization_lookup = 'wifisession__device__organization__in'
```

Please add a test for this error (which shall fail without the fix suggested above), you need to do the request to the API specifying the `HTTP_ACCEPT='text/html'`:

```python
response = self.client.get(url, HTTP_ACCEPT='text/html')
```

openwisp_monitoring/device/api/views.py
```diff
@@ -423,6 +444,75 @@ def _create_access_tech_chart(self, metric):
...
+            return WifiSessionReadSerializer
+        return WifiSessionCreateUpdateSerializer
+
+
+wifi_session_list = WifiSessionListCreateView.as_view()
+
+
+class WifiSessionDetailView(ProtectedAPIMixin, RetrieveUpdateAPIView):
```
nemesisdesign, line 480: Why aren't we allowing to delete objects here?

In general through the API we allow a bit more flexibility to allow to extend the capabilities of OpenWISP if any user needs it.

nemesisdesign, line 480: Why aren't we allowing to delete objects here?

In general through the API we allow a bit more flexibility to allow to extend the capabilities of OpenWISP if any user needs it.

openwisp_monitoring/device/api/views.py
```diff
@@ -423,6 +444,75 @@ def _create_access_tech_chart(self, metric):
...
+    queryset = WifiClient.objects.all()
+    filter_backends = [DjangoFilterBackend]
+    pagination_class = ListViewPagination
+    filterset_fields = ['vendor']
+
+
+wifi_client_list = WifiClientListCreateView.as_view()
+
+
+class WifiClientDetailView(ProtectedAPIMixin, RetrieveUpdateAPIView):
```
nemesisdesign, line 507: Why aren't we allowing to delete objects here?

In general through the API we allow a bit more flexibility to allow to extend the capabilities of OpenWISP if any user needs it.

nemesisdesign, line 507: Why aren't we allowing to delete objects here?

In general through the API we allow a bit more flexibility to allow to extend the capabilities of OpenWISP if any user needs it.

nemesisdesign (changes requested):
Good progress, see my comments below.

Aryamanz29:
Hey @nemesisdesign @pandafy ! Today I found some more issue in **wifi-session** & **wifi-client** endpoints with methods **POST, PUT, PATCH.**   

**Step to replicate :** 
1. Create a new user having **administrator group** and `is_admin : True` 
2. Try to update, create **wifi-session.**
 
**Error :** 
```bash
MultipleObjectsReturned at /api/v1/monitoring/wifi-session/39091d12-74ab-4f7a-bc1d-d589f2a874cd/ get() returned more than one Device -- it returned 6! 
```
Screenshot from 2022-06-15 16-08-37


3. Now try with **wifi-client (create, update).**
 **Error :** 
```bash
 `AttributeError at /api/v1/monitoring/wifi-client/00:00:5e:00:53:a2/ Organization not found, organization_field not implemented correctly
```
Screenshot from 2022-06-15 16-07-58

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,79 @@ class Meta:
         model = DeviceMonitoring
 
 
+class WifiClientSerializer(FilterSerializerByOrgManaged, ValidatedModelSerializer):
```
pandafy, line 29: `FilterSerializerByOrgManaged` mixin filters the choices in the queryset of the fields with respect the value of `organization_field`. We don't need `FilterSerializerByOrgManaged` here because there is no field which is tied to an organization.

pandafy: Yes, we should do that. We should only check if the user has permissions to manage `WifiClient`, if so they can perform required operation on the object. 

`DjangoModelPermission` should get us covered in this scenario.

Aryamanz29: Thanks !

Aryamanz29: @pandafy Some more questions:
1. It is still relevant to add a test for `test_wifi_clients_multi_tenancy`? (because now anyone can view this endpoint with right permission)?
2. Do you have any idea about [above error](https://github.com/openwisp/openwisp-monitoring/pull/382#issuecomment-1156348391) **(MultipleObjectsReturned)**  in `wifi-session` endpoint?

pandafy: One WiFIClient can have multiple WiFiSession. The WifiSession can be with devices of different organizations. I don't think the `organization_filed` implementation will work here.

Aryamanz29: @nemesisdesign I've fixed this in [56e87f1](https://github.com/openwisp/openwisp-monitoring/pull/382/commits/56e87f1b963c445e3e63bf501449bb569841ac23)  & also tested manually, I think now it is working as you've mentioned above. Now trying to fix https://github.com/openwisp/openwisp-monitoring/pull/382#issuecomment-1156348391 (MultipleObjectsReturned) in `wifi-session` endpoint.

pandafy: @Aryamanz29 can you write a test for the issue you are facing (`MultipleObjectsReturned`). It will help me easily debug that and help you.

Aryamanz29: @pandafy Please see https://github.com/openwisp/openwisp-monitoring/pull/382/commits/aa9a0f5c18c341700b379610f9e1b62b14910063 I have added both tests with **non-superuser administrator group** and with **superuser**.

openwisp_monitoring/device/api/serializers.py
```diff
@@ -20,6 +26,79 @@ class Meta:
...
+        read_only_fields = (
+            'created',
+            'modified',
+        )
+
+
+class WifiSessionCreateUpdateSerializer(
+    FilterSerializerByOrgManaged, ValidatedModelSerializer
+):
+    organization_lookup = 'wifisession__device__organization__in'
```
pandafy, line 52: This is causing the error of `MultipleObjectsRetuned`. This filters the Device object with following query: 
```
Device.objects.filter(wifisession_device_organization__in=['users-organization'])
```
This creates a join between `Device` and `WifiSession` and since there are multiple WifiSessions, it returns two device objects in the query: 

```
<QuerySet [<Device: default.test.device>, <Device: default.test.device>]>
```

This is the SQL query which is executed 
```
SELECT "config_device"."id", "config_device"."created", "config_device"."modified", "config_device"."organization_id", "config_device"."name", "config_device"."mac_address", "config_device"."key", "config_device"."model", "config_device"."os", "config_device"."system", "config_device"."notes", "config_device"."group_id", "config_device"."last_ip", "config_device"."management_ip", "config_device"."hardware_id" FROM "config_device" INNER JOIN "device_monitoring_wifisession" ON ("config_device"."id" = "device_monitoring_wifisession"."device_id") INNER JOIN "config_device" T3 ON ("device_monitoring_wifisession"."device_id" = T3."id") WHERE T3."organization_id" IN (18e9ef4bb6d64f768a4cf17de94a28b1)
```

I have opened an issue for this in openwisp-users https://github.com/openwisp/openwisp-users/issues/319

pandafy: @nemesisdesign what do you suggest to do here? Should we complete this PR without the filtering for device and wifi-client fields and open an issue for it?

pandafy (changes requested):
I did some debugging today and found the underlying cause for `MultipleObjectsReturned` error.

openwisp_monitoring/device/api/serializers.py
Aryamanz29: > @nemesisdesign I've fixed this in [56e87f1](https://github.com/openwisp/openwisp-monitoring/pull/382/commits/56e87f1b963c445e3e63bf501449bb569841ac23) & also tested manually, I think now it is working as you've mentioned above. Now trying to fix [#382 (comment)](https://github.com/openwisp/openwisp-monitoring/pull/382#issuecomment-1156348391) (MultipleObjectsReturned) in `wifi-session` endpoint.

While working on this [67d0235](https://github.com/openwisp/openwisp-monitoring/pull/382/commits/67d02356ad6da80195199f72d62f99e5dd4f2d92) I found that we can only filter **wificlient** **APIVIEWS** according to user's managed organization only when clients have session (since filtering is based on  **wifisession__device__organization**). Let say if any user **(org A)** created a new **wificlient** using API endpoint and that client doesn't have any session, then that client will be visible to **org B** user.

openwisp_monitoring/device/api/serializers.py
nemesisdesign: Can you tell me the steps to replicate this issue please?

pandafy: @nemesisdesign Aryaman wrote a test for this in https://github.com/openwisp/openwisp-monitoring/pull/382/commits/aa9a0f5c18c341700b379610f9e1b62b14910063

Aryamanz29: Hey @nemesisdesign, what you do think about https://github.com/openwisp/openwisp-monitoring/pull/382#discussion_r900104443?

nemesisdesign: I think the problem is first of all a design issue.

Let me explain the initial design.

We store the clients to avoid repeating the same information when a single client performs multiple sessions, but since a client could perform WiFi sessions in multiple organizations, it doesn't have the organization field, otherwise we would have to duplicate the client information in each organization and the goal of avoiding the repetition would not be met.

If we analyze the statement above more closely, we can figure out that the `WiFiClient` model is an implementation detail of the WiFi Sessions feature, which we hide in the admin interface, therefore I am thinking that probably my initial request to add an API endpoint for it was wrong in the first place.

Probably we have to expose only the session list information with the appropriate list filters (as in the admin) I am not sure allowing to create a session even has any use case right now.

@pandafy @Aryamanz29 what do you think?

Aryamanz29:
### `Updates`

Below is a comparison between the **WiFi session admin (list and detail view)** and the **WiFi session API (detail and list view)**:

Screenshot from 2023-05-25 16-05-33

Screenshot from 2023-05-25 15-58-35

Screenshot from 2023-05-25 15-59-07

Screenshot from 2023-05-25 15-58-57

@nemesisdesign Now, I have a question regarding how we should include WiFi client field information in the API response. Should we use a nested serializer (as demonstrated in the API detail view) or should we directly add the client serializer fields (as displayed in the API list) to the response?

nemesisdesign (commented):
@Aryamanz29 the nested serialzier solution looks nice and clean!

nemesisdesign (approved):

nemesisdesign (approved):

PR #392, opened 2022-06-02
[feature] Device status: make bridge members clickable #352

### LINKED ISSUE #352: [feature] Device status: make bridge members clickable
It would be very useful if the bridge members could be clickable and when clicking on them the UI would automatically move to show the clicked bridge member.

yashikajotwani12:
updated:
chrome-capture-2022-5-14

Fixes #352

Checks:

### DIFF
--- a/openwisp_monitoring/device/templates/admin/config/device/change_form.html
+++ b/openwisp_monitoring/device/templates/admin/config/device/change_form.html
@@ -335,7 +335,7 @@
                       <label>{% trans 'Bridge Members' %}:</label>
                       <div class="readonly">
                           {% for element in interface.bridge_members %}
-                          <a class="bridge-member" href="javascript:scrollToElement('status-if-{{ interface.name }}')">{{ element }}</a>{% if not forloop.last %}, {% endif %}
+                          <a class="bridge-member" href="javascript:scrollToElement('status-if-{{ element }}')">{{ element }}</a>{% if not forloop.last %}, {% endif %}
                           {% endfor %}
                       </div>
                   </div>
--- a/openwisp_monitoring/device/templates/admin/config/device/change_form.html
+++ b/openwisp_monitoring/device/templates/admin/config/device/change_form.html
@@ -110,7 +110,7 @@
               {% endif %}
               {% if device_data.interfaces %}
               {% for interface in device_data.interfaces %}
-              <div class="inline-group">
+              <div class="inline-group" id="status-if-{{ interface.name }}">
                   <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   {% if interface.mac %}
                   <div class="form-row">
@@ -334,7 +334,9 @@
                   <div class="form-row">
                       <label>{% trans 'Bridge Members' %}:</label>
                       <div class="readonly">
-                          {{ interface.bridge_members|join:", " }}
+                          {% for element in interface.bridge_members %}
+                          <a class="bridge-member" href="javascript:scrollToElement('status-if-{{ interface.name }}')">{{ element }}</a>{% if not forloop.last %}, {% endif %}
+                          {% endfor %}
                       </div>
                   </div>
                   {% endif %}
@@ -545,6 +547,9 @@
         }
     });
   });
+  function scrollToElement(id) {
+    document.getElementById(id).scrollIntoView();
+  }
 </script>
 {% endif %}
 {% endblock %}
--- a/openwisp_monitoring/device/tests/test_admin.py
+++ b/openwisp_monitoring/device/tests/test_admin.py
@@ -185,7 +185,7 @@ class TestAdmin(
             r1,
             """
                 <a class="bridge-member"
-                   href="javascript:scrollToElement("tap0")">
+                   href="javascript:scrollToElement('status-if-tap0')">
                     tap0
                 </a>
             """,
@@ -195,7 +195,7 @@ class TestAdmin(
             r1,
             """
                 <a class="bridge-member"
-                    href="javascript:scrollToElement("wlan0")">
+                    href="javascript:scrollToElement('status-if-wlan0')">
                      wlan0
                 </a>
             """,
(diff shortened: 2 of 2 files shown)


openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -333,8 +333,10 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   {% if interface.bridge_members %}
                   <div class="form-row">
                       <label>{% trans 'Bridge Members' %}:</label>
-                      <div class="readonly">
-                          {{ interface.bridge_members|join:", " }}
+                      <div class="readonly" >
+                          {% for element in interface.bridge_members %}
+                            <span class="bridgeMember">{{element}}</span>,
```
nemesisdesign, line 338: Please make this clickable by using an <a> element, use bridge-member as class name for consistency and also the variable syntax need to be consistent with the rest of the document, so `{{ element }}` instead of `{{element}`, please put effort in maintaining consistency, look around the code you write to make sure your code stays consistent.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -620,6 +622,12 @@ <h2>{% trans 'Charts' %}</h2>
       location.href = baseUrl + time + '&csv=1';
     })
   });
+  let elements = document.getElementsByClassName("bridgeMember");
+  for(let i=0;i<elements.length;i++) {
+      elements[i].addEventListener("click", ()=>{
+          window.scrollTo(0, 1800+500*i);
```
nemesisdesign, line 628: What is this? I believe you should scroll to a specific ID, why these numbers? Are you sure it's gonna work? Looks fragile to me.

Also here the code style is not consistent, look at quote style and spacing within the loop condition.

nemesisdesign (changes requested):
The build is failing, please run tests locally before pushing.
See also my comments below.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
yashikajotwani12: Yeah! I understand this, I will make sure to take care of this for sure and follow to make code consistent, will improve this.

openwisp_monitoring/device/templates/admin/config/device/change_form.html
atb00ker: I agree, it feels like a scary piece of code for two reasons:
1. I don't understand it completely, why is it working?
2. I think if something changes, it'll break.

Might be worth thinking of another solution indeed! 🤔

atb00ker: Ideally, it'll be something like:

```
{% for element in interface.bridge_members %}
    <span class="bridge-member" onclick="scrollToInterface(<arguement-identifing-target>)">{{ element }}</span>,
{% endfor %}

function scrollToInterface(target) {
// your-logic-here
}
```

This is still better, because there is no external loop in which scrolling part depends.
More maintainable.

> I believe you should scroll to a specific ID,

This idea sounds much less scary and reliable, can we try it out instead of those numbers? 😄

yashikajotwani12:
Screenshot from 2022-06-08 10-44-24
The test is working fine without `html=True`. If you suggest I will remove that.

nemesisdesign (changes requested):

atb00ker:
This approach looks much better! 😄

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -620,6 +625,9 @@ <h2>{% trans 'Charts' %}</h2>
       location.href = baseUrl + time + '&csv=1';
     })
   });
+    function scrollToElement(id) {
```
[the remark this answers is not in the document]
nemesisdesign: At the same level of `django.jQuery(function($) {` I meant, sorry, please fix the indentation, it shouldn't be indentent if it's not contained in any code block.

nemesisdesign (changes requested):
There's a space before the comma and is visible.
This is probably cause by the whitespace (new line) before the closing `</a>`, please try removing that whitespace to see if this goes away.

Screenshot from 2022-06-09 11-10-26

yashikajotwani12:
> There's a space before the comma and is visible. This is probably cause by the whitespace (new line) before the closing `</a>`, please try removing that whitespace to see if this goes away.
> 
> Screenshot from 2022-06-09 11-10-26

Screenshot from 2022-06-09 21-14-11

atb00ker (commented):
I think it looks good now, (although pl. wait for Fed's review)! 😓

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -334,7 +334,9 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                   <div class="form-row">
                       <label>{% trans 'Bridge Members' %}:</label>
                       <div class="readonly">
-                          {{ interface.bridge_members|join:", " }}
+                          {% for element in interface.bridge_members %}
+                          <a class="bridge-member" href="javascript:scrollToElement('bridge-member{{ forloop.counter }}')">{{ element }}</a>{% if not forloop.last %}, {% endif %}
```
nemesisdesign, line 338: Replace `bridge-member{{ forloop.counter }}` with `status-if-{{ interface.name }}`.

nemesisdesign (changes requested):
During testing I found out that this isn't working, when I am clicking on the link I am not taken to the right interface.

I will send you a link in private to a demo system where you can see this for yourself @yashikajotwani12 @atb00ker.

nemesisdesign (changes requested):
please update your local master branch and then rebase on the latest master, eg:

```
git checkout master
git pull origin master
git checkout <your-branch>
git rebase master
```

openwisp_monitoring/device/templates/admin/config/device/change_form.html
yashikajotwani12: it worked well and fine when replacing it with `status-if-{{ element }}`

nemesisdesign (approved):

PR #397, opened 2022-06-19
[ux] Show size in (KB, MB or GB) adaptively in charts #87

### LINKED ISSUE #87: [ux] Show size in (KB, MB or GB) adaptively in charts
Use the template tag [filesizeformat](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#filesizeformat) for `Monitoring Graphs`.
It formats the value to a `human-readable` file size (i.e. **13 KB**, **4.1 MB**, **102 bytes**, etc.).

For example:
```js
{{ value|filesizeformat }}
If value is 123456789, the output would be 117.7 MB.
```
There's also another template tag [humanize](https://docs.djangoproject.com/en/3.0/ref/contrib/humanize/) which helps to make it more human-readable.

We need to flag a chart in some way such that it's bytes are formatted automatically.
Currently, the values of the chart have been formatted as per the units.
This works well mostly, but sometimes it doesn't. For example, when not much traffic is being consumed, the points of the graph show numbers like **0.004 GB** which looks very bad.

yashikajotwani12:
Fixes #87

   Before        |  After
:-------------------------:|:-------------------------:
Screenshot from 2022-06-19 17-45-56 | Screenshot from 2022-06-19 17-42-21

**After**

1. Screenshot (70)

2. Screenshot (71)

**Changes**:
* All the data points have their data converted as per the range, and a particular chart has one specific unit.
* The percircles have their data converted and have different units as per the range.
* The side y-axis has a unit as same as what data points have.

**Demo**:

https://user-images.githubusercontent.com/77020164/175221352-22b969b6-05c2-499c-aac5-32bedc8369b4.mp4

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -918,6 +918,29 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
+Adaptive Filtering in Charts
+----------------------------
+
+This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart. The chart automatically sets it's unit according to a specific range of values of data. 
+For the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart. 
+The summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
+
+For Example:
+~~~~~~~~~~~~
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
+   :align: center
+
+configuration
+~~~~~~~~~~~~~
+.. code-block:: python
+    
+    'traffic': {
+                'unit': '{adaptive_bytes}',
+                },
+
+This is enabled for the traffic charts where the data is in `B`, `KB`, `MB` and `GB` respectively.
+
 Settings
 --------
 
--- a/README.rst
+++ b/README.rst
@@ -918,15 +918,11 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
-Adaptive Filtering in Charts
-----------------------------
+Byte Adaptive Measuring in Charts
+---------------------------------
 
-This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart. The chart automatically sets it's unit according to a specific range of values of data. 
-For the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart. 
-The summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
-
-For Example:
-~~~~~~~~~~~~
+This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart, summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
+The chart automatically sets it's unit according to a specific range of values of data, for the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart.
 
 .. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
    :align: center
@@ -936,7 +932,7 @@ configuration
 .. code-block:: python
     
     'traffic': {
-                'unit': '{adaptive_bytes}',
+                'unit': 'adaptive_bytes',
                 },
 
 This is enabled for the traffic charts where the data is in `B`, `KB`, `MB` and `GB` respectively.
--- a/README.rst
+++ b/README.rst
@@ -937,12 +937,13 @@ which makes it easier to understand the unit of points in the chart.
 configuration
 ~~~~~~~~~~~~~
 .. code-block:: python
-    
-    'traffic': {
-                    'unit': 'adaptive_bytes',
-                },
 
-This is enabled for the traffic charts where the data is in `B`, `KB`, `MB` and `GB` respectively.
+    'traffic': {
+        # other configurations for this chart
+        'unit': 'adaptive_bytes',
+    },
+
+The default traffic chart uses this feature to automatically adapt units for the traffic data.
 
 Settings
 --------
@@ -1384,17 +1385,19 @@ This setting allows to define additional charts or to override
 the default chart configuration defined in
 ``openwisp_monitoring.monitoring.configuration.DEFAULT_CHARTS``.
 
-For example, if you want to change the traffic chart to show
-MB (megabytes) instead of GB (Gigabytes) you can use:
+For example, the units in the traffic charts automatically 
+adapt according to the traffic data and is set to ``adaptive_bytes``
+by default.
 
 .. code-block:: python
 
     OPENWISP_MONITORING_CHARTS = {
         'traffic': {
-            'unit': ' MB',
+            'unit': 'adaptive_bytes',
             'description': (
                 'Network traffic, download and upload, measured on '
-                'the interface "{metric.key}", measured in GB.'
+                'the interface "{metric.key}", and automatically '
+                'adapt units for the traffic data.'
             ),
             'query': {
                 'influxdb': (
--- a/README.rst
+++ b/README.rst
@@ -921,18 +921,25 @@ if there's anything that is not working as intended.
 Byte Adaptive Measuring in Charts
 ---------------------------------
 
-This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart, summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
-The chart automatically sets it's unit according to a specific range of values of data, for the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart.
-
 .. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
    :align: center
+
+This behavior helps us to show the data and unit of the chart in a more readable way,
+the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have
+same unit to maintain consistency in the chart, summarycircles have different units
+and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more 
+readable way.
+
+The chart automatically sets it's unit according to a specific range of values of data,
+for the quick glance the Y-Axis Legend has same unit as the datapoints 
+which makes it easier to understand the unit of points in the chart.
 
 configuration
 ~~~~~~~~~~~~~
 .. code-block:: python
     
     'traffic': {
-                'unit': 'adaptive_bytes',
+                    'unit': 'adaptive_bytes',
                 },
 
 This is enabled for the traffic charts where the data is in `B`, `KB`, `MB` and `GB` respectively.
--- a/README.rst
+++ b/README.rst
@@ -925,9 +925,9 @@ Byte Adaptive Measuring in Charts
    :align: center
 
 This behavior helps us to show the data and unit of the chart in a more readable way,
-the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have
+the units are shown in `B`, `KB`, `MB`, `GB` and `TB` respectively, and datapoints have
 same unit to maintain consistency in the chart, summarycircles have different units
-and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more 
+and are shown in `B`, `KB`, `MB`, `GB` and `TB` to overview chart summary in a more 
 readable way.
 
 The chart automatically sets it's unit according to a specific range of values of data,
@@ -1394,7 +1394,7 @@ MB (megabytes) instead of GB (Gigabytes) you can use:
             'unit': ' MB',
             'description': (
                 'Network traffic, download and upload, measured on '
-                'the interface "{metric.key}", measured in MB.'
+                'the interface "{metric.key}", measured in GB.'
             ),
             'query': {
                 'influxdb': (
--- a/README.rst
+++ b/README.rst
@@ -818,6 +818,26 @@ You can configure the interfaces included in the **General traffic chart** using
 the `"OPENWISP_MONITORING_DASHBOARD_TRAFFIC_CHART"
 <#openwisp_monitoring_dashboard_traffic_chart>`_ setting.
 
+Adaptive byte charts
+--------------------
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/adaptive-chart.png
+   :align: center
+
+When configuring charts, it is possible to flag their unit
+as ``adaptive_bytes``, this allows to make the charts more readable because
+the units are shown in either `B`, `KB`, `MB`, `GB` and `TB` depending on
+the size of each point, the summary values and Y axis are also resized.
+
+Example taken from the default configuration of the traffic chart:
+
+.. code-block:: python
+
+    'traffic': {
+        # other configurations for this chart
+        'unit': 'adaptive_bytes',
+    },
+
 Monitoring WiFi Sessions
 ------------------------
 
@@ -917,33 +937,6 @@ This check runs periodically, but it is also triggered whenever the
 configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
-
-Byte Adaptive Measuring in Charts
----------------------------------
-
-.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
-   :align: center
-
-This behavior helps us to show the data and unit of the chart in a more readable way,
-the units are shown in `B`, `KB`, `MB`, `GB` and `TB` respectively, and datapoints have
-same unit to maintain consistency in the chart, summarycircles have different units
-and are shown in `B`, `KB`, `MB`, `GB` and `TB` to overview chart summary in a more 
-readable way.
-
-The chart automatically sets it's unit according to a specific range of values of data,
-for the quick glance the Y-Axis Legend has same unit as the datapoints 
-which makes it easier to understand the unit of points in the chart.
-
-configuration
-~~~~~~~~~~~~~
-.. code-block:: python
-
-    'traffic': {
-        # other configurations for this chart
-        'unit': 'adaptive_bytes',
-    },
-
-The default traffic chart uses this feature to automatically adapt units for the traffic data.
 
 Settings
 --------
@@ -1385,7 +1378,7 @@ This setting allows to define additional charts or to override
 the default chart configuration defined in
 ``openwisp_monitoring.monitoring.configuration.DEFAULT_CHARTS``.
 
-For example, the units in the traffic charts automatically 
+For example, the units in the traffic charts automatically
 adapt according to the traffic data and is set to ``adaptive_bytes``
 by default.
 
--- a/README.rst
+++ b/README.rst
@@ -1379,28 +1379,16 @@ This setting allows to define additional charts or to override
 the default chart configuration defined in
 ``openwisp_monitoring.monitoring.configuration.DEFAULT_CHARTS``.
 
-For example, the units in the traffic charts automatically
-adapt according to the traffic data and is set to ``adaptive_bytes``
-by default.
+In the following example, we modify the description of the traffic chart:
 
 .. code-block:: python
 
     OPENWISP_MONITORING_CHARTS = {
         'traffic': {
-            'unit': 'adaptive_bytes',
             'description': (
                 'Network traffic, download and upload, measured on '
-                'the interface "{metric.key}", and automatically '
-                'adapt units for the traffic data.'
+                'the interface "{metric.key}", custom message here.'
             ),
-            'query': {
-                'influxdb': (
-                    "SELECT SUM(tx_bytes) / 1000000 AS upload, "
-                    "SUM(rx_bytes) / 1000000 AS download FROM {key} "
-                    "WHERE time >= '{time}' AND content_type = '{content_type}' "
-                    "AND object_id = '{object_id}' GROUP BY time(1d)"
-                )
-            },
         }
     }
 
@@ -1435,7 +1423,7 @@ In case you just want to change the colors used in a chart here's how to do it:
 
     OPENWISP_MONITORING_CHARTS = {
         'traffic': {
-            'colors': ['#000000', '#cccccc']
+            'colors': ['#000000', '#cccccc', '#111111']
         }
     }
 
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -199,8 +199,7 @@ DEFAULT_METRICS = {
                 'label': _('Traffic'),
                 'description': _(
                     'Network traffic (total, download and upload) '
-                    'of the interface "{ifname}", and automatically '
-                    'adapt units for the traffic data.'
+                    'of the interface "{ifname}".'
                 ),
                 'summary_labels': [
                     _('Total traffic'),
@@ -236,9 +235,7 @@ DEFAULT_METRICS = {
                 'title': _('General Traffic'),
                 'label': _('General Traffic'),
                 'description': _(
-                    'Network traffic of the whole network'
-                    ' (total, download, upload), and automatically '
-                    'adapt units for the traffic data.'
+                    'Network traffic of the whole network ' '(total, download, upload).'
                 ),
                 'summary_labels': [
                     _('Total traffic'),
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -243,7 +243,7 @@ DEFAULT_METRICS = {
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'unit': _(' GB'),
+                'unit': 'adaptive_bytes',
                 'order': 240,
                 'query': chart_query['general_traffic'],
                 'query_default_param': {
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -206,7 +206,9 @@ DEFAULT_METRICS = {
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'unit': _(' GB'),
+                'conversion': {
+                    'filter': '{unit}',
+                },
                 'order': 240,
                 'query': chart_query['traffic'],
                 'colors': [
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -235,7 +235,7 @@ DEFAULT_METRICS = {
                 'title': _('General Traffic'),
                 'label': _('General Traffic'),
                 'description': _(
-                    'Network traffic of the whole network ' '(total, download, upload).'
+                    'Network traffic of the whole network (total, download, upload).'
                 ),
                 'summary_labels': [
                     _('Total traffic'),
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -206,7 +206,7 @@ DEFAULT_METRICS = {
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'unit': '{adaptive_bytes}',
+                'unit': 'adaptive_bytes',
                 'order': 240,
                 'query': chart_query['traffic'],
                 'colors': [
--- a/openwisp_monitoring/monitoring/configuration.py
+++ b/openwisp_monitoring/monitoring/configuration.py
@@ -206,9 +206,7 @@ DEFAULT_METRICS = {
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'conversion': {
-                    'filter': '{unit}',
-                },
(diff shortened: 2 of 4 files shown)


openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -193,6 +193,40 @@
             charts.push(options);
         }
         charts = sortByTraceOrder(data.trace_order, charts, '_key');
+        if(data.trace_order !== undefined)  {
+            var total_traffic_charts = charts[0].y;
+            var sum = 0;
+            for(i=0; i<total_traffic_charts.length; i++){
+                sum += total_traffic_charts[i];
+            }
```
atb00ker, line 201: Nitpick, you don't *need* to do this change, but just want to teach you a cleaner way to do this:

```
const sum = total_traffic_charts.reduce((sum, chart) => sum = sum + chart, 0);
```

Why is this better? Pl. google `pure functions`. :smile:

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -193,6 +193,40 @@
             charts.push(options);
         }
         charts = sortByTraceOrder(data.trace_order, charts, '_key');
+        if(data.trace_order !== undefined)  {
+            var total_traffic_charts = charts[0].y;
```
atb00ker, line 197: I think we can stop using `var` and use `let` and `const` only.

`const` support: https://caniuse.com/?search=const
`let` support: https://caniuse.com/?search=let

CC: @nemesisdesign

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -193,6 +193,40 @@
...
+                sum += total_traffic_charts[i];
+            }
+            var average = sum / total_traffic_charts.length;
+            if (average < 0.01){
+                for(i=0; i<total_traffic_charts.length; i++){
+                    charts[0].y[i] *= 1000000;
+                    charts[1].y[i] *= 1000000;
+                    charts[2].y[i] *= 1000000;
+                }
+                unit = 'KB';
```
atb00ker, line 209: Again, nitpick for learning, can use `map` here! 😄

atb00ker (commented):
I haven't tested it but some early nitpicks:

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
yashikajotwani12: Actually! I did this in my first commit, but to maintain consistency I again updated this.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
yashikajotwani12: I agree this looks cleaner :+1:

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -206,7 +206,9 @@ def _get_access_tech():
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'unit': _(' GB'),
+                'conversion': {
+                    'filter': '{unit}',
```
nemesisdesign, line 210: this name kind of puzzles me, why did you choose `{unit}`?

nemesisdesign (changes requested):
Good progress, for the moment the most obvious points to improve I see are the following, but first, let's look at the screenshot:

yashika-1

1. there should be a space between the number and the unit, so 20GB > 20 GB
2. in the example I'm showing, the unit chosen is MB but I think GB would be more suited, what do you think?

Regarding the code, I think we can see this in our next meeting, but for the moment @atb00ker lert's focus on the functionality, let's make it work as we need first, that's the most important thing, code quality comes later.

nemesisdesign (changes requested):
Here's another thing which I have noticed during my testing:

Screenshot from 2022-06-21 23-52-18

I would expect to see MB in the summary in a case like this.

openwisp_monitoring/monitoring/configuration.py
yashikajotwani12: > Here's another thing which I have noticed during my testing:
> 
> Screenshot from 2022-06-21 23-52-18
> 
> I would expect to see MB in the summary in a case like this.

yeah! I did not apply those conditions for percircles, but I would change.

yashikajotwani12:
> Here's another thing which I have noticed during my testing:
> 
> Screenshot from 2022-06-21 23-52-18
> 
> I would expect to see MB in the summary in a case like this.

Yeah! I did not apply those conditions for percircles(because it contains the total data) but would apply.

openwisp_monitoring/monitoring/configuration.py
yashikajotwani12: I choose `{unit}` because like the conversion happening in values would pass a specific unit either of these `GB, MB, KB`. If you suggest something else I will update it.

yashikajotwani12: Yeah! as an average of the data would have been less than `1 GB`, that's why it would show all data in `MB`, basically where the points which have no data are also calculated in the average that is why the average has been lowered and the data it converts shown to be in  `MB`.

yashikajotwani12:
> Good progress, for the moment the most obvious points to improve I see are the following, but first, let's look at the screenshot:
> 
> yashika-1
> 
> 1. there should be a space between the number and the unit, so 20GB > 20 GB
> 2. in the example I'm showing, the unit chosen is MB but I think GB would be more suited, what do you think?
> 
> Regarding the code, I think we can see this in our next meeting, but for the moment @atb00ker lert's focus on the functionality, let's make it work as we need first, that's the most important thing, code quality comes later.


Yeah! as an average of the data would have been less than `1 GB`, that's why it would show all data in `MB`, basically where the points which have no data are also calculated in the average that is why the average has been lowered and the data it converts shown to be in  `MB`.

openwisp_monitoring/monitoring/configuration.py
yashikajotwani12: > Good progress, for the moment the most obvious points to improve I see are the following, but first, let's look at the screenshot:
> 
> yashika-1
> 
> 1. there should be a space between the number and the unit, so 20GB > 20 GB
> 2. in the example I'm showing, the unit chosen is MB but I think GB would be more suited, what do you think?
> 
> Regarding the code, I think we can see this in our next meeting, but for the moment @atb00ker lert's focus on the functionality, let's make it work as we need first, that's the most important thing, code quality comes later.


Yeah! as an average of the data would have been less than `1 GB`, that's why it would show all data in `MB`, basically where the points which have no data are also calculated in the average that is why the average has been lowered and the data it converts shown to be in  `MB`.

nemesisdesign (changes requested):
Testing round:

Screenshot from 2022-06-23 17-56-04

Why 8400 KB and not 8.4 MB?

nemesisdesign (changes requested):
@yashikajotwani12 in your algorithm, when points are empty (no traffic) are you taking those into account or not?

If in 30 days 27 days are missing, then the average should be counted only on 3 days, not 30, let me know if it's not clear.

yashikajotwani12:
> Testing round:
> 
> Screenshot from 2022-06-23 17-56-04
> 
> Why 8400 KB and not 8.4 MB?

I did this intentionally, I mentioned in the PR description that the percircles won't have the same unit, I thought that if we cannot have different units for datapoints but we can have it for percircles, the value would be `0.0084 GB` that why it got converted to `KB`. If you want like a same single unit for the percircles, we can change it.

yashikajotwani12:
> @yashikajotwani12 in your algorithm, when points are empty (no traffic) are you taking those into account or not?
> 
> If in 30 days 27 days are missing (because there's no data for those day, say the device was offline), then the average should be counted only on the 3 days which have data, not on the whole month (30 days), let me know if it's not clear.

Yeah! I understood, I am taking the whole length for the average, I would change it and  I will take account of those days which have data and then calculate the average rather than dividing with the whole length which includes non-data points too. :+1:

atb00ker:
No data scenario seems to be working fine.
n-data

atb00ker:
> I did this intentionally, I mentioned in the PR description that the percircles won't have the same unit, 

I don't think they need to be the same unit, different units should be good for these circles because unit it's abundantly clear.
However, let @nemesisdesign confirm, because even I feel it should be 8.4 MB because that looks... better!

Maybe, above 1000 KBs we should start showing as MB?
wdyt @nemesisdesign ?

openwisp_monitoring/monitoring/configuration.py
nemesisdesign: what if we change this to:

```
'unit': 'adaptive_bytes`,
```

Then just pass `{unit}` where the code needs this, if at all.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -265,7 +311,27 @@
                     percircleOptions.progressBarColor = data.colors[data.trace_order.indexOf(key)];
                 }
                 percircles.push(percircleOptions);
+                if(data.trace_order !== undefined) {
```
nemesisdesign, line 314: same here regarding `data.trace_order !== undefined` and `adaptiveFilterSummary(data, percircles)`.

README.rst
```diff
@@ -918,6 +918,29 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
+Adaptive Filtering in Charts
+----------------------------
+
+This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart. The chart automatically sets it's unit according to a specific range of values of data. 
+For the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart. 
+The summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
```
nemesisdesign, line 926: Please break the lines at a length similar to the rest of the document.

README.rst
```diff
@@ -918,6 +918,29 @@ configuration status of a device changes, this ensures the check reacts
...
+
+For Example:
+~~~~~~~~~~~~
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
+   :align: center
```
nemesisdesign, line 932: let's move this after the first heading and get rid of this heading.

README.rst
```diff
@@ -918,6 +918,29 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
+Adaptive Filtering in Charts
```
nemesisdesign, line 921: I'd call this "Byte Adaptive Measuring in Charts"

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,72 @@
         }
         return newArr;
     }
+    function adaptiveFilterPoints(charts, layout, adaptive_bytes, data) {
```
nemesisdesign, line 19: leave one blank line before please

nemesisdesign (changes requested):

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,75 @@
...
+            }
+            layout.yaxis.title = 'KB';
+            unit = 'KB';
+        }
+        else if (average < 1) {
+            for(i=0; i<newArr.length; i++) {
+                multiplier = 1000;
+                charts[0].y[i] = Math.round((charts[0].y[i] * multiplier) * 100) / 100;
+                charts[1].y[i] = Math.round((charts[1].y[i] * multiplier) * 100) / 100;
+                charts[2].y[i] = Math.round((charts[2].y[i] * multiplier) * 100) / 100;
```
nemesisdesign, line 48: isn't the same operation repeated multiple times here and above? 

What about:

```javascript
adaptValue = function(value, multiplier) {
  return Math.round((value * multiplier) * 100) / 100;
}
```

README.rst
```diff
@@ -918,6 +918,25 @@ configuration status of a device changes, this ensures the check reacts
...
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
+   :align: center
+
+configuration
+~~~~~~~~~~~~~
+.. code-block:: python
+    
+    'traffic': {
+                'unit': 'adaptive_bytes',
+                },
```
nemesisdesign, line 936: fix the indentation here please

README.rst
nemesisdesign: I meant to move it right after 

```
Byte Adaptive Measuring in Charts
---------------------------------
```

README.rst
```diff
@@ -918,6 +918,25 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
+Byte Adaptive Measuring in Charts
+---------------------------------
+
+This behavior helps us to show the data and unit of the chart in a more readable way, the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have same unit to maintain consistency in the chart, summarycircles have different units and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more readable way.
+The chart automatically sets it's unit according to a specific range of values of data, for the quick glance the Y-Axis Legend has same unit as the datapoints which makes it easier to understand the unit of points in the chart.
+
```
nemesisdesign, line 926: please break the lines above at a line length similar to the rest of the document, eg:

```
This behavior helps us to show the data and unit of the chart in a more readable way,
the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have
same unit to maintain consistency in the chart, summarycircles have different units
... etc ...
```

nemesisdesign (changes requested):

nemesisdesign (commented):
Please also try that change we discussed regarding the single data points in the overlay.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
yashikajotwani12: Yeah! I will improve this

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,121 @@
...
+
+    function adaptiveFilterPoints(charts) {
+
+        var convertToAdaptiveBytes = function(charts, type, i, multiplier, unit) {
+            charts[type].y[i] = Math.round((charts[type].y[i] * multiplier)*100)/100;
+            charts[type].hovertemplate[i] = charts[type].y[i] + ' ' + unit;
+        };
+
+        for(var i=0; i<charts[0].y.length; i++)
+        {
```
nemesisdesign, line 28:
```suggestion
        for(var i=0; i<charts[0].y.length; i++) {
```

Why suddenly the style of indenting these curly braces is changed? Can you maintain consistency with the rest of the code please?

Applies for all the other lines as well.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -193,6 +308,12 @@
             charts.push(options);
         }
         charts = sortByTraceOrder(data.trace_order, charts, '_key');
+
+        if(unit == 'adaptive_bytes') {
```
nemesisdesign, line 312:
```suggestion
        if (unit == 'adaptive_bytes') {
```

Let's leave a space before the rounded parenthesis in code blocks like this one or in `for`, please apply the same changes to other lines as well.

@codesankalp @pandafy how can we enforce this in the linter?

nemesisdesign (changes requested):
Great progress Yashika, please see my comments below.

## Adaptive quirk

Screenshot from 2022-06-30 15-39-13

In this case, I would have expected 3000 KB to show 3 MB, can you fix this please?
If we get to a thousand of any measure, we should step up to the next measure, so 1000 B should become 1 KB, 1000 KB should become 1 MB, 1000 MB should become 1 GB.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
codesankalp: I think we should start using `eslint` in `openwisp-utils`, It is much more configurable than any other linter package.
Also, Django is using this as a linter in JavaScript code: https://github.com/django/django/blob/main/.eslintrc

nemesisdesign (changes requested):
I found another issue.

When a device is offline and is not sending any data, we should see something like the following (N/A stands for not applicable):

Screenshot from 2022-06-30 15-48-52

With this PR instead, I see that even when there's no data sent, we see 0 B.

Screenshot from 2022-06-30 15-46-05

This is an issue because this way we cannot differentiate at glance between when the device is offline or online but not doing any traffic.

yashikajotwani12:
> Screenshot from 2022-06-30 15-46-05
**Update**: 
CPT2207011633-1440x350

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,114 @@
...
+        if (average == 0) {
+            layout.yaxis.title = 'B';
+        }
+        else if (average < 0.001) {
+            layout.yaxis.title = 'KB';
+        }
+        else if (average < 1) {
+            layout.yaxis.title = 'MB';
+        }
+        else {
```
nemesisdesign, line 102: Can you please add also support for TB (on some large systems this can happen).

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,114 @@
...
+                convertToAdaptiveBytes(charts, 2, i, 1000, 'MB');
+            }
+            else {
+                convertToAdaptiveBytes(charts, 2, i, 1, 'GB');
+            }
+        }
+    }
+
+    function adaptiveFilterLayout(charts, layout) {
+        var newArr = charts[0].y, sum = 0, count = 0;
```
nemesisdesign, line 83: I don't think `newArr` is a good variable name here, because it's not really a new array but it's the same array which we are reading. I'd call it simply `y`.

README.rst
```diff
@@ -918,6 +918,32 @@ configuration status of a device changes, this ensures the check reacts
...
+Byte Adaptive Measuring in Charts
+---------------------------------
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/chartsize.png
+   :align: center
+
+This behavior helps us to show the data and unit of the chart in a more readable way,
+the units are shown in `B`, `KB`, `MB` and `GB` respectively, and datapoints have
+same unit to maintain consistency in the chart, summarycircles have different units
+and are shown in `B`, `KB`, `MB` and `GB` to overview chart summary in a more
```
nemesisdesign, line 930: if you can implement the TB as I am asking, here and below we'll have to update it to mention it.

nemesisdesign (changes requested):
We'll have to modify this example here: https://github.com/openwisp/openwisp-monitoring#openwisp_monitoring_charts

Because it doesn't make sense anymore to change the measure there because it will break this behavior, since I understand that the code is assuming sizes in GB.

Maybe we can show an example that only changes the description of the chart and nothing else, what do you think?

nemesisdesign (changes requested):
I loaded the latest version of this PR on the demo system and in the main dashboard, in the general traffic chart, I see the following:

Screenshot from 2022-07-01 12-20-57

Shouldn't 0.42 GB be rendered as 420 MB?

yashikajotwani12:
> I loaded the latest version of this PR on the demo system and in the main dashboard, in the general traffic chart, I see the following:
> 
> Screenshot from 2022-07-01 12-20-57
> 
> Shouldn't 0.42 GB be rendered as 420 MB?
Yeah! cause it has not been updated for General Traffic.

nemesisdesign:
> > I loaded the latest version of this PR on the demo system and in the main dashboard, in the general traffic chart, I see the following:
> > Screenshot from 2022-07-01 12-20-57
> > Shouldn't 0.42 GB be rendered as 420 MB?
> 
> Yeah! cause it has not been updated for General Traffic, we need to change the conf of general traffic

Ah ok go ahead please.

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -243,7 +243,7 @@ def _get_access_tech():
                     _('Total download traffic'),
                     _('Total upload traffic'),
                 ],
-                'unit': _(' GB'),
+                'unit': 'adaptive_bytes',
```
nemesisdesign, line 246: reminder: update help text (unit of measure is now adaptive so we have to account for this)

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,128 @@
...
     }
+
+    function adaptiveFilterPoints(charts, yRawVal) {
+
+        var convertToAdaptiveBytes = function(charts, type, i, multiplier, unit) {
+            charts[type].y[i] = Math.round((charts[type].y[i] * multiplier) * 100) / 100;
+            charts[type].hovertemplate[i] = charts[type].y[i] + ' ' + unit;
+        };
+
+        for (var i=0; i<charts[0].y.length; i++) {
```
pandafy, line 27: @yashikajotwani12 I think the following code can be reduced by using nested `for` loops. E.g.: 

```js
for (var i=0; i<charts.length; i++) {
    for (var j=0; j<charts[i].y.length; j++) {
          // logic for converting bytes 
   }
}
```

Is there any specific reason that you wrote it in the following manner (does logic of these loops differ in a way which is not immediately visible) ?

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,128 @@
         }
         return newArr;
     }
+
+    function adaptiveFilterPoints(charts, yRawVal) {
+
+        var convertToAdaptiveBytes = function(charts, type, i, multiplier, unit) {
+            charts[type].y[i] = Math.round((charts[type].y[i] * multiplier) * 100) / 100;
+            charts[type].hovertemplate[i] = charts[type].y[i] + ' ' + unit;
+        };
```
pandafy, line 25: You can improve this function which will also improve the readability. This is an utility function. This does not need to know anything about the chart because it does not perform any operation on it. I would update the signature of this function in following way.  


```js
var convertToAdaptiveBytes = function(value, multiplier) {
    return  Math.round((value * multiplier) * 100) / 100;
}
```
And you can do assignments on `charts[type].y[i]` and `charts[type].hovertemplate[i]` in the loop that calls this function below. 


@yashikajotwani12 work on this suggestion at the very last, because it does not affect the functionality anyway. Also, wait for green flag from @nemesisdesign on this suggestion before proceeding.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,128 @@
...
+        }
+        else if (average < 1) {
+            layout.yaxis.title = 'MB';
+        }
+        else if (average < 1000) {
+            layout.yaxis.title = 'GB';
+        }
+        else if (average > 1000) {
+            layout.yaxis.title = 'TB';
+        }
```
pandafy, line 113:
```suggestion
        else if (average >= 1000) {
            layout.yaxis.title = 'TB';
        }
```

Otherwise, 1TB data would appear in bytes.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,128 @@
...
+            else if (charts[2].y[i] < 1000) {
+                convertToAdaptiveBytes(charts, 2, i, 1, 'GB');
+            }
+            else {
+                convertToAdaptiveBytes(charts, 2, i, 0.001, 'TB');
+            }
+        }
+    }
+
+    function adaptiveFilterLayout(charts, layout) {
```
pandafy, line 91: I would add a comment here explaining what this function does.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -193,6 +315,16 @@
             charts.push(options);
         }
         charts = sortByTraceOrder(data.trace_order, charts, '_key');
+
+        if (unit == 'adaptive_bytes') {
+            var yRawVal;
+            for( i =0; i<charts.length; i++){
```
pandafy, line 321: Please pay attention to the formatting. 


This is something that I would expect the QA checks to take care for me. I will look into QA checks later on.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
pandafy, line 91: @yashikajotwani12 it is not clear to me how this function is related to the `filter` as indicated by the name. Can you please explain it?

pandafy (changes requested):
I have manually tested this PR locally. I found this weird thing in one of the charts 

Screenshot from 2022-07-04 22-02-16

The unit says that the data is in KB. But if I try to infer it from the graph, I would say it is somewhere about 100 MB given that the y-scale is in MB. I think this is happening because different traces are plotted in different units. 

cc @nemesisdesign @atb00ker

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
yashikajotwani12: Because the Y-axis layout has a specific unit that has been set by calculating the values of data points, basically filtering through the average of the set of values.

README.rst
```diff
@@ -918,6 +918,32 @@ configuration status of a device changes, this ensures the check reacts
...
+for the quick glance the Y-Axis Legend has same unit as the datapoints 
+which makes it easier to understand the unit of points in the chart.
+
+configuration
+~~~~~~~~~~~~~
+.. code-block:: python
+    
+    'traffic': {
+                    'unit': 'adaptive_bytes',
+                },
```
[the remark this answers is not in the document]
yashikajotwani12: Yeah! This is because every data point has a different unit for total, download, and upload.

yashikajotwani12:
> I have manually tested this PR locally. I found this weird thing in one of the charts
> 
> Screenshot from 2022-07-04 22-02-16
> 
> The unit says that the data is in KB. But if I try to infer it from the graph, I would say it is somewhere about 100 MB given that the y-scale is in MB. I think this is happening because different traces are plotted in different units.
> 
> cc @nemesisdesign @atb00ker



> I have manually tested this PR locally. I found this weird thing in one of the charts
> 
> Screenshot from 2022-07-04 22-02-16
> 
> The unit says that the data is in KB. But if I try to infer it from the graph, I would say it is somewhere about 100 MB given that the y-scale is in MB. I think this is happening because different traces are plotted in different units.
> 
> cc @nemesisdesign @atb00ker

Yeah! This is because every data point has a different unit for total, download, and upload.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
nemesisdesign: we'd need to rename this `getAdaptiveBytes`.

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
nemesisdesign: definitely there's a lot of repetition here, the same code is repeated 3 times @yashikajotwani12!

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
nemesisdesign: One observation: isn't this series of if/else which checks the size of the value repeated all over the place?
Why don't we turn that into a function that returns a list with two values: the unit and the scale to use in `convertToAdaptiveBytes`?

That should allow to have code like:

``` 
var scales = getAdaptiveScale(value),
    unit = scales[0],
    scale = scales[1];

layout.yaxis = unit;
```

openwisp_monitoring/monitoring/static/monitoring/js/chart.js
```diff
@@ -16,6 +16,128 @@
...
+            }
+            else if (charts[2].y[i] < 1) {
+                convertToAdaptiveBytes(charts, 2, i, 1000, 'MB');
+            }
+            else if (charts[2].y[i] < 1000) {
+                convertToAdaptiveBytes(charts, 2, i, 1, 'GB');
+            }
+            else {
+                convertToAdaptiveBytes(charts, 2, i, 0.001, 'TB');
+            }
```
nemesisdesign, line 87: same here, eg:

```
var scales = getAdaptiveScale(value),
    unit = scales[0],
    scale = scales[1];

convertToAdaptiveBytes(charts, 2, i, scale, unit);
```

nemesisdesign (changes requested):

nemesisdesign (changes requested):
> > I loaded the latest version of this PR on the demo system and in the main dashboard, in the general traffic chart, I see the following:
> > Screenshot from 2022-07-01 12-20-57
> > Shouldn't 0.42 GB be rendered as 420 MB?

I still see this issue in the general traffic chart.

yashikajotwani12:
> 

[ScreenRecorder_2022-07-05_f6a50db7-9930-40d4-8553-01c349afd82f.webm](https://user-images.githubusercontent.com/77020164/177249001-f845f26e-c1d2-440e-9852-2298fe6c6c44.webm)
It is working locally, I have updated the PR for general traffic.

yashikajotwani12:
**when I am only taking hovering into consideration and multiplying that according to the conversion:**

Screenshot from 2022-07-06 23-40-20

nemesisdesign (approved):

nemesisdesign (approved):

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -235,15 +235,14 @@ def _get_access_tech():
                 'title': _('General Traffic'),
                 'label': _('General Traffic'),
                 'description': _(
-                    'Network traffic of the whole network'
-                    ' (total, download, upload) measured in GB.'
+                    'Network traffic of the whole network ' '(total, download, upload).'
```
codesankalp, line 238: Minor change:
```suggestion
                    'Network traffic of the whole network (total, download, upload).'
```

nemesisdesign: Thanks!

atb00ker:
Are the following outdated (checked on demo.openwisp.org) on 20th July or do these errors persist?

fme
dele
lld

PR #409, opened 2022-07-01
[fix] Fixed monitoring test failing due to openwisp-notification 1.0.2

pandafy:
Bug:
In openwisp-notification 1.0.2, the operation for creating notification
settings when an organization is created is wrapped in transaction.on_commit.
TestMonitoringNotifications.test_cpu_metric_threshold_crossed was failing
because of this.

Fix:
Moved this test to a TransactionTestCase class.


codesankalp (approved):

nemesisdesign (approved):
Thanks!

PR #410, opened 2022-07-04
[feature] Charts: allow specifying range of dates #26

### LINKED ISSUE #26: [feature] Charts: allow specifying range of dates
Allow specify by custom date range: the user will be able to select a range of dates and reload the graphs for the dates specified.

First the backend and API must be changed, then the HTML template and JS code.
The grouping of the timeseries must adjust to the length of the range automatically.

The grouping is defined here: https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/models.py#L310

For example, right now, 1 day intervals group results by 10 minutes. While 3 day intervals group results by 20 minutes.

The system could determine the optimum grouping automatically, ranging from 5 minutes (specific days) up to 24 hours (ranges up to 1 year).

Possible steps:

- [ ]  add possibility to filter data by date range in API / model
- [ ] change group_by so that the amount of points shown is constant
- [ ] offer different resolutions (standard, high, maximum), which will reflect on group_by so that the number of points shown (resolution) can be increased if needed

Notes:
- [ ] ensure group by is never lower than default interval
- [ ] make default interval configurable (default to 5 minutes) and change the celery beat setting to use it + update docs

The minimum range is 1 day, the maximum range is 1 year.

yashikajotwani12:
[ScreenRecorder_2022-07-04_b5fd60c1-f402-4f61-bc68-aa57b897f09b.webm](https://user-images.githubusercontent.com/77020164/177163904-1135ac49-91a7-45ff-bcab-9137cd2ff4f4.webm)

Screenshot (89)

Fixes #26

Checks:


yashikajotwani12:
**Update**:

[ScreenRecorder_2022-07-05_a3beda32-34a2-47b0-bab2-8eb11b1f9992.webm](https://user-images.githubusercontent.com/77020164/177336487-a1f50637-f359-4bb2-9ce9-9dc1d98d8bab.webm)

yashikajotwani12:
Here is the flow chart, of the basic functionality of thecustom date range:
https://lucid.app/lucidspark/1106a39d-adc1-40fa-bb43-c4a935419eb2/edit?viewport_loc=-3580%2C-2252%2C7328%2C3792%2C0_0&invitationId=inv_a267a3c5-6241-4d28-8467-22be64dffadf#, it is editable.

nemesisdesign:
> Hey @nemesisdesign! How can we define grouping for a selected number of the date range?
> 
> * We have the functionality currently for the last specific last `3 days` from the current date but if any user selects a 3-day range from any particular month then how we can access that.

I don't understand this question.

> * or if a selection of more than a month and less than a year maybe `7 months` what number of bars we should show and how we can calculate the grouping, any rough idea.

A good start would be: anything more than 28 days (to account for February) can be grouped as 1d.

Anything less than 28 days but more than 3 days should have the same number of points that the weekly timeframe has: right now the weekly time frame is grouped by 1h, which gives around 164 values. Not sure why I see 164 values, if I calculate this I get the 168, but it's not a big deal, here's how I calculate it:

24 hours (1 day) multiplied by 7 days (1 week) divided by 1 (hour)
7 * 24 / 1 = 168

So let's say a user selects a period of 3 weeks:

7 * 24 * 3 = 504
let's divide it by 168:
504 / 168 = 3

So if a user selects 3 weeks we should group by 3 hours in order to get 168 points.

This means the grouping should be dynamic in this situations.

In the future we can see if it makes sense to allow changing it and how to do that.

Let me know if it's clear, if not, please ask questions!

nemesisdesign:
> Here is the flow chart, of the basic functionality of thecustom date range: https://lucid.app/lucidspark/1106a39d-adc1-40fa-bb43-c4a935419eb2/edit?viewport_loc=-3580%2C-2252%2C7328%2C3792%2C0_0&invitationId=inv_a267a3c5-6241-4d28-8467-22be64dffadf#, it is editable.

I did a minor modification to the message displayed to the user in case there's no data to display, the rest looks good.

pandafy (changes requested):
**[deleted]**

yashikajotwani12:
**Working Demo**:
As per the selection of custom daterange from the widget the data is shown on the charts with the x axis denoting the start and the end date.

https://user-images.githubusercontent.com/77020164/180427376-4c4d294a-51a6-4d1e-84b5-a5c4b312f9be.mp4

nemesisdesign (changes requested):
I see progress @yashikajotwani12! :+1: 

When I am testing the widget I noticed two quirks:

- when I first click on date range, it seems that the charts reload, this shouldn't happen, the widget should allow me to select the date range and only then load the data
- after chosing the date range, the current charts are shown briefly, this shouldn't happen, the loading overlay should cover everything while the data loads.

I will continue reviewing and testing in the coming days.

yashikajotwani12:
**1: This is with all red selected ranges:**

https://user-images.githubusercontent.com/77020164/181543624-e0a2c60f-8233-497f-9cfe-925cdfe3b932.mp4

------

**2. This is with light red selected ranges but the start and end point are dark.**

https://user-images.githubusercontent.com/77020164/181543756-85d756bb-60fa-4dfa-91cd-3bd6c036d5e5.mp4

openwisp_monitoring/monitoring/templates/monitoring/paritals/chart.html
```diff
@@ -4,13 +4,14 @@
     <span id="monitoring-timeseries-api-url" data-value="{{ api_url }}"></span>
     <span id="monitoring-timeseries-original-key" data-value="{{ original.key }}"></span>
     <span id="monitoring-timeseries-default-time" data-value="{{ default_time }}"></span>
+    <script type="text/javascript" src="https://cdn.jsdelivr.net/momentjs/latest/moment.min.js"></script>
+    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/daterangepicker/daterangepicker.min.js"></script>
+    <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/daterangepicker/daterangepicker.css" />
```
pandafy, line 9: This is not the correct way to load scripts on the dashboard page because the scripts will get loaded before jQuery is initialised. It appears to me that openwisp-utils does not allow defining additional scripts for the dashboard at the moment. 

For now, you can download these files and serve them from Django itself. `openwisp_monitoring/device/static/monitoring/` would be the directory to store this files. Then you will need to add them in the following code so it will be loaded in the dashboard page. 


https://github.com/openwisp/openwisp-monitoring/blob/76942a31e0d655a246eb846c94434fc53e14795c/openwisp_monitoring/device/apps.py#L266-L281

yashikajotwani12:
The `daterangepicker.min.js`, `moment.min.js`, and `daterangepicker.css` all are included  in `openwisp_monitoring/device/static/monitoring/` but then too the dashboard is having errors, I have followed the procedure as suggested above.

Screenshot from 2022-07-29 18-09-42

yashikajotwani12:
https://user-images.githubusercontent.com/77020164/183014666-2bc70288-a3bd-436a-93ff-99f4dbd79009.mp4

nemesisdesign (changes requested):
When opening `http://localhost:8000/admin/config/device/add/`, after a few secs, the page tries to load the charts (this obviously fails because the device does not exist yet), which shouldn't happen.

nemesisdesign (changes requested):
Good progress @yashikajotwani12!

Here's the problems I found during the last round of testing (using the branch `demo-gsoc22` on demo.openwisp.io, please use the same branch to test locally).

## Hover state color of widget needs improvement

The hovering colour when dates are selected doesn't work well. 

Screenshot from 2022-08-09 18-20-51
Hovering on June 16 when dates are already selected.
In this case I think the hover color should not change but remain red.

Screenshot from 2022-08-09 18-20-59
Hovering on May 11 while selecting, I think the hover color should be of the same color of the rest of the selection.

## Dates of following month when selected look confusing

Screenshot from 2022-08-09 18-20-59

In the image above, if I select the dates of June 1 to 4, both calendar boxes highlight the dates, which is super weird.

Can we configure the date widget to avoid showing the dates of the following month? That is, the calendar of May should not show the dates of June.

Alternatively, if the above cannot be done, we could simply style it so that if a selection is taking place, the color of the disabled dates does not become red, which is kinda working now, but if you click on a specific date like June 1 it does become red and it shouldn't.

## Export button doesn't work for custom range

Try selecting a custom range and export data.

Expected behavior: CSV is downloaded according to selected date
Actual result: API browser opens up with:

```
HTTP 400 Bad RequestAllow:GET, POST, HEAD, OPTIONSContent-Type:application/jsonVary:Accept
[
    "Time range not supported"
]
```

## Start of date ranges or times of each point are buggy

I am comparing the time series data shown in this branch to the time series data shown on other deployed instances and I am finding some inconsistencies in the starting dates or in the date time of each point which are alarming.

### 1 day

I don't notice problems here.

### 3 days

**OpenWISP 22.05:** starts on August 6 at 16:40 and ends on last available point of August 9 (18:40).

**OpenWISP gsoc22-demo branch on demo:** starts on August 6 at 21:40 and ends on last available point of August 9 (18:40).
The starting point is wrong, it should be at least 18:40 or 16:40 as in the stable OpenWISP 22.05. Starting at 21:40 is less than 3 days (72 hours).

### 1 week

**OpenWISP 22.05:** starts on August 3 at 00:00 and ends on last available point of August 9.

**OpenWISP gsoc22-demo branch on demo:** starts on August 3 at 21:00 and ends on last available point of August 9.
The starting point is wrong, it's better to have it start at 00:00.

### 1 month

**OpenWISP 22.05:** Starts on July 10, it doesn't mention any time, the date range starts from 00:00 of July 10 and each data point groups the whole day.
Each bar does not mention the time.

**OpenWISP gsoc22-demo branch on demo:** Starts on July 10 - 16:00, next bar is July 11 - 18:00, next bar is July 12 - 20:00 and so on, each bar increments of 2 hours the starting time, which is wrong.

### 1 year

I don't notice problems here.

### Custom date range

**OpenWISP 22.05:** feature not available.

**OpenWISP gsoc22-demo branch on demo:** I am selecting July 10 to July 16, the starting point is July 10 at 21:20, this is problematic, because if am telling the system that I want to analyze July 10, I mean the whole day, not just the end of the day, here we must start from July 10 at 00:00.
The last point instead is at July 17 at 16:00, which is far out of the range, here we should show until the last available point of July 16, or at max 00:00 of July 17.

yashikajotwani12:
Requested UI Changes:

https://user-images.githubusercontent.com/77020164/183847145-6a537d6c-2be9-4606-8e05-f48954b3bd58.mp4

yashikajotwani12:
> ### Custom date range
> **OpenWISP 22.05:** feature not available.
> 
> **OpenWISP gsoc22-demo branch on demo:** I am selecting July 10 to July 16, the starting point is July 10 at 21:20, this is problematic, because if am telling the system that I want to analyze July 10, I mean the whole day, not just the end of the day, here we must start from July 10 at 00:00. The last point instead is at July 17 at 16:00, which is far out of the range, here we should show until the last available point of July 16, or at max 00:00 of July 17.

**July 10 00:00 - July 16 20:00**

Screenshot from 2022-08-10 20-05-06

atb00ker:
dele

Problem: I see Aug 10 when the date range ends at Aug 6.
1. Open demo instance.
2. Go to the first graph on home page
3. Select range `July 11 - Aug 6`

atb00ker:
dele

I noticed another corner case in the demo instance.
Notice the data is shown *after* the end date.
Pl. check it on `pandafy-archer-c50` in demo instance. 😄 


---

P.S: Just to summarize, 2 + 3 issues are reported on this PR now.
(Pl. do message once you address each one of them!)

atb00ker:
P.S: @yashikajotwani12,  I found something similar to what Fed reported. 

> some start from other times of the day.

I think it might be a timezone related issue.
Somewhere UTC is used and at other places a TZ(like IST in my case) is picked causing shift in time.
You might wanna check that. 
(Let me know if that didn't make sense, I'll explain with an example! 😄 )

yashikajotwani12:
> Another problem I have noticed:
> 
> I select a custom range, eg: June 1 to August 11. When I select August 11, the widget goes back to July (why??).
 
This is its default behavior to show the users the start date they selected as from there they can see their range.

yashikajotwani12:
> dele
> 
> I noticed another corner case in the demo instance. Notice the data is shown _after_ the end date. Pl. check it on `pandafy-archer-c50` in demo instance. smile
 
For me in demo instance, in `pandafy-archer-c50`, it shows from 5 July 00:00 to 29 July 6:00, kind of weird 🤔.

https://user-images.githubusercontent.com/77020164/184279439-1e2f99e1-78f5-467a-980d-5b83efcaae94.mp4

atb00ker:
I have noticed a problem sometimes:

1. Logout on demo instance
2. Log back in.
3. The graph on the dashboard doesn't load. Is it potentially related to this? 🤔 
4. Refresh page and the graph loads!

atb00ker:
> For me in demo instance, in pandafy-archer-c50, it shows from 5 July 00:00 to 29 July 6:00, kind of weird

If you move you mouse a little, it should show 30th as well! 😓

atb00ker:
Either all weird bugs are happening on my browser right now, or this should happen for all.

First load after login on chrome:
vm1

atb00ker:
Now, what is going on here? 
I selected one day and it's giving me 13th and 14th as well?
vm1

openwisp_monitoring/device/admin.py
```diff
@@ -136,9 +136,15 @@ class Media:
             )
```
atb00ker, line 136: Notice range and date: 
vm1

atb00ker (commented):
I was going to comment for larger selections things look fine but then this happened:

yashikajotwani12:
> Now, I selected one day and it's giving me 13th and 14th as well and showing 15th August - 15th August? open_mouth
> 
> vm1

Hey @atb00ker because this chart is for 1 day and it is the same as we see without the daterangepicker widget but on the widget head it shows 15 Aug because it denotes that the chart is of 1 day and the data loaded is the same as that we see without widget.

openwisp_monitoring/device/admin.py
yashikajotwani12: > Notice range and date: vm1

This is because the data was not set in local storage as the previous one and hence after getting the chart on refresh the data is the previous one. Now I have improved this.

yashikajotwani12:
> Either all weird bugs are happening on my browser right now, or this should happen for all.
> 
> First load after login on chrome: vm1

Updated for this condition as well, won't see this now again.

atb00ker:
> Hey @atb00ker because this chart is for 1 day 

Yes, I understand that, but:
1. It's more than 24 hours.
2. 15 Aug - 15 Aug is still wrong, if this is not a side effect of this PR, can you please open a issue to fix that separately! 😄

yashikajotwani12:
> > Hey @atb00ker because this chart is for 1 day
> 
> Yes, I understand that, but:
> 
> 1. It's more than 24 hours.
> 2. 15 Aug - 15 Aug is still wrong, if this is not a side effect of this PR, can you please open a issue to fix that separately! smile

If it's this case, we can change it for 1 day `14 Aug - 15 Aug`. something like this.

yashikajotwani12:
https://user-images.githubusercontent.com/77020164/185428743-20097634-4a93-4124-b142-8c018ee3e77d.mp4

When selecting `1 month` range `30 July ` was partially reachable and when selecting data between `25 July` to `1 Aug` we can see that `30 July ` data are partially reachable.
Export functionality works for selected daterange.
General charts also work for selected daterange.

openwisp_monitoring/db/backends/influxdb/queries.py
```diff
@@ -2,37 +2,38 @@
     'uptime': {
         'influxdb': (
             "SELECT MEAN({field_name})*100 AS uptime FROM {key} WHERE "
-            "time >= '{time}' AND content_type = '{content_type}' AND "
```
atb00ker, line 5: This is on home page,
pl. notice the date range and what it is showing: 

(I think I have pointed this out before, are you having difficulties in pin pointing the reason for the problem?)

dev-01

openwisp_monitoring/db/backends/influxdb/queries.py
```diff
@@ -2,37 +2,38 @@
...
+            "time >= '{time}' AND time < '{time}' + {timespan} AND content_type = '{content_type}' AND "
             "object_id = '{object_id}' GROUP BY time(1d)"
```
atb00ker, line 6: ~Selecting single date on home page (`https://demo.openwisp.io/admin/`) simply breaks:~

Wait, it's not just single date, the thing worked once and now the home page custom range is simply broken! :-O

dev-01

atb00ker (commented):
Hey @yashikajotwani12 

The demo instance seems to be still completely broken, the custom date change works once then it doesn't!

Is this problem solved already, can we deploy again and test or are you stuck with it?

openwisp_monitoring/db/backends/influxdb/queries.py
yashikajotwani12: The demo system is not updated with the latest commit ig. I have to make sure the build pass and then we can have that on the demo. :+1:

openwisp_monitoring/db/backends/influxdb/queries.py
yashikajotwani12: Hey @atb00ker, now the general charts work fine in my last pushed commit, and the demo system is not updated with that, I will make sure the build pass asap.  And then we can test if the things work or not.

nemesisdesign (changes requested):
Please see my comments below.

## Other problems found during testing

### Date range not showing the selected date range after reload

After selecting a custom date range, if I reload the page and I open the date picker widget, this is what I see:
Screenshot from 2022-08-24 16-58-16

The current date is selected,  but the currently selected date range should be selected and shown instead.

### Selecting another custom date range does not work

Try selecting a custom date range.
Now, after the new data loads, try selecting a new custom date range, it's not working.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -523,7 +525,9 @@ def get_top_fields(self, number):
 
     def _get_query_params(self, time):
         m = self.metric
-        params = dict(field_name=m.field_name, key=m.key, time=self._get_time(time))
+        params = dict(
+            field_name=m.field_name, key=m.key, time=self._get_time(time), timespan=time
+        )
```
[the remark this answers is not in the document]
yashikajotwani12: Okay! Would try that way.

openwisp_monitoring/views.py
```diff
@@ -32,7 +36,46 @@ def _get_additional_data(request, *args, **kwargs):
...
+            )
+            daterange = int(daterange)
+            if daterange > 0 and daterange < 3:
+                value = '10m'
+            elif daterange >= 3 and daterange < 7:
+                value = str(round((daterange / 3) * 20)) + 'm'
+            elif daterange >= 7 and daterange < 28:
+                value = str(round((daterange / 7) * 1)) + 'h'
+            elif daterange >= 28 and daterange < 365:
+                value = str(round((daterange / 28) * 24)) + 'h'
```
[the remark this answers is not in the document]
yashikajotwani12: Yes, I would.

nemesisdesign (changes requested):
I am not seeing progress.

When I try to load charts, I get the following error:
```
Internal Server Error: /api/v1/monitoring/dashboard/
  File "/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/openwisp-monitoring/openwisp_monitoring/views.py", line 94, in get
    data = self._get_charts_data(charts, time, timezone)
  File "/openwisp-monitoring/openwisp_monitoring/views.py", line 116, in _get_charts_data
    chart_dict = chart.read(
  File "/openwisp-monitoring/openwisp_monitoring/monitoring/base/models.py", line 622, in read
    data_query = self.get_query(**query_kwargs)
  File "/openwisp-monitoring/openwisp_monitoring/monitoring/base/models.py", line 504, in get_query
    params = self._get_query_params(time)
  File "/openwisp-monitoring/openwisp_monitoring/monitoring/base/models.py", line 532, in _get_query_params
    end_date=self.get_end_date(time),
  File "/openwisp-monitoring/openwisp_monitoring/monitoring/base/models.py", line 589, in get_end_date
    cls.END_DATE = cls.END_DATE.replace(microsecond=0, tzinfo=None)
TypeError: 'microsecond' is an invalid keyword argument for replace()
````

When opening the first page of the admin I get the following errors In the JS console:

```
not available
﻿
daterangepicker.js:1565 Uncaught TypeError: Cannot read properties of undefined (reading 'fn')
    at daterangepicker.js:1565:7
    at daterangepicker.js:29:32
    at daterangepicker.js:9:1
```

Most of the issues highlighted in my previous review are still standing.

nemesisdesign:
Superseeded by #441. Thanks for your effort.

PR #427, opened 2022-08-09
[feature] Added alert on metric fields #406

### LINKED ISSUE #406: [feature] Possibility to add alerts for related_fields
- If we look at current implementation of alerts, We see that alerts checks thresholds on only `value/field_name`, it would be nice if we can add alert functionality for `extra_values/related_fields` as well.
https://github.com/openwisp/openwisp-monitoring/blob/6119ac0dfa7221e7e06d88fdcd628ec0d629a3a7/openwisp_monitoring/monitoring/base/models.py#L315-L358

Aryamanz29:
- This change will add possibility to check `alert threshold` on the specific metric field if provided in configuration otherwise, it uses `field_name`, which is current alert behaviour.

- Removed `alert_field` from the iperf metric configuration.

https://user-images.githubusercontent.com/56113566/178989966-f007a9b9-c9fd-4877-ac01-539e38b3331f.mp4

Closes #406 

Checks:


pandafy (commented):
@Aryamanz29 I have requested some minor changes below

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -355,6 +355,21 @@ def write(
                 'send_alert': send_alert,
             }
             options['metric_pk'] = self.pk
+
+        # check if alert_on_related_field is present in metric configuration
+        if 'alert_on_related_field' in self.config_dict:
+            related_field = self.config_dict['alert_on_related_field']
+            if not extra_values:
+                raise ValueError(
+                    'write() missing positional argument: "extra_values" required for alert on related field'
```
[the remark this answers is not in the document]
Aryamanz29: my bad 😅

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -355,6 +355,21 @@ def write(
...
+                )
+            if related_field not in extra_values.keys():
+                raise ValueError(
+                    f'"{key}" is not defined for alert_on_related_field in metric configuration'
+                )
+            options['check_threshold_kwargs'].update(
+                {'value': extra_values[related_field]}
+            )
```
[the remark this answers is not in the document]
Aryamanz29: Yes, you're right 👍 I'll update that

nemesisdesign (commented):
One question? Why do we need this functionality?

What do we need to alert?

I am scared by this code, I do not understand it:

- I do not understand what specific need is pushing us to add it
- I do not understand what we want to achieve and how

I see a bunch of lines being added in one of the most delicate methods of OpenWISP, **I am really scared**.

README.rst
```diff
@@ -1965,6 +1965,13 @@ The ``AlertSettings`` of ``ping`` metric will by default use ``threshold`` and `
 defined in the ``alert_settings`` key.
 You can always override them and define your own custom values via the *admin*.
 
+You can also use the ``alert_on_related_field`` key in metric configuration
+which allows ``AlertSettings`` to use the value specified in ``related_field``to check the ``threshold``, instead
```
nemesisdesign, line 1969:
```suggestion
which allows ``AlertSettings`` to use the value specified in ``related_field`` to check the ``threshold``, instead
```

README.rst
```diff
@@ -1965,6 +1965,13 @@ The ``AlertSettings`` of ``ping`` metric will by default use ``threshold`` and `
...
+of the default ``field_name`` key. A real world example of this can be seen in the
+`Iperf metric configuration
+<https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/configuration.py#L546-L700>`_,
+where we used ``jitter`` (``related_field``) for alerts.
```
nemesisdesign, line 1973: I am not sure whether we should have this enabled by default in the code.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -355,6 +355,21 @@ def write(
                 'send_alert': send_alert,
             }
             options['metric_pk'] = self.pk
+
+            # check if alert_on_related_field is present in metric configuration
+            if 'alert_on_related_field' in self.config_dict:
```
nemesisdesign, line 360: maybe we can call this `alert_field`, then we can define that:
if `alert_field` is not supplied, the system assumes it's "field_name".

I would then create a method for this property, which behaves as described above, eg:

```python
@property
def alert_field(self):
    return self.config_dict.get('alert_field', self.field_name)

@property
def alert_on_related_field(self):
    return self.alert_field != self.field_name
```

```suggestion
            if alert_on_related_field:
                # etc.
```

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -355,6 +355,21 @@ def write(
...
+                related_field = self.config_dict['alert_on_related_field']
+                if not extra_values:
+                    raise ValueError(
+                        'write() missing keyword argument: "extra_values" required for alert on related field'
+                    )
+                if related_field not in extra_values.keys():
+                    raise ValueError(
+                        f'"{key}" is not defined for alert_on_related_field in metric configuration'
+                    )
+                options['check_threshold_kwargs'].update(
```
nemesisdesign, line 370: add comment hinting that this will cause the check of the threshold on the related field

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -771,7 +792,15 @@ def _is_crossed_by(self, current_value, time=None, retention_policy=None):
...
-                results.append(self._value_crossed(point[self.metric.field_name]))
+                # check if alert_on_related_field is present in metric configuration
+                if 'alert_on_related_field' in self.metric.config_dict:
+                    results.append(
+                        self._value_crossed(
+                            point[self.metric.config_dict['alert_on_related_field']]
+                        )
+                    )
+                else:
+                    results.append(self._value_crossed(point[self.metric.field_name]))
```
nemesisdesign, line 803:
```suggestion
                results.append(self._value_crossed(point[self.metric.alert_field]))
```

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -753,13 +768,19 @@ def _is_crossed_by(self, current_value, time=None, retention_policy=None):
             return value_crossed
         # tolerance is set, we must go back in time
         # to ensure the threshold is trepassed for enough time
+        # check if alert_on_related_field is present in metric configuration
+        if 'alert_on_related_field' in self.metric.config_dict:
+            alert_on_related_field = [self.metric.config_dict['alert_on_related_field']]
+        else:
+            alert_on_related_field = []
```
nemesisdesign, line 775:
```suggestion
        # if alert field is supplied, retrieve such field when reading
        # so that we can let the system calculate the threshold on it
        extra_values = []
        if self.metric.alert_on_related_field:
           extra_values = [self.metric.alert_field]
```

nemesisdesign (changes requested):
Let's make very clear in the commit message why are we changing this.

README.rst
```diff
@@ -1965,6 +1965,77 @@ The ``AlertSettings`` of ``ping`` metric will by default use ``threshold`` and `
...
+                    ),
+                    'message': _(
+                        'The device [{notification.target}]({notification.target_link}) '
+                        '{notification.verb}.'
+                    ),
+                },
+            },
+            # iperf charts configuration here
+        },
+    }
```
nemesisdesign, line 2028: is it really necessary to redefine all this?

Wouldn't the following be enough?

```python
# Main project settings.py

DEFAULT_METRICS = {
    'iperf': {
        'alert_field': 'sent_bps_tcp',
        # Alert is sent immediately (tolerance 0) when bandwidth less than 10 Mbps
        'alert_settings': {'operator': '<', 'threshold': 10000000, 'tolerance': 2880},
    },
    'notification': {
        'problem': {
            'verbose_name': 'Bandwidth PROBLEM',
            'verb': _('Iperf bandwidth is less than normal value'),
        },
        'recovery': {
            'verbose_name': 'Bandwidth RECOVERY',
            'verb': _('Iperf bandwidth now back to normal'),
        },
    },
}
```

Please double check this and test the pseudo code above.

nemesisdesign (changes requested):
Good progress @Aryamanz29, see my comments below.

README.rst
Aryamanz29: We need to create `AlertSettings` for `iperf check` first, then we can add the above configuration to `settings.py`.

README.rst
```diff
@@ -1965,6 +1965,77 @@ The ``AlertSettings`` of ``ping`` metric will by default use ``threshold`` and `
 defined in the ``alert_settings`` key.
 You can always override them and define your own custom values via the *admin*.
 
+You can also use the ``alert_field`` key in metric configuration
+which allows ``AlertSettings`` to check the ``threshold`` on
+``alert_field`` instead of the default ``field_name`` key.
+
+In the following example, we added ``alert_field`` to the metric configuration
+for the iperf metric in order to check the threshold on **iperf bandwidth**.
```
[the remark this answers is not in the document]
Aryamanz29: I'll take care of this next time :+1:

README.rst
```diff
@@ -1965,6 +1965,77 @@ The ``AlertSettings`` of ``ping`` metric will by default use ``threshold`` and `
...
+                'sent_bps_udp',
+                'jitter',
+                'total_packets',
+                'lost_packets',
+                'lost_percent',
+            ],
+            # Check alert threshold on iperf bandwidth (TCP)
+            'alert_field': 'sent_bps_tcp',
+            # Alert is sent immediately (tolerance 0) when bandwidth less than 10 Mbps
+            'alert_settings': {'operator': '<', 'threshold': 10000000, 'tolerance': 0},
```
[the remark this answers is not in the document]
Aryamanz29: ok, I'll that :+1:

README.rst
```diff
@@ -1969,79 +1969,69 @@ You can also use the ``alert_field`` key in metric configuration
...
-        },
-    }
+   # Main project settings.py
+   from django.utils.translation import gettext_lazy as _
+
+   OPENWISP_MONITORING_METRICS = {
+       'iperf': {
+           # Check alert threshold on iperf bandwidth (TCP)
+           'alert_field': 'sent_bps_tcp',
+           'notification': {
```
nemesisdesign, line 2000: isn't `alert_settings` missing here?

README.rst
```diff
@@ -1969,79 +1969,69 @@ You can also use the ``alert_field`` key in metric configuration
...
+The following example shows how to use the
+`setting <https://github.com/openwisp/openwisp-monitoring#openwisp_monitoring_metrics>`_
+to reconfigure the system for `iperf check <#iperf-1>`_ to send an alert if
+the measured **TCP bandwidth** has been less than **10 Mbit/s** for more than **2 days**.
+
+1. Iperf check doesn't have any ``AlertSettings`` by default,
+but we can easily create one through the OpenWISP admin interface.
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/alert_field_setttings.png
+  :align: center
```
nemesisdesign, line 1987: Let's not recommend it this way but rather recommend the usage of the feature you're introducing in https://github.com/openwisp/openwisp-monitoring/pull/431 which will be accessible to most users and should also be easier to use.
Please leave it pending once https://github.com/openwisp/openwisp-monitoring/pull/431 is merged, then you can update this branch accordingly.

This could become a GIF.

README.rst
```diff
@@ -1969,79 +1969,69 @@ You can also use the ``alert_field`` key in metric configuration
...
+If you don't need to register a new metric but need to change a specific key of an
+existing metric configuration, you can use `OPENWISP_MONITORING_METRICS <#openwisp_monitoring_metrics>`_.
+
+The following example shows how to use the
+`setting <https://github.com/openwisp/openwisp-monitoring#openwisp_monitoring_metrics>`_
+to reconfigure the system for `iperf check <#iperf-1>`_ to send an alert if
+the measured **TCP bandwidth** has been less than **10 Mbit/s** for more than **2 days**.
+
+1. Iperf check doesn't have any ``AlertSettings`` by default,
+but we can easily create one through the OpenWISP admin interface.
```
nemesisdesign, line 1984: 1. By deafult, Iperf checks do not have any alert settings defined,
but it is easy to create one through the admin interface as shown below.

nemesisdesign (changes requested):

openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
```diff
@@ -402,6 +402,118 @@ def test_alerts_disabled(self):
...
+            m.write(10, extra_values={'test_related_2': 40})
+            m.refresh_from_db()
+            self.assertEqual(m.is_healthy, False)
+            self.assertEqual(m.is_healthy_tolerant, False)
+            self.assertEqual(Notification.objects.count(), 1)
+            n = notification_queryset.first()
+            self.assertEqual(n.recipient, admin)
+            self.assertEqual(n.actor, m)
+            self.assertEqual(n.action_object, m.alertsettings)
+            self.assertEqual(n.level, 'warning')
```
pandafy, line 460: This test is not creating any new notification because of the double alarm prevention mechanism. This tests looks bogus. 

I suggest deleting notifications after every sub-test. This will ensure that only one notification will be created per testing environment and we will avoid writing any bogus tests in the future as well.

README.rst
```diff
@@ -1023,16 +1023,8 @@ jitter, datagram loss etc of the device using `iperf3 utility <https://iperf.fr/
...
-.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/device-inline-check.png
-  :align: center
-
-**Note:** To access the above page, the user must have some special permissions,
-which are included by default in the groups "Administrator" and "Operator" and are shown in the screenshot below.
-
-.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/inline-permissions.png
-  :align: center
+You can also `add the iperf check and it's alert settings
+<#add-checks-and-alert-settings-from-the-device-page>`_ directly from the device page.
```
pandafy, line 1027: I think we should also add a note that even when `OPENWISP_MONITORING_AUTO_IPERF` is set to True, users need to update the metric configuration to enable alerts for the iperf check.

README.rst
```diff
@@ -1250,6 +1242,79 @@ in the settings:
...
+               },
+           },
+       },
+   }
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/alert_field_warn.png
+  :align: center
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/alert_field_info.png
+  :align: center
```
pandafy, line 1310: I think it would easier to read the text in the image if the image width was lesser. In my experience, 800px wide images looks good on GitHub REAMDE. Please try cropping the image and check if it makes the image more useful.

pandafy (changes requested):
@Aryamanz29 I left some suggesting for improving the README and tests. I think after that we can merge this PR.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -205,6 +221,14 @@ def content_type_key(self):
         except AttributeError:
             return None
 
+    @property
+    def alert_field(self):
+        return self.config_dict.get('alert_field', self.field_name)
```
pandafy, line 226: If the `alert_field` is configured for the metric in the project's setting as shown below: 

```python
OPENWISP_MONITORING_METRICS = {
    'iperf': {
        'alert_field': 'iperf_result',
    }
}
```

Then, changing the alert field from the AlertInline on the device change admin will have no effect. The "Field name" in the AlertInline represents `metric.field_name`. 

I think we can solve this problem by modifying this code as follows

```python
    @property
    def alert_field(self):
        if self.field_name != self.config_dict['field_name']:
            return self.field_name
        return self.config_dict.get('alert_field', self.field_name)
```

openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
```diff
@@ -402,6 +402,132 @@ def test_alerts_disabled(self):
...
+            n.delete()
```
pandafy, line 453: If this sub-test fails, then the notification will not get deleted. And, this would affect the following sub-tests. We should always perform operations related to resetting of the testing environment outside of the sub-tests.

```suggestion
        Notification.objects.all().delete()
```

Same applies for other occurrences.

pandafy (changes requested):

openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
Aryamanz29: Noted!

openwisp_monitoring/monitoring/base/models.py
Aryamanz29: Will update it, thanks!

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -223,6 +223,8 @@ def content_type_key(self):
 
     @property
     def alert_field(self):
+        if self.field_name != self.config_dict['field_name']:
+            return self.field_name
```
pandafy, line 227: Is there a dedicated test for this condition? When I comment these lines, the test suite passes. But, I expect atleast one test to fail.

Aryamanz29: I'll write one 👍🏻

README.rst
```diff
@@ -1250,6 +1246,77 @@ in the settings:
...
 
+Adding Checks and Alert settings from the device page
+-----------------------------------------------------
+
+We can add checks and define alert settings directly from the **device page**.
+
+To add a check, you just need to select an available **check type** as shown below:
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/device-inline-check.gif
+  :align: center
```
nemesisdesign, line 1257: this GIF is 1 MB, can we find a way to be less heavy? Either optmize the GIF or use one or more screenshots?

README.rst
```diff
@@ -1250,6 +1246,77 @@ in the settings:
...
+The following example shows how to use the
+`OPENWISP_MONITORING_METRICS setting <#openwisp_monitoring_metrics>`_
+to reconfigure the system for `iperf check <#iperf-1>`_ to send an alert if
+the measured **TCP bandwidth** has been less than **10 Mbit/s** for more than **2 days**.
+
+1. By default `Iperf checks <#iperf-1>`_ do not have any alert settings defined,
+but it is easy to create one through the device page as shown below:
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/device-inline-alertsettings.gif
+  :align: center
```
nemesisdesign, line 1268: this GIF is 6 MB, can we find a way to be less heavy? Either optmize the GIF or use one or more screenshots?

nemesisdesign (changes requested):
All good, my remaining concern is only the GIF size which is adding +7MB to the README, see below.

Aryamanz29:
@nemesisdesign Should I upload those GIFs to `docs` after compressing them to **5.2 MB (1.2 + 4)** from **9.1MB (2 + 7.1)**?

README.rst
Aryamanz29: I'm unable to compress those GIFs to less than **5MB**, so I'm replacing them with screenshots.

README.rst
Aryamanz29: I'm unable to compress those GIFs to less than **5MB**, so I'm replacing them with screenshots.

nemesisdesign (approved):

PR #428, opened 2022-08-09
[feature] Added support for configuring more parameters for iperf check

### LINKED ISSUE #416: [feature] Add support for configuring more parameters for iperf3 checks
The `iperf3` programs support different options. Supporting all those options in OpenWISP will unnecessarily bloat the iperf3 check. Instead, we should start with the following options and we can add more after receiving feedback from users. 

- [x] `--connect-timeout`
- [x] `--bitrate`
- [x] `--bytes`
- [x] `--blockcount`
- [x] `--length`
- [x] `--parallel`
- [x] `--reverse`
- [x] `--bidir`
- [x] `--window`

The default values for these options should follow the defaults set in iperf3

Aryamanz29:
Closes #416

Checks:


openwisp_monitoring/check/classes/iperf.py
```diff
@@ -115,21 +186,71 @@ def check(self, store=True):
...
+        # use it instead of 'time' or 'bytes'
+        if blockcount:
+            test_end_condition = f'-k {blockcount}'
+        # only one reverse condition can be use
+        # reverse or bidirectional not both
+        rev_or_bidir = ''
+        if reverse:
+            rev_or_bidir = '--reverse'
+        if bidirectional:
+            rev_or_bidir = '--bidir'
```
pandafy, line 253: I think we should move this to a separate method which returns the two commands.

pandafy (changes requested):
Good work @Aryamanz29! We will also need to update the README to point out the options available for configuration. 

I did a manual testing for all the flags.

openwisp_monitoring/check/classes/iperf.py
Aryamanz29: yes, thanks!

README.rst
```diff
@@ -1029,6 +1029,58 @@ You can also change the parameters used for iperf checks (e.g. timing, port, use
 password, rsa_publc_key etc) using the `OPENWISP_MONITORING_IPERF_CHECK_CONFIG
 <#OPENWISP_MONITORING_IPERF_CHECK_CONFIG>`_ setting.
 
+Iperf Check Parameters
+######################
+
+Currently, iperf check supports the following parameters:
```
pandafy, line 1035: I would move this to the end of the "Iperf" section. I would add links to this in the following sections of the documentation: 

1. ["3. Set up and configure Iperf server settings"](https://github.com/openwisp/openwisp-monitoring/blob/0ae7702d3cd0ddea77c7a075046697f644f9ba07/README.rst#3-set-up-and-configure-iperf-server-settings)
2. ["OPENWISP_MONITORING_IPERF_CHECK_CONFIG"](https://github.com/openwisp/openwisp-monitoring/blob/0ae7702d3cd0ddea77c7a075046697f644f9ba07/README.rst#openwisp_monitoring_iperf_check_config)

README.rst
```diff
@@ -1029,6 +1029,58 @@ You can also change the parameters used for iperf checks (e.g. timing, port, use
...
+| password        | str      | ''                |
++-----------------+----------+-------------------+
+| rsa_public_key  | str      | ''                |
++-----------------+----------+-------------------+
+| client_options  | dict     | {}                |
++-----------------+----------+-------------------+
+| port            | int      | 5201              |
++-----------------+----------+-------------------+
+| time            | int      | 10                |
++-----------------+----------+-------------------+
```
pandafy, line 1053: ⚠️ Before spending time on this, let's double check with @nemesisdesign. 

I would restructure this table to show that other parameters should be defined under `client_options` 

```rst
+--------------------+-------------------------------------+
| ``client_options`` | +----------+---------+------------+ |
|                    | | ``port`` | ``int`` | ``5201``   | |
|                    | +----------+---------+------------+ |
|                    | | ``time`` | ``int`` | ``10``     | |
|                    | +----------+---------+------------+ |
+--------------------+-------------------------------------+
```

openwisp_monitoring/check/classes/iperf.py
```diff
@@ -320,6 +287,46 @@ def check(self, store=True):
         device_connection.disconnect()
         return result
 
+    def _get_iperf_test_conditions(self):
+        """
+        Returns iperf check test conditions (rev_or_bidir, end_condition)
+        """
```
pandafy, line 293: This is better, but I meant to add a method which returns the commands

```python
def _get_check_command(self):
    # build tcp and udp commands
    return tcp_command, udp_command
```    


Then from `check` method, we would call this method

```python
def check(self):
     tcp_command, udp_command = self._get_check_command()
     # perform check
```

This will make the code of `check` method easy to follow.

pandafy (changes requested):
This PR is functional. but it requires more work before merging. We must ensure that we are properly guiding a user in the documentation to the right part instead of just putting it out information for them to hunt.

openwisp_monitoring/check/classes/iperf.py
Aryamanz29: Sure, I'll update that :+1:

README.rst
Aryamanz29: I also think a nested table will fit better for `client_options`.

Aryamanz29: Screenshot from 2022-08-21 20-31-27

README.rst
Aryamanz29: I'll update that :+1:  thanks!

README.rst
nemesisdesign: looks good!

README.rst
```diff
@@ -1127,6 +1127,66 @@ After that, you should see the iperf network measurements charts.
 .. image:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/iperf-charts.png
   :alt: Iperf network measurement charts
 
+5. Iperf check parameters
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
pandafy, line 1131: I don't think this needs to be numbered heading.

pandafy (commented):
Good work Aryaman! 👏🏼 We need to do some minor changes and after that we can merge this.

pandafy (changes requested):
Please see https://github.com/openwisp/openwisp-monitoring/pull/428#pullrequestreview-1081562916

README.rst
```diff
@@ -1127,6 +1127,66 @@ After that, you should see the iperf network measurements charts.
...
+| **Parameter**         | **Type** | **Default Value**                                                  |
++-----------------------+----------+--------------------------------------------------------------------+
+|``host``               | ``list`` | ``[]``                                                             |
++-----------------------+----------+--------------------------------------------------------------------+
+|``username``           | ``str``  | ``''``                                                             |
++-----------------------+----------+--------------------------------------------------------------------+
+|``password``           | ``str``  | ``''``                                                             |
++-----------------------+----------+--------------------------------------------------------------------+
+|``rsa_public_key``     | ``str``  | ``''``                                                             |
++-----------------------+----------+--------------------------------------------------------------------+
```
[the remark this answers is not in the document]
Aryamanz29: It's not required, I'll remove that :+1:

pandafy (approved):
LGTM 👍🏼 

Deferring merge to @nemesisdesign

pandafy (approved):
I have test this PR again today and it works as expected. It should be good to merge. 👏🏼

README.rst
```diff
@@ -1134,6 +1134,64 @@ After that, you should see the iperf network measurements charts.
 .. image:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/iperf-charts.png
   :alt: Iperf network measurement charts
 
+Iperf check parameters
```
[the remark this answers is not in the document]
Aryamanz29: I'll update that :+1:

PR #429, opened 2022-08-11
[change] Timeseries data is deleted when device is deleted #419

### LINKED ISSUE #419: [change] Delete tsdb data when device is deleted
Delete tsdb data when device is deleted.

pandafy:
Closes #419

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -33,7 +33,7 @@ from ..configuration import (
 )
 from ..exceptions import InvalidChartConfigException, InvalidMetricConfigException
 from ..signals import pre_metric_write, threshold_crossed
-from ..tasks import timeseries_write
+from ..tasks import delete_device_tsdb_data, timeseries_write
 
 User = get_user_model()
 logger = logging.getLogger(__name__)
@@ -117,7 +117,7 @@ class AbstractMetric(TimeStampedEditableModel):
 
     @classmethod
     def post_delete_receiver(cls, instance, *args, **kwargs):
-        timeseries_db.delete_metric_data(key=instance.key, tags=instance.tags)
+        delete_device_tsdb_data.delay(instance.key, instance.tags)
 
     @classmethod
     def _get_or_create(cls, **kwargs):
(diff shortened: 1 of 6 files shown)


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -115,6 +115,10 @@ def full_clean(self, *args, **kwargs):
         self.key = self._makekey(self.key)
         return super().full_clean(*args, **kwargs)
 
+    @classmethod
+    def post_delete_receiver(cls, instance, *args, **kwargs):
+        timeseries_db.delete_metric_data(key=instance.key, tags=instance.tags)
```
nemesisdesign, line 120: The problem with this is that this may fail, moreover if multiple devices are deleted, the likelyhood of failure is greater.

We need to execute this in a background task with some retry mechanism.
It seems to me that this particular method `delete_metric_data`, does not have a built-in retry mechanism, therefore we could simply let the celery task raise the exception on failures and let the celery task retry whenever the exception is raised, as we do for writing the data.

To sum up, I believe deleting data should work in a similar manner to how the data is written, I think we can simply replicate that kind of structure (executed in a dedicated celery task, which has the retry on failure argument specified).

What do you think?

nemesisdesign (changes requested):

openwisp_monitoring/monitoring/base/models.py
pandafy: I have added a celery task for deletion of the data. The celery task does not have retry configured because the underlying method already has a retry mechanism. 

https://github.com/openwisp/openwisp-monitoring/blob/e563520c7765d4e638f8e65b8b8e4095fa2fc4cc/openwisp_monitoring/db/backends/influxdb/client.py#L194-L196

openwisp_monitoring/monitoring/tasks.py
```diff
@@ -36,6 +36,11 @@ def timeseries_write(
         post_metric_write.send(**signal_kwargs)
 
 
+@shared_task
+def delete_device_tsdb_data(key, tags):
```
nemesisdesign, line 40: Why are you splitting some code in `openwisp_monitoring.device` and some other lines in `openwisp_monitoring.monitoring`?

It could make sense to have a generic way to remove timeseries data asynchronously, I like the idea, but this task shouldn't have `device` in its name, I'd rename it simply `timeseries_delete`, which is also more consistent with the naming used in the method above.

nemesisdesign (changes requested):
Good progress, see below.

nemesisdesign (approved):

PR #430, opened 2022-08-12
[fix] Fixed traffic charts showing wrong traffic total #415

### LINKED ISSUE #415: [bug] Traffic total sometimes is very wrong
This is happening in the current development version (1.1.0 alpha).

Screenshot from 2022-07-15 17-05-23

The total amount in some cases does not match the download + upload exactly, but it should.

pandafy:
The traffic charts were showing the wrong total because of the
approximation (round-off) done in the backend to decrease the
the precision of values to two decimal places.

Since the adaptive feature already does this approximation
in the frontend, the rounding off is skipped from the backend.

Closes #415

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -1,4 +1,3 @@
-import decimal
 import json
 import logging
 from collections import OrderedDict
@@ -38,7 +37,6 @@ from ..tasks import timeseries_write
 
 User = get_user_model()
 logger = logging.getLogger(__name__)
-adaptive_chart_decimal_context = decimal.Context(prec=3, rounding=decimal.ROUND_DOWN)
 
 
 class AbstractMetric(TimeStampedEditableModel):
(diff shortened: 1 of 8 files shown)


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -586,7 +586,14 @@ def read(
                 if key == 'time':
                     continue
                 traces.setdefault(key, [])
-                if decimal_places and isinstance(value, (int, float)):
+                if (
+                    # The adaptive feature does unit conversion and approximation
+                    # in the frontend, hence rounding off is skipped here.
+                    # TODO: Update this code to check for 'adaptive_prefix'
+                    not self.config_dict.get('unit', '').startswith('adaptive')
```
nemesisdesign, line 593: why this TODO here?

nemesisdesign (commented):
Looks good, see my question below.

openwisp_monitoring/monitoring/base/models.py
pandafy: In `gsoc22-demo` PR, Aryaman has changed `adaptive_bytes` to `adaptive_prefix`. The current code with work with both `adaptive_bytes` and `adaptive_prefix`. But, later on we might choose to use `adaptive_prefix` only.

nemesisdesign: ok clear now, thanks! We shall create an issue for this to remind ourselves.

nemesisdesign (approved):
Looks good! Let's keep testing it a bit more.

PR #431, opened 2022-08-12
[change] Added checks & alertsettings permission for device tab #424

### LINKED ISSUE #424: [change] Add additional permission classes for accessing "Checks" and "Alert Settings" tabs on device page
We need to add additional permission objects (view, add, change, delete) to only provide access to "Checks" and "Alert Settings" tab on the device change page when the user does not have access to general pages for "Checks" and "Alert Settings". 

We will have to override the `has_<operation>_permission` methods of the InlineAdmin classes to check for the new permission object that we will create.

### LINKED ISSUE #426: [change] Allow users to add checks
If users have the right permissions (#424), allow them to be able to add checks.

This is useful when a check is not created automatically and we want to create it for specific devices only.

Aryamanz29:
Screenshot from 2022-08-12 16-49-25

Screenshot from 2022-08-12 16-48-48

Screenshot from 2022-08-12 16-49-10

Closes #424

Closes #426

Checks:


openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +63,32 @@ class CheckInline(GenericStackedInline):
     fields = ['check_type', 'is_active']
     readonly_fields = ['check_type']
 
+    def _set_check_inline(self):
+        self.fields = []
+        self.readonly_fields = ['created', 'modified']
```
nemesisdesign, line 68: what is this doing? I don't understand.

Aryamanz29: `CheckInline` (with current fields ['check_type', 'is_active'])

Screenshot from 2022-08-14 15-36-00

`CheckInline` (with all fields [])

I think this will be more useful for the users because here they have more options to configure e.g. `checks params, name` etc & also it's similar to general check page, Although we can remove `description` field if required.
 
Screenshot from 2022-08-14 15-39-41

nemesisdesign: why do you set it this way instead of just setting it at class level? I can't make sense of it.

Aryamanz29: I do this way because we only want to show these fields when the user has `check_inline` permissions if we set those at the class level then those form fields are also visible for `check model` permissions that have already access to `check general page`.

openwisp_monitoring/monitoring/migrations/0010_add_group_permissions.py
```diff
@@ -8,7 +8,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('device_monitoring', '0004_wificlient_wifisession'),
+        ('monitoring', '0009_alter_alertsettings_options'),
```
nemesisdesign, line 11: why this renaming here?

openwisp_monitoring/device/migrations/__init__.py
```diff
@@ -90,36 +86,3 @@
...
-            permission = Permission.objects.get(codename='view_{}'.format(model_name))
-            operator.permissions.add(permission.pk)
-        except Permission.DoesNotExist:
-            pass
-        for operation in manage_operations:
-            admin.permissions.add(
-                Permission.objects.get(
-                    codename='{}_{}'.format(operation, model_name)
-                ).pk
-            )
```
nemesisdesign, line 125: why is this being moved?

openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +64,38 @@ class CheckInline(GenericStackedInline):
     fields = ['check_type', 'is_active']
     readonly_fields = ['check_type']
 
+    @property
+    def check_app(self):
+        if 'SAMPLE_APP' in environ:
+            return 'sample_check'
+        return 'check'
+
```
pandafy, line 72: This is wrong Aryaman. The application code should not know anything about `SAMPLE_APP`. We use sample app to test the extensibility of OpenWISP modules. Please update your test to check for correct label.

openwisp_monitoring/monitoring/migrations/__init__.py
```diff
@@ -26,15 +28,22 @@ def assign_permissions_to_groups(apps, schema_editor):
...
-                    codename='{}_{}'.format(operation, model_name)
-                ).pk
+            permission = Permission.objects.get(
+                codename='{}_{}'.format(operation, model_name)
             )
+            if model_name in ['check', 'alertsettings']:
+                permission = Permission.objects.get(
+                    codename='{}_{}_inline'.format(operation, model_name)
+                )
+            admin.permissions.add(permission.pk)
```
nemesisdesign, line 46: I think it would be simpler and clearer to add new data migration code instead of editing the existing data migrations.
Moving around logic which deals with data migrations is error prone and not worth the effort at the moment.

openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +64,38 @@ class CheckInline(GenericStackedInline):
...
+    def _set_check_inline(self):
+        self.fields = []
+        self.readonly_fields = ['created', 'modified']
+
     def has_add_permission(self, request, obj=None):
+        if request.user.has_perm(f'{self.check_app}.add_check_inline'):
```
pandafy, line 78: You should be able to get the `model_name` like this: 

```suggestion
        if request.user.has_perm(f'{self.model._meta.model_name}.add_check_inline'):
```

openwisp_monitoring/monitoring/migrations/0010_add_group_permissions.py
Aryamanz29: Because I have moved `add_group_permissions` migration to `monitoring app`.

openwisp_monitoring/device/migrations/__init__.py
Aryamanz29: `assign_permissions_to_groups()` is placed both in `device/migrations/__init__.py` and `monitoring/migrations/__init__.py`
and one of them never get used in our code.

openwisp_monitoring/monitoring/migrations/__init__.py
Aryamanz29: Ok, I'll create a separate data migration in `monitoring` app named `add_inline_permissions.py` :+1:

openwisp_monitoring/device/admin.py
Aryamanz29: Yea, I feel the same this is not a correct way to do that, I'll update it :+1:

openwisp_monitoring/device/admin.py
Aryamanz29: I think it should be `if request.user.has_perm(f'{self.model._meta.app_label}.add_check_inline'):`

openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +63,32 @@ class CheckInline(GenericStackedInline):
...
+        if request.user.has_perm(f'{self.model._meta.app_label}.add_check_inline'):
+            self._set_check_inline()
+            return True
+        return False
+
+    def has_view_permission(self, request, obj=None):
+        if request.user.has_perm(f'{self.model._meta.app_label}.view_check_inline'):
+            self._set_check_inline()
+            return True
+        return super(GenericStackedInline, self).has_view_permission(request, obj)
```
pandafy, line 80: Why did you only extend the `has_view_permission` method but override other method?

openwisp_monitoring/device/admin.py
pandafy, line 68: I am not sure if this is a wrong way to achieve this functionality, but it does not feel right. Django provides hooks for returning `fields` and `readonly_fields`. Using them makes the code more readable.

openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +63,32 @@ class CheckInline(GenericStackedInline):
...
+
+    def has_change_permission(self, request, obj=None):
+        if request.user.has_perm(f'{self.model._meta.app_label}.change_check_inline'):
+            self._set_check_inline()
+            return True
         return False
```
pandafy, line 86: The current code won't allow a user that has permission to edit the `Check` model to edit the inline form. @nemesisdesign is this a desired quality?

openwisp_monitoring/device/admin.py
```diff
@@ -63,10 +63,32 @@ class CheckInline(GenericStackedInline):
...
 
     def has_delete_permission(self, request, obj=None):
+        if request.user.has_perm(f'{self.model._meta.app_label}.delete_check_inline'):
+            self._set_check_inline()
+            return True
         return False
```
pandafy, line 92: I would re-write this code in this way: 

```python
    fields = [
        'name',
        'is_active',
        'description',
        'check_type',
        'params',
    ]

    def get_fields(self, request, obj=None):
        if not self.has_change_permission(request, obj) or not self.has_view_permission(
            request, obj
        ):
            return ['check_type', 'is_active']
        return super().get_fields(request, obj)

    def get_readonly_fields(self, request, obj=None):
        if not self.has_change_permission(request, obj) or not self.has_view_permission(
            request, obj
        ):
            return ['check_type']
        return super().get_readonly_fields(request, obj)

    def has_add_permission(self, request, obj=None):
       # User will be able to add check from inline even if it only has permission 
       # to add a check object. 
        return super().has_add_permission(request, obj) or request.user.has_perm(
            f'{self.model._meta.app_label}.add_check_inline'
        )

    def has_change_permission(self, request, obj=None):
        return super().has_change_permission(request, obj) or request.user.has_perm(
            f'{self.model._meta.app_label}.change_check_inline'
        )

    def has_view_permission(self, request, obj=None):
        return super().has_view_permission(request, obj) or request.user.has_perm(
            f'{self.model._meta.app_label}.view_check_inline'
        )

    def has_delete_permission(self, request, obj=None):
        return super().has_delete_permission(request, obj) or request.user.has_perm(
            f'{self.model._meta.app_label}.delete_check_inline'
        )
```

I did rudimentary testing and it works. @Aryamanz29 please test it more and improve it

openwisp_monitoring/device/admin.py
```diff
@@ -84,7 +106,7 @@ def __init__(self, *args, **kwargs):
 
 class AlertSettingsInline(NestedStackedInline):
     model = AlertSettings
-    extra = 0
+    extra = 1
```
pandafy, line 109: Why do we need an extra here?

openwisp_monitoring/device/admin.py
```diff
@@ -109,15 +153,44 @@ class MetricInline(NestedGenericStackedInline):
     verbose_name = _('Alert Settings')
     verbose_name_plural = verbose_name
 
+    def _set_alertsettings_inline(self):
+        self.fields = ['name', 'is_healthy', 'field_name', 'configuration']
+        self.readonly_fields = ['is_healthy']
+
```
pandafy, line 159: Similar improvements here to use `get_fields` and `get_readonly_fields` here

openwisp_monitoring/monitoring/migrations/__init__.py
```diff
@@ -37,6 +37,38 @@ def assign_permissions_to_groups(apps, schema_editor):
...
+    # that do not have to be dealt with
+    except Group.DoesNotExist:
+        return
+
+    for model_name in operators_read_only_admins_manage:
+        try:
+            permission = Permission.objects.get(
+                codename='view_{}_inline'.format(model_name)
+            )
+            operator.permissions.add(permission.pk)
```
pandafy, line 62: We should also add `view` permission to the administrator for completeness.

pandafy (changes requested):
Good start @Aryamanz29! The current code is functional, but we need to improve this otherwise it will be come harder to maintain in future. Please see my comments below.

openwisp_monitoring/device/admin.py
Aryamanz29: I change this way to make sure `Inlines` works similar to our current version of code, i.e. When we have model permissions other than `view`, then they should **add/change/delete** only through  `CheckAdmin` & `MetricAdmin` but not from device `Inlines`.

**Ref** : https://github.com/openwisp/openwisp-monitoring/blob/52cf287a31026874fe73948badbbe174adfd250d/openwisp_monitoring/device/admin.py#L66-L70

https://github.com/openwisp/openwisp-monitoring/blob/52cf287a31026874fe73948badbbe174adfd250d/openwisp_monitoring/device/admin.py#L95-L99

https://github.com/openwisp/openwisp-monitoring/blob/52cf287a31026874fe73948badbbe174adfd250d/openwisp_monitoring/device/admin.py#L112-L116

openwisp_monitoring/device/admin.py
Aryamanz29: I have updated this :+1:

openwisp_monitoring/device/admin.py
Aryamanz29: thanks! Yes, it was working as expected :+1:

openwisp_monitoring/device/admin.py
Aryamanz29: I'll update that :+1:  thanks :)

openwisp_monitoring/device/admin.py
Aryamanz29: I changed `extra = 1` to get the complete form, but now it looks weird when the user has `view` permission (it shows empty input fields as well) so I reverted this change :+1: 

#### `With extra = 0`

Screenshot from 2022-08-20 17-07-43

#### `With extra = 1`

Screenshot from 2022-08-20 17-03-46

openwisp_monitoring/monitoring/migrations/__init__.py
Aryamanz29: Sure :+1:

openwisp_monitoring/device/admin.py
nemesisdesign: so it sounds like this was not intentional, right @Aryamanz29?

Aryamanz29: @nemesisdesign https://github.com/openwisp/openwisp-monitoring/pull/431#discussion_r950688809

openwisp_monitoring/device/admin.py
```diff
@@ -93,30 +128,64 @@ def get_queryset(self, request):
...
-        return super().get_queryset(request).filter(alertsettings__isnull=False)
+    def has_change_permission(self, request, obj=None):
+        return super().has_change_permission(request, obj) or request.user.has_perm(
+            f'{self.model._meta.app_label}.change_alertsettings_inline'
+        )
+
+    def has_delete_permission(self, request, obj=None):
+        return super().has_delete_permission(request, obj) or request.user.has_perm(
+            f'{self.model._meta.app_label}.delete_alertsettings_inline'
+        )
```
nemesisdesign, line 188: These methods to check for the additional permissions are always the same, except the suffix of the permission, therefore we could create a base class / mixin which has this logic and requires only a class attribute to be set which specifies the suffix, eg: `inline_permission_suffix = 'alertsettings_inline'`.

This allows us to follow the DRY principle closely and reduce maintenance whenever we need to change or upgrade this logic.

nemesisdesign (changes requested):
Looks good Aryaman!

It would be great to explain in the README how to use this feature to create the Iperf check when it is not created automatically (which will be the default behavior). What do you think @Aryamanz29?

See also my comments below.

openwisp_monitoring/monitoring/migrations/__init__.py
```diff
@@ -37,6 +37,38 @@ def assign_permissions_to_groups(apps, schema_editor):
             )
 
 
+def assign_inline_permissions_to_groups(apps, schema_editor):
+    create_default_permissions(apps, schema_editor)
+    operators_read_only_admins_manage = [
+        'check',
```
pandafy, line 43: @nemesisdesign should we move the data migration for the `Check` model to the `check` sub-app or is it okay here?

pandafy (changes requested):
@Aryamanz29 I found this weird behaviour with Alert Settings inline. It not occurring with Checks inline. 

Screenshot from 2022-08-23 13-25-28

Aryamanz29:
> @Aryamanz29 I found this weird behaviour with Alert Settings inline. It not occurring with Checks inline.
> 
> Screenshot from 2022-08-23 13-25-28

@pandafy https://github.com/openwisp/openwisp-monitoring/pull/431#discussion_r950689689
We can customize this form behaviour with `get_extra` hook as follows:

```py
     def get_extra(self, request, obj=None, **kwargs):
        # Get full alertsettings form
        # when user has add permission
        if self.has_add_permission(request, obj):
            return 1
        return super().get_extra(request, obj, **kwargs)
```
Screenshot from 2022-08-23 16-03-51

Aryamanz29:
https://user-images.githubusercontent.com/56113566/186871310-78f5cf55-2406-45f2-ac8b-eaee69881036.mp4

openwisp_monitoring/monitoring/migrations/__init__.py
nemesisdesign: this one makes more sense to have it in the checks app

nemesisdesign (commented):
While testing locally, I see a lot more empty alert settings, is this happening to you too?
If yes, why?

Screenshot from 2022-08-26 17-51-56

pandafy:
> While testing locally, I see a lot more empty alert settings, is this happening to you too? If yes, why?
> 
> Screenshot from 2022-08-26 17-51-56

I can confirm this behaviour on my local environment. With both superuser and non-superuser accounts.

nemesisdesign (changes requested):
Trying to filter for the current alert setting of the device:
```
AlertSettings.objects.filter(metric__object_id=<DEVICE_ID>')
```
I see that the shell returns the right ones.

But for some reason the admin UI is showing one additional empty alert setting for each metric, this does not look intentional, but by looking at the code it is not evident why this is happening. @Aryamanz29 do you have any idea?

Aryamanz29:
@nemesisdesign  @pandafy, I think it is due to `extra=1` in `AlertSettingsInline` when user `has_add_permission`

So currently we have following options for `AlertsettingsInline`:

- Let fix `extra=0` to avoid those empty `alertsettings`, then it will show `Add another Alert settings` button only for those metrics that doesn't have any `alertsettings`.

Screenshot from 2022-08-31 20-58-52

- But in that case, @pandafy reported that the form doesn't look good. See below :

Screenshot from 2022-08-31 21-04-13

- We could change `MetricInline` from **verbose_name** to `verbose_name = _('Metrics')` for that. 

Screenshot from 2022-08-31 21-07-51

Now question is why don't we just filter out metric queryset [ie. `super().get_queryset(request).filter(alertsettings__isnull=False)`] ❔ We can't do this because if user try to create a metric from the **'AlertSettings' tab** without having `alertsettings` then that metric can't be visible due to `alertsettings__isnull=False` filter.

If there is something we can try other than what is mentioned above, please let me know 😄

nemesisdesign (changes requested):
```diff
diff --git a/openwisp_monitoring/device/admin.py b/openwisp_monitoring/device/admin.py
index 0e3c7d4..ddcbe3f 100644
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -85,12 +85,13 @@ class CheckInline(InlinePermissionMixin, GenericStackedInline):
     extra = 0
     formset = CheckInlineFormSet
     fields = [
-        'name',
-        'is_active',
-        'description',
+        # 'name',
         'check_type',
-        'params',
+        'is_active',
+        # 'description',
+        # 'params',
     ]
+    # readonly_fields = ['name']
     inline_permission_suffix = 'check_inline'
 
     def get_fields(self, request, obj=None):
@@ -122,18 +123,18 @@ class AlertSettingsForm(ModelForm):
 
 class AlertSettingsInline(InlinePermissionMixin, NestedStackedInline):
     model = AlertSettings
-    extra = 0
-    max_num = 0
-    exclude = ['created', 'modified']
+    extra = 1
+    max_num = 1
+    exclude = ['name', 'created', 'modified']
     form = AlertSettingsForm
     inline_permission_suffix = 'alertsettings_inline'
 
-    def get_extra(self, request, obj=None, **kwargs):
-        # Get full alertsettings form
-        # when user has add permission
-        if self.has_add_permission(request, obj):
-            return 1
-        return super().get_extra(request, obj, **kwargs)
+    # def get_extra(self, request, obj=None, **kwargs):
+    #     # Get full alertsettings form
+    #     # when user has add permission
+    #     if self.has_add_permission(request, obj):
+    #         return 0
+    #     return super().get_extra(request, obj, **kwargs)
 
     def get_queryset(self, request):
         return super().get_queryset(request).order_by('created')
@@ -155,8 +156,8 @@ class MetricInline(InlinePermissionMixin, NestedGenericStackedInline):
     model = Metric
     extra = 0
     inlines = [AlertSettingsInline]
-    fields = ['is_healthy', 'configuration']
-    readonly_fields = ['is_healthy']
+    fields = ['name', 'is_healthy', 'configuration']
+    readonly_fields = ['name', 'is_healthy']
     # Explicitly changed name from Metrics to Alert Settings
     verbose_name = _('Alert Settings')
     verbose_name_plural = verbose_name

```

- can we show the metrics which have the alert settings first? 
- check name: can we add validation logic that copies the check config name in the name field (if the name is empty) for consistency?
- metric name: can we add validation logic that copies the metric config name in the name field (if the name is empty) for consistency?

Aryamanz29:
@nemesisdesign  @pandafy 
- [x] Can we show the metrics which have the alert settings first?

- Yes, it is possible. We can change **MetricInline**  `ordering = [F('alertsettings').desc(nulls_last=True)]`

- alertsettings_ordering

- [x] check name: Can we add validation logic that copies the check config name in the name field (if the name is empty) for consistency?

- Yes, as shown below, overriding the **CheckInline** formset `save_new()` method will work.

- check_inline

- [x] metric name: Can we add validation logic that copies the metric config name in the name field (if the name is empty) for consistency?

- Yes, overriding **MetricInline** form `save()` method will work, see below:

-alertsettings_inline

(Note : I have removed the `check name` and `metric name` fields from the form because it will be the same as configuration chosen by the user)

README.rst
```diff
@@ -1023,6 +1023,17 @@ jitter, datagram loss etc of the device using `iperf3 utility <https://iperf.fr/
...
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/device-inline-check.png
+  :align: center
+
+**Note:** To access the above page, the user must have some special permissions,
+which are included by default in the groups "Administrator" and "Operator" and are shown in the screenshot below.
+
+.. figure:: https://github.com/openwisp/openwisp-monitoring/raw/docs/docs/1.1/inline-permissions.png
+  :align: center
+
```
nemesisdesign, line 1036: Shall we mention also how to create the alert setting and indicate how to get alerted if the bandwidth is less than, for example, 30 Mmbit/s?

openwisp_monitoring/device/admin.py
```diff
@@ -48,26 +48,70 @@ def full_clean(self):
...
+                    obj,
                     self.ct_field.get_attname(),
                     ContentType.objects.get_for_model(self.instance).pk,
                 )
-                setattr(form.instance, self.ct_fk_field.get_attname(), self.instance.pk)
+                setattr(obj, self.ct_fk_field.get_attname(), self.instance.pk)
         super().full_clean()
 
+    def save_new(self, form, commit=True):
+        instance = super().save_new(form, commit=False)
```
nemesisdesign, line 59: is it possible to obtain the same by moving this logic to the model instead of here?
The problem with this code is that if we implement similar logic in the REST API or in other views, we will have to repeat it.

openwisp_monitoring/device/admin.py
```diff
@@ -82,41 +126,60 @@ def __init__(self, *args, **kwargs):
...
+class MetricForm(ModelForm):
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        # The name of the metric will be the same
+        # as the configuration chosen by the user
+        instance.name = instance.get_configuration_display()
+        if commit:
+            instance.save()
+            self.save_m2m()
+        return instance
```
nemesisdesign, line 150: same here

nemesisdesign (commented):
Great progress @Aryamanz29!

I think we're almost done, see my comments below.

openwisp_monitoring/device/admin.py
Aryamanz29: No problem, I'll move it to their respective models 👍

openwisp_monitoring/device/admin.py
Aryamanz29: Sure

README.rst
Aryamanz29: Yea, for this we have to change `MetricInline` visible fields to include `field_name` as well (we may add it in advance mode),  For example:
```py
    fieldsets = [
        (None, {'fields': ('is_healthy', 'configuration')}),
        (
            _('Advanced options'),
            {'classes': ('collapse',), 'fields': ('field_name',)},
        ),
    ]
```

Screenshot from 2022-09-03 16-42-56

@nemesisdesign I think we should add this in https://github.com/openwisp/openwisp-monitoring/pull/427 as you've already suggested in https://github.com/openwisp/openwisp-monitoring/pull/427#discussion_r951633955, what do you think?

nemesisdesign: Yes, this would work, as long as the field name defaults to the one configured and as long as we do validation of this value before saving to ensure it exists.

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -345,6 +345,195 @@ def test_wifisession_inline(self):
...
+            self.assertContains(
+                response,
+                'metric-content_type-object_id-0-alertsettings-0-DELETE',
+            )
+
+        with self.subTest('Test check & alert settings with model permissions'):
+            test_user_permissions = test_user.user_permissions.all().values_list(
+                'codename', flat=True
+            )
+            self.assertEqual(test_user_permissions.count(), 0)
```
pandafy, line 406: I would re-write this as following 

You are not using the content of the returned list, hence there's no point in requesting that data from the database.

pandafy, line 406: I would re-write this as following 

```suggestion
            self.assertEqual(test_user.user_permissions.count(), 0)
```

You are not using the content of the returned list, hence there's no point in requesting that data from the database. Please update other occurrences as well.

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -345,6 +345,195 @@ def test_wifisession_inline(self):
...
+            device_permissions = Permission.objects.filter(codename__endswith='device')
+            # Permissions required to access device page
+            test_user.user_permissions.add(*device_permissions),
+            test_user_permissions = test_user.user_permissions.all().values_list(
+                'codename', flat=True
+            )
+            self.assertEqual(test_user_permissions.count(), 4)
```
pandafy, line 413: This logic is shared between all subtests. I would make an internal method for this test and re-use it.

pandafy (changes requested):
I have manually tested the changes and it works as expected. 

The ordering of the AlertSettings objects in the inline appears to be random (apart from showing the Metrics without AlertSettings at the last). Can we add another ordering field `name`? 

I have suggested some improvements to the test to make it more readable.

openwisp_monitoring/device/tests/test_admin.py
Aryamanz29: I'll update it, thanks!

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -345,6 +345,195 @@ def test_wifisession_inline(self):
...
+            )
+            alertsettings_model_permissions = Permission.objects.filter(
+                codename__endswith='alertsettings'
+            )
+            # Add check, metric & alert model permissions
+            test_user.user_permissions.add(
+                *check_model_permissions,
+                *metric_model_permissions,
+                *alertsettings_model_permissions,
+            )
```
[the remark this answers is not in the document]
Aryamanz29: Yes, we can :+1:

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -345,6 +345,195 @@ def test_wifisession_inline(self):
...
+                # Permissions required to access device page
+                test_user.user_permissions.add(*device_permissions),
+                test_user_permissions = test_user.user_permissions.all().values_list(
+                    'codename', flat=True
+                )
+                self.assertEqual(test_user_permissions.count(), 4)
+                response = self.client.get(url)
+                self.assertEqual(response.status_code, 200)
+                self.assertNotContains(response, '<h2>Checks</h2>', html=True)
+                self.assertNotContains(response, '<h2>Alert Settings</h2>', html=True)
```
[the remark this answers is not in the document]
Aryamanz29: I'll update it, thanks :+1:

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -345,6 +345,195 @@ def test_wifisession_inline(self):
...
+                'codename', flat=True
+            )
+            self.assertEqual(test_user_permissions.count(), 16)
+            response = self.client.get(url)
+            self.assertEqual(response.status_code, 200)
+            _assert_check_inline_in_response(response)
+            _assert_alertsettings_inline_in_response(response)
+
+        with self.subTest('Test check & alert settings with inline permissions'):
+            with self.subTest('Test all inline permissions'):
```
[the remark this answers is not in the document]
Aryamanz29: I'll update that :+1:

Aryamanz29:
> I have manually tested the changes and it works as expected.
> 
> The ordering of the AlertSettings objects in the inline appears to be random (apart from showing the Metrics without AlertSettings at the last). Can we add another ordering field `name`?
> 
> I have suggested some improvements to the test to make it more readable.

@pandafy  I tried adding both orders, but only one works, which I believe is due to one ordering with `desc(nulls_last=True)` and then ordering with `name`, which simply overrides the previous one.

README.rst
Aryamanz29: > Yes, this would work, as long as the field name defaults to the one configured and as long as we do validation of this value before saving to ensure it exists.

@nemesisdesign I think we should add this in https://github.com/openwisp/openwisp-monitoring/pull/427 as you've already suggested in https://github.com/openwisp/openwisp-monitoring/pull/427#discussion_r951633955, what do you think?

nemesisdesign: Yes, right, ok!

pandafy (commented):
In my recent testing, I found out that it is not possible to remove an AlertSetting form the the inline without deleting the metric. 

Steps to replicate the issue: 
1. Edit a traffic metric and add update `Operator`, `Tolerance` and `Threshold` to some value
2. Save the device
3. I tried to remove the AlertSetting by clearing out the field, but that didn't work and it used the default values as fallback. 

I think it will be better to address this problem in a separate issue.

Aryamanz29:
> In my recent testing, I found out that it is not possible to remove an AlertSetting form the the inline without deleting the metric.
> 
> Steps to replicate the issue:
> 
> 1. Edit a traffic metric and add update `Operator`, `Tolerance` and `Threshold` to some value
> 2. Save the device
> 3. I tried to remove the AlertSetting by clearing out the field, but that didn't work and it used the default values as fallback.
> 
> I think it will be better to address this problem in a separate issue.

@pandafy I opened this here : https://github.com/openwisp/openwisp-monitoring/issues/435

nemesisdesign (approved):

PR #432, opened 2022-08-17
Zooming charts: reload data in order to provide a detailed view 

### LINKED ISSUE #27: [feature] Zooming graphs: reload data in order to provide a detailed view
Depends on https://github.com/openwisp/openwisp-monitoring/issues/26.

Let's say we're viewing a graph for 1 year of data and we zoom in to 2 specific days in which an incident happened.

Right now we wouldn't see much, because the zoom will simply enlarge the current monthly graph, which is grouping results by 24 hours, so we'll simply see 2 big bars representing the 2 days we're interested into.

The ideal behavior would be to reload the data of those 2 days by leveraging the work that will be done in #26.
The expected result would be see something similar to the 3 day graph (many bars because results are grouped by 10-20 minutes), but it would happen by just zooming on the graph.

On the JS side, we'll have to bind to zoom events of the plotly library: https://plotly.com/javascript/zoom-events/
When a zoom event is fired, get the date range, and query the API to get that range accordingly.

Then we shall also implement sliding as discussed.

yashikajotwani12:
DEMO:

https://user-images.githubusercontent.com/77020164/185146478-ee459d38-e987-46c1-8680-58579ae6a35d.mp4

Fixes #27 
Checks:


atb00ker:
@yashikajotwani12 attached video is not showing for me, can you pl. double check? 😄

yashikajotwani12:
> @yashikajotwani12 attached video is not showing for me, can you pl. double check? smile

It works for me, sorry for the late reply.

nemesisdesign:
Superseded by #450.

PR #436, opened 2022-09-07
[change] Added possibility to delete `AlertSettings` from device tab

### LINKED ISSUE #435: [change] Allow users to delete `AlertSettings` from the device tab
Reported by @pandafy in https://github.com/openwisp/openwisp-monitoring/pull/431#pullrequestreview-1098796108

```
It is not possible to remove an `AlertSettings` from the the inline without deleting the `Metric`.

Steps to replicate the issue

1. Edit a traffic metric and add update **Operator**, **Tolerance** and **Threshold** to some value.
2. Save the `device`.
3. I tried to remove the `AlertSettings` by clearing out the field, but that didn't work and it used the default values as fallback.
```

Aryamanz29:
- When user clear all `AlertSettings` fields then `AlertSettings` object will automatically deletes after hitting save button.
- If user fill any field of `AlertSettings` then other fields automatically get default value that are defined inside `AlertSettings` model.
    
Closes #435


openwisp_monitoring/device/admin.py
```diff
@@ -116,6 +116,27 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
 
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        # If all the fields are empty,
+        # then just delete the object
+        # to prevent the default value
+        # from being used as a fallback
+        if all(
+            fields is None
```
pandafy, line 126:
```suggestion
            not fields
```

Let's write is this way to also account for empty strings.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -703,15 +703,6 @@ class Meta:
...
 
-    def full_clean(self, *args, **kwargs):
-        if self.custom_threshold == self.config_dict['threshold']:
-            self.custom_threshold = None
-        if self.custom_tolerance == self.config_dict['tolerance']:
-            self.custom_tolerance = None
-        if self.custom_operator == self.config_dict['operator']:
-            self.custom_operator = None
-        return super().full_clean(*args, **kwargs)
-
```
pandafy, line 714: Why do we need to remove this?

pandafy (changes requested):
@Aryamanz29 I would add a test that performs clearing of fields action from using the AlertSettingInline (i.e. make POST request to device change page to delete the alert).

openwisp_monitoring/monitoring/base/models.py
Aryamanz29: If we try to add a new `AlertSettings` with `operator (less than)` and let other fields remain empty, then we get an empty `AlertSettings form (see below)` without having default values. This is due to the fact that the **full_clean()** method sets fields to **None**. Therefore, I removed that method to prevent this behaviour.

**Note**: This behaviour is only seen when we choose the operator `< (less than)`.

https://user-images.githubusercontent.com/56113566/191241144-9e34390d-7449-4a0e-ad3c-2cb6b6dcc234.mp4

pandafy (changes requested):
@Aryamanz29 I made these changes locally and found that the admin UI was working as expected. Can you please do some more testing with the following suggestions and make improvements if required. 

```diff
diff --git a/openwisp_monitoring/device/admin.py b/openwisp_monitoring/device/admin.py
index 4824c35..a5d9f36 100644
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -116,22 +116,26 @@ class AlertSettingsForm(ModelForm):
             }
         super().__init__(*args, **kwargs)
 
-    def save(self, commit=True):
-        instance = super().save(commit=False)
-        # If all the fields are empty,
-        # then just delete the object
-        # to prevent the default value
-        # from being used as a fallback
+    def _post_clean(self):
+        self.instance._delete_instance = False
         if all(
-            not fields
-            for fields in [
-                instance.custom_operator,
-                instance.custom_threshold,
-                instance.custom_tolerance,
+            self.cleaned_data[field] is None
+            for field in [
+                'custom_operator',
+                'custom_threshold',
+                'custom_tolerance',
             ]
         ):
-            instance.delete()
-            return instance
+            # "_delete_instance" flag signifies that
+            # the fields have been set to None by the
+            # user. Hence, the object should be deleted.
+            self.instance._delete_instance = True
+        super()._post_clean()
+
+    def save(self, commit=True):
+        if self.instance._delete_instance:
+            self.instance.delete()
+            return self.instance
         return super().save(commit)
 
 
diff --git a/openwisp_monitoring/monitoring/base/models.py b/openwisp_monitoring/monitoring/base/mo
dels.py
index dda3676..17ebf03 100644
--- a/openwisp_monitoring/monitoring/base/models.py
+++ b/openwisp_monitoring/monitoring/base/models.py
@@ -703,6 +703,15 @@ class AbstractAlertSettings(TimeStampedEditableModel):
             ('view_alertsettings_inline', 'Can view Alert settings inline'),
         )
 
+    def full_clean(self, *args, **kwargs):
+        if self.custom_threshold == self.config_dict['threshold']:
+            self.custom_threshold = None
+        if self.custom_tolerance == self.config_dict['tolerance']:
+            self.custom_tolerance = None
+        if self.custom_operator == self.config_dict['operator']:
+            self.custom_operator = None
+        return super().full_clean(*args, **kwargs)
+
     @property
     def config_dict(self):
         return self.metric.config_dict.get(
```

Aryamanz29:
> @Aryamanz29 I made these changes locally and found that the admin UI was working as expected. Can you please do some more testing with the following suggestions and make improvements if required.

I also tested these suggestions and everything is working fine :)

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -480,6 +513,118 @@ def _assert_alertsettings_inline_in_response(response):
...
+            alertsettings = AlertSettings.objects.first()
+            self.assertEqual(alertsettings.operator, operator)
+            self.assertEqual(alertsettings.threshold, threshold)
+            self.assertEqual(alertsettings.tolerance, tolerance)
+
+        with self.subTest('Test alert settings inline when all fields are provided'):
+            self._device_params.update(test_inline_params)
+            response = self.client.post(url, self._device_params)
+            _assert_alertsettings_inline(response, '<', 9, 1800)
+            _reset_alertsettings_inline()
```
pandafy, line 566: @Aryamanz29 the resetting should be done outside of the sub-test.

Aryamanz29: ahh 😅

pandafy (approved):
LGTM 👍🏼 

Deferring merge to @nemesisdesign

nemesisdesign (commented):
@pandafy @Aryamanz29 this looks good to me, however, I am seeing this in my dev env and I do not remember whether we concluded this to be ok or not.

Screenshot from 2022-09-26 20-19-38

Can you please referesh my memory on the following:

- why is it needed to display the alert settings for all metrics?
- why does the gsoc22-iperf branch display the "field_name" form field, while this branch does not?


It may be happening because of a previous test which created these objects which now are shown. Can you pelase refresh my memory? I don't remember very well.

Aryamanz29:
@nemesisdesign 
> * why is it needed to display the alert settings for all metrics?

- We decided to do this way because users have access to **add/change/delete (metrics and alertsettings)** and if a user adds a new **metric** without having any **alertsettings**, then due to `get_queryset(request).filter(alertsettings__isnull=False)`. This particular **metric** is not visible to users, and if they try to create the same metric again, they'll get `ObjectAlreadyExists`. Now we prevent **add/change/delete** action from **metric**, we can still use `get_queryset(request). filter (alertsettings__isnull=False)`, which only shows metrics that have **alertsettings** objects, but some **metrics**, such as Iperf, don't create any **alertsettings** by default, so including `metrics` that don't have any **alertsettings** in inline is still useful. 

-  We can do following steps:
   - Let the system handle the creation and deletion of the **metric**, and we only allow users to use custom **alertsettings** from inline after the creation of their respective **metric**. Users can still able to delete **alertsettings** objects by clearing out all fields without interfering with the **metric** object. 
  
   -  The second option is that we can add code to `Iperf check class` to create default **alertsettings** with **enabled=False** and use the above filtering to show only those metrics in inline that have **alertsettings** objects. This way, we get **alertsettings** with default values and users can configure it later on. One drawback to this option is that the [`Alert on Metric field feature`](https://github.com/openwisp/openwisp-monitoring/pull/427) is only limited to those **metrics** that have **alertsettings** objects created by the system.
  
> * why does the gsoc22-iperf branch display the "field_name" form field, while this branch does not?

- It's because this PR was rebased from https://github.com/openwisp/openwisp-monitoring/pull/431. Now I have rebased this with the current `gsoc22-iperf branch`.

Screenshot from 2022-09-27 15-29-44

nemesisdesign (commented):
@Aryamanz29 let's proceed as discussed today

Aryamanz29:
@nemesisdesign I think we should also add metric `configuration` to `readonly_fields` what do you think?

Screenshot from 2022-10-01 23-04-43

#### Configuration read only :-

Screenshot from 2022-10-01 23-07-13

nemesisdesign (changes requested):
What about something like:

```diff
diff --git a/openwisp_monitoring/device/admin.py b/openwisp_monitoring/device/admin.py
index bf65c2f..b5c77eb 100644
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -156,14 +156,14 @@ class MetricInline(InlinePermissionMixin, NestedGenericStackedInline):
     extra = 0
     inlines = [AlertSettingsInline]
     fieldsets = [
-        (None, {'fields': ('is_healthy', 'configuration')}),
+        (None, {'fields': ('name', 'is_healthy')}),
         (
             _('Advanced options'),
             {'classes': ('collapse',), 'fields': ('field_name',)},
         ),
     ]

-    readonly_fields = ['is_healthy']
+    readonly_fields = ['name', 'is_healthy']
     # Explicitly changed name from Metrics to Alert Settings
     verbose_name = _('Alert Settings')
     verbose_name_plural = verbose_name
```

Aryamanz29:
@nemesisdesign  I think we have to add `configuration` (as readonly)  to `AlertSettingsInline` to able to test this feature.
See below (configuration is required field to create a metric in `unittests`):
https://github.com/openwisp/openwisp-monitoring/pull/436/files#diff-bcbe5b114c5eae4533d6dfad208b8cd81e85bc45220fe3ac68598faa12d354dbR518-R539

If we remove `configuration` from Inline, we'll get following error & test fails : 
```
  File "/home/aryaman/Desktop/Iperf-GSoC/openwisp-monitoring/openwisp_monitoring/monitoring/base/models.py", line 178, in config_dict
    raise InvalidMetricConfigException(
openwisp_monitoring.monitoring.exceptions.InvalidMetricConfigException: Invalid metric configuration: "None"

```

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -480,6 +511,118 @@ def _assert_alertsettings_inline_in_response(response):
...
+            'name': device.name,
+            'organization': str(device.organization.id),
+            'mac_address': device.mac_address,
+            'key': device.key,
+            # metric & alertsettings
+            f'{metric_model_name}-content_type-object_id-TOTAL_FORMS': '1',
+            f'{metric_model_name}-content_type-object_id-INITIAL_FORMS': '0',
+            f'{metric_model_name}-content_type-object_id-MIN_NUM_FORMS': '0',
+            f'{metric_model_name}-content_type-object_id-MAX_NUM_FORMS': '1000',
+            f'{metric_model_name}-content_type-object_id-0-configuration': 'iperf',
```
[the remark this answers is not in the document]
Aryamanz29: yes :+1:

PR #449, opened 2022-10-21
[fix] Changed iperf3 check `--connect-timeout` to 1 sec #448

### LINKED ISSUE #448: [bug] Iperf3 check connection timeout default value
The `--connect-timeout` parameter accepts value in millisecond. The default value for this is set to `1` right now due to which the check fails from other devices.

https://github.com/openwisp/openwisp-monitoring/blob/11b31ffea2e32b93085e04b5fcb806682cbecc58/openwisp_monitoring/check/classes/iperf3.py#L102-L112

We should make it atleast 1 second to make it work out of the box for users.

Aryamanz29:
Fixes #448

Checks:


pandafy (approved):
Thanks for fixing this! 😄

nemesisdesign (approved):
Thanks :+1:

PR #450, opened 2022-10-21
[feature] Added zooming graphs to provide a detailed view #27

### LINKED ISSUE #27: [feature] Zooming graphs: reload data in order to provide a detailed view

Aryamanz29:
-  Added custom zooming to monitoring charts. 
- `Zoom In` : To get a detailed view of all the monitoring charts for a specific time range. 
- `Zoom Out` : To reset all monitoring charts to the last selected time range.

### `Demo`

https://user-images.githubusercontent.com/56113566/198017360-a56d1ba2-fe28-4738-9886-eb0851cc04dc.mp4

Closes #27

Checks:

### DIFF
--- a/openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
+++ b/openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
@@ -244,15 +244,11 @@ django.jQuery(function ($) {
       var startLabel = localStorage.getItem(startDayKey) || moment().format('MMMM D, YYYY');
       var endLabel = localStorage.getItem(endDayKey) || moment().format('MMMM D, YYYY');
       
-      // Disable zoom charts scrolling on page refresh
+      // Disable the zoom chart and scrolling when we refresh the page
       localStorage.setItem(isChartZoomScroll, false);
-      if (localStorage.getItem(isChartZoomed) === 'true') {
-        range = localStorage.getItem(zoomtimeRangeKey);
-        endLabel = localStorage.getItem(zoomEndDayKey);
-        startLabel = localStorage.getItem(zoomStartDayKey);
-      }
-  
-      if (localStorage.getItem(isCustomDateRange) === 'true' || localStorage.getItem(isChartZoomed) === 'false') {
+      localStorage.setItem(isChartZoomed, false);
+
+      if (localStorage.getItem(isCustomDateRange) === 'true') {
         // Add label to daterangepicker widget
         addDateRangePickerLabel(startLabel, endLabel);
         // Set last selected custom date after page reload
--- a/openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
+++ b/openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
@@ -47,22 +47,22 @@ django.jQuery(function ($) {
 
     function isMonitoringChartsLocation() {
       // If active monitoring charts location is not #ow-chart-container and not admin
-      return window.location.hash === '#ow-chart-container' ||  window.location.pathname === '/admin/'
+      return window.location.hash === '#ow-chart-container' ||  window.location.pathname === '/admin/';
     }
 
     function handleChartZoomChange(chartsContainers) {
-      console.log(window.location.pathname)
-      // If not monitoring charts location, then just return
+      // Simply return if we are not at the monitoring chart location
       if (!isMonitoringChartsLocation()) {
         return;
       }
       // Handle chart zooming with custom dates
       var zoomCharts = document.getElementsByClassName(chartsContainers);
-      // Set zoomChartId, required for scrolling after the zoom event
+      // Set zoomChartId, required for scrolling after the zoom-in event
       $('.js-plotly-plot').on("click dblclick mouseover mouseout", function () {
         var zoomChartId = $(this).parent().prop('id');
         if (zoomChartId === 'chart-0') {
-         zoomChartId = window.location.hash === '#ow-chart-container' ? 'container' : 'ow-chart-inner-container'
+         var activeChartsLocation = window.location.hash;
+         zoomChartId = activeChartsLocation === '#ow-chart-container' ? 'container' : 'ow-chart-inner-container';
         }
         localStorage.setItem(zoomChartIdKey, zoomChartId);
       });
@@ -76,8 +76,7 @@ django.jQuery(function ($) {
           function (eventdata) { // jshint ignore:line
             var eventEnd = eventdata['xaxis.range[1]'];
             var eventStart = eventdata['xaxis.range[0]'];
-            console.log(eventEnd, eventStart)
-            // If not monitoring charts location, then just return
+            // Simply return if we are not at the monitoring chart location
             if (!isMonitoringChartsLocation()) {
               return;
             }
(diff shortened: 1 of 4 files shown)


Aryamanz29:
WIP

openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
```diff
@@ -36,6 +43,73 @@ django.jQuery(function ($) {
...
+        if (!zoomChart) {
+          localStorage.setItem(isChartZoomed, false);
+          localStorage.setItem(isCustomDateRange, false);
+          return;
+        }
+        zoomChart.on('plotly_relayout',
+          function (eventdata) { // jshint ignore:line
+            var eventEnd = eventdata['xaxis.range[1]'];
+            var eventStart = eventdata['xaxis.range[0]'];
+            if (!eventEnd || !eventStart) {
```
nemesisdesign, line 59: what case is this? Do you care to clarify with a comment? Is this when the chart gets dragged?

openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
```diff
@@ -175,9 +261,14 @@ django.jQuery(function ($) {
       var time = localStorage.getItem(timeRangeKey);
       location.href = baseUrl + time + '&csv=1';
       // If custom pass pickerEndDate and pickerStartDate to csv url
-      if (localStorage.getItem(isCustomDateRange) === 'true') {
+      if (localStorage.getItem(isCustomDateRange) === 'true' || localStorage.getItem(pickerChosenLabelKey) === 'Custom Range') {
```
nemesisdesign, line 264: where is `'Custom Range'` coming from?

nemesisdesign (changes requested):
Great progress @Aryamanz29! :+1: 

It's great to see that moving the chart around with dragging also works :+1: 

This is not exahustive yet but only the first issues I found.

### Zoom reset

What should we do when the zoom is reset multiple times?
Can we keep track of the previous zoom levels so the user can go back by just doing double click?
Or do we want to keep track only of the initial zoom level and go back to it when the user resets the zoom? In this case the user will be able to reset it only one time.

### Zooming or dragging charts below the first

After zooming-in, the charts are reloaded and the scroll level messes up, so the user starts seeing the first chart instead of the one they zoomed on.

Peek 2022-11-01 16-18

Did you notice this? We need to keep track of which chart is being manipulated and make sure we scroll back to it when it reappears.

openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
Aryamanz29: When the chart is zoomed out, this code will be executed, and I'll add a comment for it.

Aryamanz29:
> ### Zoom reset

> What should we do when the zoom is reset multiple times? Can we keep track of the previous zoom levels so the user can go back by just doing double click? Or do we want to keep track only of the initial zoom level and go back to it when the user resets the zoom? In this case the user will be able to reset it only one time.

- Right now I have implemented zooming to keep track of initial zoom level and looks good to me because if we keep track of multiple zoom levels then we may need to store some more data to `localStorage` and also need to keep track of every zoom level's custom date range, which makes the current implementation complicated and users may also get lost between zooming different levels.


> ### Zooming or dragging charts below the first
>After zooming-in, the charts are reloaded and the scroll level messes up, so the user starts seeing the first chart instead of the one they zoomed on.

- This issue is fixed now, See below:


https://user-images.githubusercontent.com/56113566/199586941-4fd058a0-fa29-455a-9792-59b34d216b35.mp4

openwisp_monitoring/monitoring/static/monitoring/js/chart-utils.js
Aryamanz29: from `#daterangepicker-widget`

nemesisdesign (changes requested):
Great work @Aryamanz29, I have been stress testing this a bit more and I observed something which I think is not right:

Selecting from May 12 to June 25 doesn't exactly yield the expected result, the dates and data are a little off:

Peek 2022-11-03 19-33

On this time frame of slightly more than 1 month I would expect to see 1 bar per day but we're getting one bar every 2 days, why is that? Maybe this is why dates and data don't match what I was expecting.

Having 1 bar per day for timeframes higher than a month is most likely to be interesting for our users.

I did not notice this in the previous PR which adds the timeseries range, I think this change was introduced back there.

For anything higher than 1 month we should have 1 bar per day, maybe for anything more than 7 months we can have 1 bar per week.

One question for something I forgot to verify in the date range PR: do we have a maximum date range? It should be 365 days max.

Can you make sure we have one dedicated commit to amend this behavior?

nemesisdesign (changes requested):
I am waiting to merge the following PRs which seems to have some overlapping changes:

- https://github.com/openwisp/openwisp-monitoring/pull/454
- https://github.com/openwisp/openwisp-monitoring/pull/469

Once these two are merged we can rebase this on the latest master and test it again to confirm all is good.

nemesisdesign (changes requested):
I found 2 glitches during testing:

1. in the general charts in the main admin dashboard (`/admin/`), if you zoom on one of the charts, the view is moved from the charts to the top of the page
2. if the HTTP request for zooming fails, the charts are emptied, can you please double check if it's possible to avoid this? Ideally we should replace the charts only after the data has been loaded, as we have done in #469

Aryamanz29:
WIP

Aryamanz29:
### `Tested locally`


https://user-images.githubusercontent.com/56113566/208980712-2624f829-e2e4-43eb-9295-4102d8b421ea.mp4

Aryamanz29:
### Tested locally

https://user-images.githubusercontent.com/56113566/209671026-303a3a23-d386-4a7a-a5b3-9efc99de1091.mp4

- [x] Zoom In: To see charts for a specific date range in greater detail. 
- [x] Zooming In future date times (to fix http req fails) : On the client side, a check has been added to ensure that `end_datetime` is never greater than `current_datetime`. 
- [x] Zoom Out event: To reset charts to their initial date range.
- [x] When monitoring data is not available when zooming in, it is better to show a fallback container to prevent any zooming event abuse (which causes multiple API calls and results in an unresponsive app); see below:

#### When the chart data is unavailable, it is better to use a fallback container:

Screenshot from 2022-12-27 18-14-28

#### Zooming event abuse may occur if we check monitoring data and then replace charts:

https://user-images.githubusercontent.com/56113566/209671436-48b0618b-f9f8-4636-9a2b-47149a1de10c.mp4

nemesisdesign (changes requested):
Almost there @Aryamanz29, last detail discussed is to avoid locking in the zoomed-in date to avoid confusing users who may still think they're on the original time frame when they leave the page.

PR #454, opened 2022-10-27
[docs] Mentioned usage of custom daterange #453

### LINKED ISSUE #453: [docs] Mention custom date range usage in `List of endpoints` section
We recently added a date range widget to the project, which updates existing API endpoints (General & Device charts), so we need to update the [List of endpoints](https://github.com/openwisp/openwisp-monitoring#list-of-endpoints) section of the docs.

Aryamanz29:
Closes #453

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2362,9 +2362,8 @@ multi-tenancy and allows filtering monitoring data by ``organization_slug``,
 
     GET /api/v1/monitoring/dashboard/?organization_slug=<org1-slug>,<org2-slug>&location_id=<location1-id>,<location2-id>&floorplan_id=<floorplan1-id>,<floorplan2-id>
 
-Another parameter that can be used is ``time``, which allows to specify the time frame
-or if you want to get general monitoring charts data for a specific date range then you can also add
-parameters ``start`` and  ``end`` to the API endpoint as shown below.
+Another parameter that can be used is ``time``, which allows you to specify the time frame, or if you want to get
+general monitoring chart data for a specific date range, you can add ``start`` and ``end`` parameters to the API endpoint as shown below.
 
 - ``1d``: returns data of the last day
 - ``3d``: returns data of the last 3 days
@@ -2376,7 +2375,7 @@ To get general monitoring charts data for a specific date range:
 
 .. code-block:: text
 
-    GET /api/v1/monitoring/dashboard/?time={timeframe}&start={start_datetime}&end={end_datetime}
+    GET /api/v1/monitoring/dashboard/?start={start_datetime}&end={end_datetime}
 
 **Note**: ``start`` and  ``end`` parameters should be in the format
 ``YYYY-MM-DD H:M:S``, otherwise 400 Bad Response will be returned.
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -2344,15 +2344,25 @@ multi-tenancy and allows filtering monitoring data by ``organization_slug``,
 
     GET /api/v1/monitoring/dashboard/?organization_slug=<org1-slug>,<org2-slug>&location_id=<location1-id>,<location2-id>&floorplan_id=<floorplan1-id>,<floorplan2-id>
 
-Another parameter that can be used is ``time``, which allows to specify the
-time frame, the allowed values are:
+Another parameter that can be used is ``time``, which allows to specify the time frame
+or if you want to get general monitoring charts data for a specific date range then you can also add
+parameters ``start`` and  ``end`` to the API endpoint as shown below.
```
nemesisdesign, line 2349: Are `start` and `end` an alternative to `time`? Does `time` have to be specified when `start` and `end` are supplied?

nemesisdesign (commented):
It's still not clear to me, please see my question below.

README.rst
Aryamanz29: Thanks for pointing this out, @nemesisdesign. Currently it is necessary to pass a `time` parameter with `start` and `end`, but I think a time parameter doesn't make sense in the case of a custom date range (we should calculate days after validating custom dates in views). I have updated the code in [`6623e56` (#450).](https://github.com/openwisp/openwisp-monitoring/pull/450/commits/6623e5635d200eec17ad82eba8fe8dffd55f88ec)

nemesisdesign: I think it makes more sense to cherry-pick that commit into this patch and rename this patch to be a clarification and correction of this behavior. What do you think?

Aryamanz29: Sure, Updated :+1:

openwisp_monitoring/views.py
```diff
@@ -64,6 +64,13 @@ def get(self, request, *args, **kwargs):
         # if custom dates are provided then validate custom dates
         if start_date and end_date:
             self._validate_custom_date(start_date, end_date, timezone)
+            # if valid custom dates then calculate custom days
+            time = '1d'
+            end = datetime.strptime(end_date, '%Y-%m-%d %H:%M:%S')
+            start = datetime.strptime(start_date, '%Y-%m-%d %H:%M:%S')
+            custom_days = (end - start).days
+            if custom_days:
+                time = f'{custom_days}d'
```
nemesisdesign, line 73: I don't understand: why is this change needed?

nemesisdesign (changes requested):
I did some changes to the text, please double check, see also my question below.

openwisp_monitoring/views.py
Aryamanz29: Because we removed `time` from the query parameter, we must calculate the _`days diff (end - start)`_ after validating the custom date strings in the views in order to update [`custom_group_map`](https://github.com/openwisp/openwisp-monitoring/blob/ec248d1cee0cff8d81714d7c9d50adbab8f858f3/openwisp_monitoring/monitoring/base/models.py#L548-L577) accordingly.

Ref : https://github.com/openwisp/openwisp-monitoring/pull/454#discussion_r1020930821

nemesisdesign: ok thanks for the refersh.

I see that we are parsing the dates twice, which is bad.
We should modify the `_validate_custom_date` so that it returns the dates we need here, I would also rewrite this code as follows:

```python
start_datetime, end_datetime = self._validate_custom_date(start_date, end_date, timezone)
# if valid custom dates then calculate custom days
time = '1d'
custom_days = (start_datetime - end_datetime).days
if custom_days:
    time = f'{custom_days}d'
```

nemesisdesign (changes requested):

openwisp_monitoring/views.py
Aryamanz29: thanks! @nemesisdesign

nemesisdesign (approved):

PR #457, opened 2022-11-01
[fix] Fixed `config_applied` check to a device flags it as PROBLEM #451

### LINKED ISSUE #451: [bug] Change of config to a device flags it as PROBLEM
As soon as the configuration is changed on a device, the health status moves into PROBLEM state.

This seems an unwanted effect of changing the implementation to ignore tolerance for changing the status (some time ago we changed the code so that if a device is not reachable, the status will immediately change to not reachable, even if there is a high tolerance, because the tolerance is meant for the alert only, also because it doesn't make sense to show a device which is not reachable as OK, as it made the system not credible).

I think a possible solution for this would be to change the implementation of this check, which was written with the old tolerance implementation in mind and is super simple: 

https://github.com/openwisp/openwisp-monitoring/blob/fdfbe243428c9d7cb34e66445aa2ed2d24e87b9d/openwisp_monitoring/check/classes/config_applied.py#L10-L27

It basically just check whether the config status is applied or not and based on that changes the health status of the metric, which now does not make sense anymore for the following reasons:

- the device config is changed and its status goes into modified
- it seems the check runs immediately and flags the metric as unhealthy and hence the health status of the device moves into problem

We could change the implementation of the check to look at the modified date: if the config is not applied and the config was modified more than X minutes ago, then return zero. 

X should be configurable and we shall have a setting for it since the openwisp-config interval can be changed, eg: `OPENWISP_MONITORING_CONFIG_CHECK_INTERVAL`.

Aryamanz29:
Fixes #451 

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1488,6 +1488,18 @@ validating custom parameters of a ``Check`` object.
 This setting allows you to choose whether `config_applied <#configuration-applied>`_ checks should be
 created automatically for newly registered devices. It's enabled by default.
 
+``OPENWISP_MONITORING_CONFIG_CHECK_INTERVAL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-------------+
+| **type**:    | ``int``     |
++--------------+-------------+
+| **default**: | ``5``       |
++--------------+-------------+
+
+This setting allows you to configure the config check interval used by
+`config_applied <#configuration-applied>`_. By default it is set to 5 minutes.
+
 ``OPENWISP_MONITORING_AUTO_IPERF3``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/README.rst
+++ b/README.rst
@@ -1489,7 +1489,7 @@ This setting allows you to choose whether `config_applied <#configuration-applie
 created automatically for newly registered devices. It's enabled by default.
 
 ``OPENWISP_MONITORING_CONFIG_CHECK_INTERVAL``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +--------------+-------------+
 | **type**:    | ``int``     |
--- a/openwisp_monitoring/check/classes/config_applied.py
+++ b/openwisp_monitoring/check/classes/config_applied.py
@@ -1,12 +1,29 @@
+from datetime import datetime
+
+from django.conf import settings
+from pytz import timezone
 from swapper import load_model
 
 from ...device.utils import SHORT_RP
+from ..settings import CONFIG_CHECK_INTERVAL
 from .base import BaseCheck
 
 AlertSettings = load_model('monitoring', 'AlertSettings')
 
 
 class ConfigApplied(BaseCheck):
+    def _get_result(self, config_status, config_modified_mins_ago):
+        """
+        Returns zero, if the config is not applied and the config was modified
+        more than OPENWISP_MONITORING_CONFIG_CHECK_INTERVAL mins ago, otherwise returns one
+        """
+        return int(
+            not (
+                config_status != 'applied'
+                and config_modified_mins_ago > CONFIG_CHECK_INTERVAL
+            )
+        )
+
     def check(self, store=True):
         # If the device is down or does not have a config
         # do not run config applied check
@@ -15,7 +32,15 @@ class ConfigApplied(BaseCheck):
             'unknown',
         ] or not hasattr(self.related_object, 'config'):
             return
-        result = int(self.related_object.config.status == 'applied')
+        config_status = self.related_object.config.status
+        config_modified_datetime = self.related_object.config.modified.astimezone(
+            timezone(settings.TIME_ZONE)
+        )
+        time_now_diff = (
+            datetime.now(tz=timezone(settings.TIME_ZONE)) - config_modified_datetime
+        )
+        config_modified_mins_ago = round(time_now_diff.total_seconds() / 60)
+        result = self._get_result(config_status, config_modified_mins_ago)
         # If the device config is in error status we don't need to notify
         # the user (because that's already done by openwisp-controller)
         # but we need to ensure health status will be changed
--- a/openwisp_monitoring/check/classes/config_applied.py
+++ b/openwisp_monitoring/check/classes/config_applied.py
@@ -1,7 +1,6 @@
-from datetime import datetime
+from datetime import timedelta
 
-from django.conf import settings
-from pytz import timezone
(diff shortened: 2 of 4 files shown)


Aryamanz29:
WIP

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -162,15 +162,15 @@ def test_config_modified_device_problem(self):
...
         dm = d.monitoring
         dm.refresh_from_db()
-        self.assertEqual(m.is_healthy, False)
-        self.assertEqual(m.is_healthy_tolerant, False)
-        self.assertEqual(dm.status, 'problem')
-        self.assertEqual(Notification.objects.count(), 1)
+        self.assertEqual(m.is_healthy, True)
+        self.assertEqual(m.is_healthy_tolerant, True)
+        self.assertEqual(dm.status, 'ok')
+        self.assertEqual(Notification.objects.count(), 0)
```
nemesisdesign, line 173: These changes do not make sense. The name of the test is `test_config_modified_device_problem`, so the test code must be changed to cause a problem.

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -206,7 +206,7 @@ def test_config_error(self):
             check.perform_check()
         dm.refresh_from_db()
         self.assertEqual(dm.status, 'ok')
-        self.assertEqual(Notification.objects.filter(actor_object_id=m.id).count(), 1)
+        self.assertEqual(Notification.objects.filter(actor_object_id=m.id).count(), 0)
```
nemesisdesign, line 209: same here

nemesisdesign (changes requested):
It seems to me that this is offtrack.

The first thing we should do is to write a failing test for the bug. Keep this in mind whenever we are fixing bugs. Rule number 1: replicate the bug with a new failing test. This test will have to pass after the bug is fixed.

I do not see any new test, I see instead old tests which are changed in a way which does not make sense (see my comments below).

Please as first thing add a failing test and then address the comments below. Make sure the new test fails without the fix.

openwisp_monitoring/check/tests/test_models.py
Aryamanz29: You're right, I'll update that, thanks :+1:

openwisp_monitoring/check/tests/test_models.py
Aryamanz29: Updated

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -223,14 +223,14 @@ def test_config_check_critical_metric(self):
...
-        self.assertTrue(dm.is_metric_critical(m))
+        m = Metric.objects.get(configuration='config_applied')
+        self.assertEqual(dm.is_metric_critical(m), True)
         # must be executed twice to trepass the tolerance
         with freeze_time(now() - timedelta(minutes=6)):
             check.perform_check()
         check.perform_check()
         dm.refresh_from_db()
-        self.assertEqual(dm.status, 'critical')
+        self.assertEqual(dm.status, 'ok')
```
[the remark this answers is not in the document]
Aryamanz29: updated

openwisp_monitoring/check/classes/config_applied.py
```diff
@@ -15,7 +32,15 @@ def check(self, store=True):
...
-        result = int(self.related_object.config.status == 'applied')
+        config_status = self.related_object.config.status
+        config_modified_datetime = self.related_object.config.modified.astimezone(
+            timezone(settings.TIME_ZONE)
+        )
+        time_now_diff = (
+            datetime.now(tz=timezone(settings.TIME_ZONE)) - config_modified_datetime
+        )
+        config_modified_mins_ago = round(time_now_diff.total_seconds() / 60)
+        result = self._get_result(config_status, config_modified_mins_ago)
```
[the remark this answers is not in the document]
Aryamanz29: I have used this below condition :
```py
int(self.related_object.config.status == 'applied' or self.related_object.modified > timezone.now() - timedelta(minutes=CONFIG_CHECK_INTERVAL))
```

openwisp_monitoring/check/settings.py
```diff
@@ -19,3 +19,4 @@
 )  # 10 minutes arbitrarily chosen (must be longer than TCP + UDP test time)
 IPERF3_CHECK_DELETE_RSA_KEY = get_settings_value('IPERF3_CHECK_DELETE_RSA_KEY', True)
 CHECKS_LIST = get_settings_value('CHECK_LIST', list(dict(CHECK_CLASSES).keys()))
+CONFIG_CHECK_INTERVAL = get_settings_value('CONFIG_CHECK_INTERVAL', 5)  # in minutes
```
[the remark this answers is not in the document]
Aryamanz29: - Yes, we should convert it to `int` explicitly to avoid errors. :+1: 
- I have updated the `README`.

openwisp_monitoring/check/tests/test_models.py
```diff
@@ -274,3 +283,28 @@ def test_device_unknown_no_config_check(self):
...
         self.assertEqual(Notification.objects.count(), 0)
         self.assertIsNone(c2.perform_check())
+
+    def test_config_check_problem_with_interval(self):
+        self._create_admin()
+        d = self._create_device(organization=self._create_org())
+        self._create_config(device=d)
+        check = Check.objects.filter(check_type=self._CONFIG_APPLIED).first()
+        self.assertEqual(Check.objects.count(), 3)
+        self.assertEqual(Metric.objects.count(), 0)
```
nemesisdesign, line 293: I have tried to run this test without the changes above and it fails here, which doesn't make sense, please make sure the test fails at the right place without the fix.

README.rst
```diff
@@ -1488,6 +1488,18 @@ validating custom parameters of a ``Check`` object.
 This setting allows you to choose whether `config_applied <#configuration-applied>`_ checks should be
 created automatically for newly registered devices. It's enabled by default.
 
+``OPENWISP_MONITORING_CONFIG_CHECK_INTERVAL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 1492: the length of this must match the line above

nemesisdesign (changes requested):
Thank you @Aryamanz29! :+1: 
See my comments below.

README.rst
Aryamanz29: oops, I'll fix that :+1:

openwisp_monitoring/check/tests/test_models.py
Aryamanz29: I find out that when we run specific tests, the general metrics don't exist, so it fails at line `293`; when we run all class tests together, it fails at the right place. This behaviour can be seen all over `test_models.py`, where we assert `Metric.objects.count()`. For now, I have changed the assert condition for this test case so that it fails at the right place.

nemesisdesign (approved):

PR #469, opened 2022-12-07
[fix] Fixed blip caused by periodic chart reload #455

### LINKED ISSUE #455: [bug:ux] Periodic reloading of chart data causes blip
The UX right now has a very naive reloading of the chart data every X minutes which causes the charts to disappear completely.

It would be great to avoid this and allow the chart data to reload without hiding the charts, then replace che old charts with the new charts right away.

Aryamanz29:
Fixes #455

Checks:


nemesisdesign (approved):
I spotted some code duplication that I didn't see before, so I took advantage of this chance to remove it, please double check my changes:
https://github.com/openwisp/openwisp-monitoring/pull/469/commits/5c82c5e0d930701fc9dcb3c80cad53c9b5879d70
The rest looks good and it works well :+1: .

Aryamanz29 (commented):
LGTM thanks :+1:

nemesisdesign (approved):

PR #472, opened 2022-12-23
[feature] Allowed specifying additional options for InfluxDB backend (UDP) #458

### LINKED ISSUE #458: [change] Find faster way to write to influxdb (eg: UDP)
We should find a faster method for writing timeseries data, maybe UDP instead of HTTP.
I think this may change from influxdb 1.8 to 2.0 so we'd have to check this on Influxdb 2.0 #274.
Right now writing timeseries on big systems is quite a bottleneck!

See also https://github.com/openwisp/openwisp-monitoring/issues/466.

pandafy:
- Additional options can be specified to the TIMESERIES_DATABASE setting which are passed to the underlying backend library. This allows using UDP for writing to InfluxDB.
- "openwisp_monitoring.db.backends.influxdb.client.DatabaseClient.write" will use "InfluxDBClient.write_points" instead of "InfluxDBClient.write" because the former allows writing data using the UDP protocol.

Closes #458

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1479,6 +1479,12 @@ section of your InfluxDB should look similar to the following:
     database = "openwisp2"
     retention-policy = 'short'
 
+If you are using `ansible-openwisp2 <https://github.com/openwisp/ansible-openwisp2>`_
+for deploying OpenWISP, you can set the ``influxdb_udp_mode`` ansible variable to ``true``
+in your playbook, this will make the ansible role automatically configure the InfluxDB UDP listeners.
+You can refer to `ansible-ow-influxdb <https://github.com/openwisp/ansible-ow-influxdb#role-variables>`
+(a dependency of ansible-openwisp2) documentation to learn more.
+
 ``OPENWISP_MONITORING_DEFAULT_RETENTION_POLICY``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/README.rst
+++ b/README.rst
@@ -1424,7 +1424,8 @@ Settings
         'PORT': '8086',
         'OPTIONS': {
             'udp_writes': False,
-            'udp_port': 8089
+            'udp_port': 8089,
+        }
     }
 
 The following table describes all keys available in ``TIMESERIES_DATABASE``
--- a/README.rst
+++ b/README.rst
@@ -1482,7 +1482,7 @@ section of your InfluxDB should look similar to the following:
 If you are using `ansible-openwisp2 <https://github.com/openwisp/ansible-openwisp2>`_
 for deploying OpenWISP, you can set the ``influxdb_udp_mode`` ansible variable to ``true``
 in your playbook, this will make the ansible role automatically configure the InfluxDB UDP listeners.
-You can refer to `ansible-ow-influxdb <https://github.com/openwisp/ansible-ow-influxdb#role-variables>`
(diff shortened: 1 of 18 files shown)


pandafy:
I have found some caveats with using UDP for writing data to the InfluxDB: 

- UDP is used only for writing data to the InfluxDB. It requires adding additional configuration to the InfluxDB config. 
- It does not supports authentication. 
- It allows using only one "retention policy" that too has to be specified in the InfluxDB configuration.

nemesisdesign (changes requested):
I have mixed feelings about this, I do not feel comfortable in merging such a change while ignoring the test failures. See below.

I think the results which are coming out of the tests are encouraging on the perfomance side, but on the functionality side there's more work to do to merge this safely.

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -244,3 +245,19 @@ def _create_chart(self, metric=None, test_data=True, configuration='dummy'):
         c.full_clean()
         c.save()
         return c
+
+    def _read_chart_or_metric(self, obj, *args, **kwargs):
+        if TIMESERIES_DB.get('OPTIONS', {}).get('udp_writes', False):
```
nemesisdesign, line 250: avoid repetition of this by creating a function that can be reused across the tests

openwisp_monitoring/monitoring/tests/test_monitoring_notifications.py
```diff
@@ -127,10 +128,12 @@ def test_general_check_threshold_crossed_for_long_time(self):
             self.assertEqual(m.is_healthy_tolerant, True)
             self.assertEqual(Notification.objects.count(), 0)
 
+        time.sleep(0.1)
```
nemesisdesign, line 131: make sure this sleep is executed only when in UDP mode and use the function mentioned above

tests/openwisp2/settings.py
```diff
@@ -31,7 +31,10 @@
     'OPTIONS': {'udp_writes': True, 'udp_port': 8089},
```
nemesisdesign, line 31: Do not default to UDP as that will cause issues with new developers trying this module out. Default on TCP and enable UDP in the tests.

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Please also update the docs where it mentions the various timeseries retry options to make it explicit that it doesn't apply when using UDP mode.

I also see no mention of this new feature in the README and that's not good.

nemesisdesign (approved):
Let's test this more before merging.

README.rst
```diff
@@ -1397,6 +1404,81 @@ these permissions are included by default in the "Administrator" and "Operator"
...
+    enabled = true
+    bind-address = "127.0.0.1:8089"
+    database = "openwisp2"
+
+    # For writing data with the "short" retention policy
+    [[udp]]
+    enabled = true
+    bind-address = "127.0.0.1:8090"
+    database = "openwisp2"
+    retention-policy = 'short'
```
nemesisdesign, line 1480: Please mention that when using ansible-openwisp2 this can be done easily by setting the new var addded in https://github.com/openwisp/ansible-ow-influxdb/pull/17 to `true` and [link the role variables section of the ansible-ow-influxdb repo for reference](https://github.com/openwisp/ansible-ow-influxdb#role-variables).
This should help users figure out how to use this feature autonomously.

nemesisdesign (changes requested):
Ok great one docs change requested below.

nemesisdesign (approved):

PR #473, opened 2022-12-27
[change] Batch writes in Device metrics REST API #466

### LINKED ISSUE #466: [change] Batch timeseries writes
It should be possible to batch timeseries writes and instead of writing each metric singularly find a way to batch them and hence each call to the monitoring API view to write data would generate only 1 celery task instead of many, and that celery task would then write to influxdb only once using the fastest method available (#458).

pandafy:
Closes #466

Checks:


nemesisdesign (changes requested):
Good progress @pandafy! See my comments below.

nemesisdesign (approved):

PR #480, opened 2022-12-29
[feature] Added monitoring device list API

Aryamanz29:
- Added `/api/v1/monitoring/device` (GET, POST) by extending [`ow-controller device REST API`](https://github.com/openwisp/openwisp-controller/blob/fdffd1c8410ddbc7b30bfdf05defe21c9d1ac8a5/openwisp_controller/config/api/views.py#L80-L90).

Screenshot from 2022-12-29 16-05-03

Todos : 

- [ ]  Need to add `/api/v1/monitoring/device/<device:pk>` (device monitoring detail).
- [ ] Need to add tests.

Checks:


Aryamanz29:
WIP

Aryamanz29:
@nemesisdesign The [test build is failing](https://github.com/openwisp/openwisp-monitoring/actions/runs/4313182540/jobs/7524662095#step:12:170) because of the import of `DeviceListSerializer` in https://github.com/openwisp/openwisp-monitoring/pull/480/files#diff-650f919640f11d6704239f2d9ab34ab214e824407e621f83d98667f209b19115R4
```py
from openwisp_controller.config.api.serializers import DeviceListSerializer
```
which further imports
```py
from ..admin import BaseConfigAdmin
# and then
from openwisp_users.admin import OrganizationAdmin
```
which in result unregisters email and social apps of `openwisp-users` multiple times
 
https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/admin.py#L647-L656

and raises `django.contrib.admin.sites.NotRegistered` in https://github.com/openwisp/openwisp-monitoring/actions/runs/4313182540/jobs/7524662095#step:12:170

I tried to change the order of installed apps, but it didn't work. Another way is to move the [download API views](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/api/views.py#L71-L77) of the **config app** to a different python file since only those views depend on `BaseConfigAdmin`.

If there any other way we can fix this issue please let me know.

Aryamanz29:
**Todos**

- [ ]  Fix current failing build. 
- [ ] Add API tests and update docs.

README.rst
```diff
@@ -2500,13 +2500,25 @@ The format used for Device Status is inspired by
...
 
+List device monitoring information
+##################################
+
+.. code-block:: text
+
+    GET /api/v1/monitoring/device/
+
+Response contains device information, monitoring status, and a list of metrics with their respective statuses.
+
```
nemesisdesign, line 2521: We need another section which explains how to fetch the device data and metrics, which also explains the endpoint can be accessed with session authentication, token authentication or alternatively with the device key (this method is meant to be used by the devices).

We must mention that the list view does not include metrics details while the detail view does.

nemesisdesign (changes requested):
The list page generates too many queries, look at what happens if I increase the `page_size` argument to display more items in one page:

Screenshot from 2023-03-06 19-46-08

Look at the right, you see those 57 queries? Not good.

To enable `debug_toolbar` install it in your python virtualenv and then add this to the `settings.py`:

```python

settings.INSTALLED_APPS += [
    'debug_toolbar'
]

settings.MIDDLEWARE += [
    'debug_toolbar.middleware.DebugToolbarMiddleware'
]

INTERNAL_IPS = ['127.0.0.1']
```

Now back to the real issue. I think we cannot list the metric details in the list page without generating 1 query for each device. While the monitoring status can be fetched without an extra query by using `select_related()` (which translates to SQL as a JOIN query), the metrics details will need an extra query for each device, so in the list I would remove this information.

The docs and view description must mention these details.

openwisp_monitoring/device/api/views.py
```diff
@@ -129,10 +153,20 @@ class MonitoringGeoJsonLocationList(GeoJsonLocationList):
...
+    serializer_class = MonitoringLocationDeviceSerializer
 
     def get_queryset(self):
         return super().get_queryset().select_related('monitoring').order_by('name')
 
 
 monitoring_location_device_list = MonitoringLocationDeviceList.as_view()
+
+
+class MonitoringDeviceList(DeviceListCreateView):
```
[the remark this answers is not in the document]
Aryamanz29: `DeviceListCreateView` is a derived class of `ProtectedAPIMixin` (includes session & bearer auth) and `ListCreateAPIView`  and device key is not required for listing devices with monitoring info.

openwisp_monitoring/device/api/views.py
```diff
@@ -38,18 +45,35 @@
...
 
+    def get_permissions(self):
+        if self.request.method in SAFE_METHODS:
+            self.permission_classes = ProtectedAPIMixin.permission_classes
+        return super().get_permissions()
+
+    def get_authenticators(self):
+        if self.request.method in SAFE_METHODS:
+            self.permission_classes = ProtectedAPIMixin.authentication_classes
+        return super().get_authenticators()
```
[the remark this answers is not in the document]
Aryamanz29: We need different permissions for different methods and scenarios; for example, when making a GET request, if the key query parameter is present, then `DevicePermission` should be our permission class; if it's not present, then `ProtectedAPIMixin.permission_classes`; and for a POST request, we only use `DevicePermission` as our permission class.
Yes `SessionAuthentication` and `BearerAuthentication` are supported when the device key is absent in the GET request.
Sure, I have removed the key query parameters from the frontend calls; should I remove them from the unit tests as well?

README.rst
Aryamanz29: sure

Aryamanz29:
@nemesisdesign I have a question:

For `api/v1/monitoring/device/` : We only want to list devices with monitoring information, right? So for that, we need to restrict the `DeviceListCreateView` of `openwisp-controller` to allow only **GET, HEAD, and OPTIONS** (safe methods) only.

Aryamanz29:
#### Added a monitoring status filter to the `MonitoringDeviceList` api view

Screenshot from 2023-03-13 23-25-36

Aryamanz29:
#### Tested locally

Screenshot from 2023-03-14 00-12-30

Screenshot from 2023-03-14 00-13-53

nemesisdesign (changes requested):
Tests keep failing also after fixing the issue in openwisp-users: https://github.com/openwisp/openwisp-users/pull/345. 
I get the same failures locally.

Something in these changes is screwing the tests.

Aryamanz29:
@nemesisdesign https://github.com/openwisp/openwisp-users/pull/345 only fixes the import from `openwisp_users.admin import OrganizationAdmin`, which was causing multiple unregistering of Django apps. However, the test build is still failing due to an import of `BaseConfigAdmin` used in the `download_api_views` of the config app which breaks `openwisp-monitoring` admin tests. 

To address this issue, I have moved those views to a separate file in OpenWISP/OpenWISP-controller#750.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/master
+openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/move-config-download-api-views
```
nemesisdesign, line 1:
```suggestion
openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/master
```

nemesisdesign (approved):

README.rst
```diff
@@ -2523,23 +2524,25 @@ List device monitoring information
...
+**Available filters**
+
+Data can be filtered by health status (e.g. critical, ok, problem, and unknown)
+to obtain the list of devices in the corresponding status, for example,
+to retrieve the list of devices which are in critical conditions
+(eg: unreachable), the following will work:
+
+.. code-block:: text
+
+   GET /api/v1/monitoring/device/?monitoring__status=ok
```
Aryamanz29, line 2545: @nemesisdesign for critical (unreachable) condition, I think it should be : `?monitoring__status=critical`

nemesisdesign: ups sorry, I was too in a hurry, thanks!

nemesisdesign (approved):

Issue #484, opened 2022-12-29
[change] Move device API view writing logic to background worker


nemesisdesign:
I have been thinking about a possible change which would reduce the load on the HTTP application server and move it on the celery workers, this would allow OpenWISP to be able to process more HTTP requests per second, but we have to test this assumption with some measurements.

I think the view should limit itself to perform only the validation of the incoming payload:

https://github.com/openwisp/openwisp-monitoring/blob/2e153a247ef0f3967ff5bd7fe8eb0b4137755f00/openwisp_monitoring/device/api/views.py#L67-L100

I think the logic in `self._write(request, self.instance.pk, time=time)` can be moved to a celery task (which should have a default time limit).

Right now that instruction is wrapper in a try/except block, which if it fails it returns a 400 bad request error, we won't be able to do that if we move this instruction to a celery worker, in the case we catch any error there we should simply log the error and later on devise changes which will allow us to prevent these errors when possible.

PR #490, opened 2023-01-16
[change] Moved device API view writing logic to background worker #484 

pandafy:
Closes https://github.com/openwisp/openwisp-monitoring/issues/484

### DIFF
--- a/openwisp_monitoring/device/writer.py
+++ b/openwisp_monitoring/device/writer.py
@@ -19,6 +19,13 @@ logger = logging.getLogger(__name__)
 
 
 class DeviceDataWriter(object):
+    """
+    This class is in charge of writing the device metric data.
+    Before these methods were shipped in the REST API view
+    but later have been moved here to allow writing this data in
+    the background processes of OpenWISP.
+    """
+
     def __init__(self, device_data):
         self.device_data = device_data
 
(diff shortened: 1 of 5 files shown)


openwisp_monitoring/device/writer.py
```diff
@@ -0,0 +1,398 @@
...
+
+Chart = load_model('monitoring', 'Chart')
+Metric = load_model('monitoring', 'Metric')
+AlertSettings = load_model('monitoring', 'AlertSettings')
+Device = load_model('config', 'Device')
+
+logger = logging.getLogger(__name__)
+
+
+class DeviceDataWriter(object):
```
nemesisdesign, line 21: Let's add a description here. Something like:

This class is in charge of writing the device metric data.
Before these methods were shipped in the REST API view
but later have been moved here to allow writing this data in
the background processes of OpenWISP.

nemesisdesign (changes requested):
It's looking good and working fine in my dev env, see my comments below but let's handle and merge https://github.com/openwisp/openwisp-monitoring/pull/473 first.

.github/workflows/ci.yml
```diff
@@ -6,10 +6,12 @@ on:
     branches:
       - master
       - dev
+      - issues/466-batch-timeseries-write
```
nemesisdesign, line 9: I think we can remove this now

nemesisdesign (changes requested):
Conflicts appeared after merging #473.

nemesisdesign (approved):

PR #496, opened 2023-01-23
[change] Added default timeout to celery tasks #474

### LINKED ISSUE #474: [change] Add timeout to all celery tasks
Let's ensure all tasks have a reasonable timeout.

Some tasks like `run_checks` must have a higher timeout (eg: 1 hour), we should avoid hitting this under all conditions, all the other tasks should have much shorter timeout but not too short.

pandafy:
Closes #474

Checks:

### DIFF
--- a/openwisp_monitoring/check/tasks.py
+++ b/openwisp_monitoring/check/tasks.py
@@ -7,6 +7,8 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured, ObjectDoesNotExist
 from swapper import load_model
 
+from openwisp_utils.tasks import OpenwispCeleryTask
+
 from .settings import CHECKS_LIST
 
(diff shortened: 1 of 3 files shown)


openwisp_monitoring/check/tasks.py
```diff
@@ -16,7 +18,7 @@ def get_check_model():
     return load_model('check', 'Check')
 
 
-@shared_task
+@shared_task(time_limit=60 * 60)
```
nemesisdesign, line 21: please increase to 2 hours just in case we hit this limit, in the future we'll probably have to make this configurable for very large instances... unless this 2 hours limit never gets hit, but if it does it will be a great problem to have! :grin: but we can deal with it later so for now just increase it to 2 hours.

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #499, opened 2023-02-03
[change] Close WiFi Session respecting tolerance of critical metrics #493

### LINKED ISSUE #493: [change] If VPN link is unstable, wifi sessions are closed and opened all the time
I spotted this on a couple of instances: an unstable management VPN link fails often, wifi sessions are flagged as closed and then reopen, so in 1 day there's hundred of wifi sessions for the same client.

Maybe we should trigger the wifi close operation using the threshold crossed signal instead of using the health change signal (as it is programmed now), but we have to make sure this signal is sent by the system also when alerts are disabled, which is something I am not 100% sure of and we have to double check.

But definitely we need to make sure to have a way to trigger these kind of operations with a certain degree of tolerance to avoid this annoying flapping which can also add useless burden to the system.

pandafy:
WiFi sessions an of offline devices are closed when the threshold of device's critical metrics (ping) are crossed for longer than the tolerance set in metric's alert settings.

Closes #493

Checks:

### DIFF
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -20,6 +20,7 @@ from openwisp_utils.admin_theme import (
 from openwisp_utils.admin_theme.menu import register_menu_subitem
 
 from ..check import settings as check_settings
+from ..monitoring.signals import threshold_crossed
 from ..settings import MONITORING_API_BASEURL, MONITORING_API_URLCONF
 from ..utils import transaction_on_commit
 from . import settings as app_settings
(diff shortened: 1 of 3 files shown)


nemesisdesign (changes requested):
Unfortunately after testging this solution the conclusion is that this solution is not effective in solving the problem, sessions are still closed and reopened all the time when the management tunnel flaps.

More investigation is needed here.

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #505, opened 2023-03-17
[change] Renamed daterangepicker widget labels #487

### LINKED ISSUE #487: [change] Shall we rename last week, last month, last year?
In many popular analytics visualization software, these phrases have different meanings than how we use it here.
I think my initial interpretation of these was wrong.
We may have to change this to last 7 days, last 30 days and last 365 days to avoid misunderstandings.
Any thoughts?

Aryamanz29:
Screenshot from 2023-03-17 17-00-42

Renamed date range widget label from last week, month and year to 7 days, 30 days and 365 days to avoid any misunderstandings in the future.

Closes #487

Checks:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,7 +95,7 @@ Available Features
 * Maintains a record of `WiFi sessions <#monitoring-wifi-sessions>`_ with clients'
   MAC address and vendor, session start and stop time and connected device
   along with other information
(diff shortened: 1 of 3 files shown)


nemesisdesign (changes requested):
Please flag the strings as translatable using `gettext()`, which should be already loaded in the global namespace (please double check).

nemesisdesign (approved):
Thanks :+1:

Issue #350, opened 2022-04-07
[feature] Add wifi version capability to wifi interface status


nemesisdesign:
In the device status, we're able to see different WiFi info like channel, frequency, noise, etc, but it's not evident whether that interface is working with in High Throughput mode (802.11n/wifi4), Very High Throughput mode (802.11ac/wifi5) or High Efficiency (802.11ax/wifi6).

It would be great to add this information.

Aryamanz29:
To add wifi version capability to wifi interface status, I think first we need to add the `wifi-device` -> [`htmode`](https://github.com/openwisp/openwrt-openwisp-monitoring/search?q=htmode) option to [`netjson-monitoring`](https://github.com/openwisp/openwrt-openwisp-monitoring/blob/7001e772395212e76c0279b53315a7bb634aedb5/openwisp-monitoring/files/sbin/netjson-monitoring.lua#L85-L94) and then update the code here.

Ref : https://openwrt.org/docs/guide-user/network/wifi/basic#wi-fi_devices

```diff
diff --git a/openwisp-monitoring/files/sbin/netjson-monitoring.lua b/openwisp-monitoring/files/sbin/netjson-monitoring.lua
index 7965769..6190746 100755
--- a/openwisp-monitoring/files/sbin/netjson-monitoring.lua
+++ b/openwisp-monitoring/files/sbin/netjson-monitoring.lua
@@ -90,7 +90,8 @@ for _, radio in pairs(wireless_status) do
           tx_power = iwinfo.txpower,
           signal = iwinfo.signal,
           noise = iwinfo.noise,
-          country = iwinfo.country
+          country = iwinfo.country,
+          htmode = radio.config.htmode
         }
       }
       if iwinfo.mode == 'Ad-Hoc' or iwinfo.mode == 'Mesh Point' then
```

nemesisdesign:
@Aryamanz29 yes but `htmode` should be contained in the `wireless` section.

PR #506, opened 2023-04-04
[feature] Flagged the `HT/VHT` field of WiFi clients as null #351

### LINKED ISSUE #351: [feature] If not using HT/VHT, flag the HT/VHT field of WiFi clients as null
Depends on https://github.com/openwisp/openwisp-monitoring/issues/350.

If not using HT/VHT, flag the HT/VHT field of WiFi clients as null and show it as with the unkonwn gray question mark icon instead of the red x icon.

This is because if, for example, VHT is not used, then all the clients will have the VHT flag set as False, which may not be true, some clients may be VHT capable but VHT is not enabled at the radio level, hence is shown as false, but it's misleading.

Aryamanz29:
#### `Device`
Screenshot from 2023-04-04 17-17-14

#### `Wifi Session Inline`
Screenshot from 2023-04-04 17-17-05

#### `Wifi Session List`
Screenshot from 2023-04-04 16-00-03

#### `Wifi Session Change`
Screenshot from 2023-04-04 15-59-20

Closes #351

Checks:


openwisp_monitoring/device/base/models.py
```diff
@@ -183,6 +183,14 @@ def _transform_data(self):
                 for signal_key, signal_values in interface['mobile']['signal'].items():
                     for key, value in signal_values.items():
                         signal_values[key] = float(value)
+            # If HT/VHT is not being used, set the HT/VHT field
+            # of WiFi clients to None. This is necessary because
+            # some clients may be VHT capable but VHT is not enabled
+            # at the radio level, which can lead to false information.
```
nemesisdesign, line 189:
```suggestion
            # at the radio level, which can mislead into thinking the client is not HT/VHT capable.
```

openwisp_monitoring/device/base/models.py
```diff
@@ -183,6 +183,15 @@ def _transform_data(self):
...
                         signal_values[key] = float(value)
+            # If HT/VHT is not being used, set the HT/VHT
+            # field of WiFi clients to None. This is necessary
+            # because some clients may be VHT capable but VHT is
+            # not enabled at the radio level, which can mislead
+            # into thinking the client is not HT/VHT capable.
+            if 'wireless' in interface and 'clients' in interface['wireless']:
+                for client in interface['wireless']['clients']:
+                    client['ht'] = client['ht'] if client['ht'] else None
+                    client['vht'] = client['vht'] if client['vht'] else None
```
nemesisdesign, line 194: This code doesn't look right.
Are you sure this code implements what is requested in the issue description?

> Depends on https://github.com/openwisp/openwisp-monitoring/issues/350.
> 
> If not using HT/VHT, flag the HT/VHT field of WiFi clients as null and show it as with the unkonwn gray question mark icon instead of the red x icon.
> 
> This is because if, for example, VHT is not used, then all the clients will have the VHT flag set as False, which may not be true, some clients may be VHT capable but VHT is not enabled at the radio level, hence is shown as false, but it's misleading.

It's being asked to show "None" if the interface is not using either HT or VHT.

nemesisdesign (changes requested):

openwisp_monitoring/device/base/models.py
Aryamanz29: Sorry, I misunderstood the issue description. Therefore, I believe we need to update this implementation to use the radio config `htmode`. If that flag is set to `NOHT`, it means that `HT` and `VHT` are disabled at the radio level. In that case, we should set the `HT` and `VHT` fields to `None`, which would result in showing an unknown icon on the admin.

Aryamanz29:
I think there is some bug from Django's end. Getting a [similar error](https://github.com/openwisp/openwisp-monitoring/actions/runs/4731933794/jobs/8397454138?pr=506) on the master branch too.

```bash
(env) ➜  tests git:(master) py manage.py runserver
from django.conf import DEFAULT_STORAGE_ALIAS, settings
ImportError: cannot import name 'DEFAULT_STORAGE_ALIAS' from 'django.conf' (/openwisp-monitoring/env/lib/python3.8/site-packages/django/conf/__init__.py)
```

nemesisdesign (changes requested):

.github/workflows/ci.yml
```diff
@@ -78,6 +78,7 @@ jobs:
       - name: Install openwisp-monitoring
         run: |
           pip install -U -I -e .
+          pip uninstall -y django
           pip install ${{ matrix.django-version }}
```
pandafy, line 82:
```suggestion
          pip install --force-reinstall ${{ matrix.django-version }}
```

I would rather use this one-liner 😄

pandafy: I see that you already tried this.

This is bizarre though. The [logs of failed CI buid](https://github.com/openwisp/openwisp-monitoring/actions/runs/4752709040/jobs/8443391378#step:10:406) shows that correct version of Django was installed 
```
Successfully installed asgiref-3.6.0 django-3.2.18 pytz-2023.3 sqlparse-0.4.4
```

Aryamanz29: @pandafy : Please see CI build status on this change [`f5769bd` (#506)](https://github.com/openwisp/openwisp-monitoring/pull/506/commits/f5769bdcf5015faef209e23d88d551d039c53fd7)

CI build status: https://github.com/openwisp/openwisp-monitoring/actions/runs/4752709040

nemesisdesign (approved):
Great, thanks! :+1:

nemesisdesign (changes requested):
@Aryamanz29 please squash the commits except the CI fix which is better to show as a separate commit, then we can merge.

openwisp_monitoring/device/base/models.py
```diff
@@ -183,6 +183,27 @@ def _transform_data(self):
...
+                    # NOHT : disables 11n (ie. ht and vht both are False)
+                    if interface['wireless']['htmode'] == 'NOHT':
+                        client['ht'] = None
+                        client['vht'] = None
+                    # If only HT is enabled, set client vht to None
+                    elif (
+                        interface['wireless']['htmode'].startswith('HT')
+                        and not client['vht']
+                    ):
+                        client['vht'] = None
```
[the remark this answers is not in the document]
nemesisdesign: It's correct, because if VHT is not enabled at interface level, OpenWrt will report VHT to be false even if the client is VHT capable.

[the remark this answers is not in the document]
pandafy: Thank you for the clarification. 😄

PR #507, opened 2023-04-11
[fix] Fixed wifi session tabular inline error #504

### LINKED ISSUE #504: [bug] WifiSession tabular inline may throw error innecessarily
This bug affects only the development version (current master branch).

If the WiFi session information of a device changes while the user is modifying any detail of the device, the system will throw a validation error:

> (Hidden field TOTAL_FORMS) This field is required.
> (Hidden field INITIAL_FORMS) This field is required.
> 
> ManagementForm data is missing or has been tampered with. Missing fields: wifisession_set-TOTAL_FORMS, wifisession_set-INITIAL_FORMS. You may need to file a bug report if the issue persists.

This happens because the WiFi session information is a form.

We have to rewrite this part to avoid using a real inline and simulate it somehow while building the same output without a form.
This will allow us to get rid of this nasty error.

Aryamanz29:
If wifi sessions are created in the background while the user is still on the device change page, saving the device object can raise a formset error. To prevent this, we can check if wifi sessions exist before saving the device object and provide default values for the complaining formset fields.

Fixes #504

Checks:

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -275,6 +275,22 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
 
     health_status.short_description = _('health status')
 
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        device = self.get_object(request, object_id)
+        if device and device.wifisession_set.exists():
+            # We need to provide default formset values
+            # to avoid management formset errors when wifi sessions
+            # are created while editing the DeviceAdmin change page
+            wifisession_formset_data = {
+                'wifisession_set-TOTAL_FORMS': '1',
+                'wifisession_set-INITIAL_FORMS': '1',
+            }
+            request.POST = request.POST.copy()
+            request.POST.update(wifisession_formset_data)
+        return super().change_view(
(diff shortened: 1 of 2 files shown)


openwisp_monitoring/device/admin.py
```diff
@@ -275,6 +275,22 @@ def health_status(self, obj):
 
     health_status.short_description = _('health status')
 
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        device = self.get_object(request, object_id)
```
nemesisdesign, line 279: This will add a duplicate query to the view generation.
See how this query is repeated later on:
https://github.com/django/django/blob/ef62a3a68c9d558486145a42c0d71ea9a76add9e/django/contrib/admin/options.py#L1769

It should be possible to avoid this.
There's several ways. What about overriding get_object instead? That way we would also avoid other issues.
For example, with this solution I am pretty sure that if you try to access a device with a valid UUID but which does not exist an uncaught exception will be raised.

nemesisdesign (changes requested):
Great progress @Aryamanz29, see below.

nemesisdesign (approved):

PR #516, opened 2023-05-19
[change] Reduced network calls in device metric flow #501

### LINKED ISSUE #501: [change] Reduce network calls in device metric flow
I think that when using UDP mode it does not make sense to keep executing the timeseries write tasks asynchronously.
The main advantage of writing data asynchronously is that if it fails it can be retried, but when usind influxdb in UDP mode we do not get a reply from Influxdb and hence this feature loses meaning, but UDP increases efficiency a lot so it really makes sense on large instances.
Right now on the latest dev version, all metric writing is already performed in the background, I believe avoiding to create an additional celery taks to write metrics and just do it synchronously (**when in UDP mode**) will improve things further because it will reduce overhead.

Moreover, in the device metric view and task the logic which fetches the device and the charts could be cached so we would avoid hitting the DB every 5 minutes multiplied by the number of devices.

Also we didn't think about it, but now that we moved the device write logic to the background, we repeat the DB query to fetch the device one more time (one time in the view and one time in the celery task), so it's 2 * <number-of-devices>.

These changes would allow to reduce the amount of expensive calls significantly:

- 1 call to redis to pass the device data for saving it
- 1 call to redis to pass the device data again for writing wifi sessions
- less memory used by redis, it stays in the application a little longer and gets thrown away, instead of being passed to redis, thrown away, fetched in another process, used and thrown away again
- celery workers wouldn't have to process 2 additional tasks
- 2 calls to postgres to fetch the device instance, which we can store in the cache (default duration 1 day)
- several calls to fetch the chart instances which we can also store in the cache (default duration 1 day)
- moreover, also the checks (ping, config check, iperf) shall write the timeseries data synchronously when in UDP mode

All these saved operations have to be multiplied by the number of devices and every 5 minutes, I believe this is going to be significant but again we have to measure it to be 100% sure.

pandafy:
Closes #501

Checks:

### DIFF
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -25,6 +25,7 @@ from openwisp_controller.geo.api.views import (
 )
 from openwisp_users.api.mixins import FilterByOrganizationManaged
 
+from ...settings import CACHE_TIMEOUT
 from ...views import MonitoringApiViewMixin
 from ..schema import schema
 from ..signals import device_metrics_received
@@ -66,6 +67,10 @@ def get_device_args_rewrite(view, pk):
     return pk.hex
 
 
+def get_charts_args_rewrite(view, request, pk):
+    return (pk,)
+
+
 class DeviceMetricView(MonitoringApiViewMixin, GenericAPIView):
     """
     Retrieve device information, monitoring status (health status),
@@ -95,6 +100,16 @@ class DeviceMetricView(MonitoringApiViewMixin, GenericAPIView):
         view.get_object.invalidate(view, str(instance.pk))
         logger.debug(f'invalidated view cache for device ID {instance.pk}')
 
+    @classmethod
+    def invalidate_get_charts_cache(cls, instance, *args, **kwargs):
+        if isinstance(instance, Device):
+            pk = instance.id
+        elif isinstance(instance, Metric):
+            pk = instance.object_id
+        elif isinstance(instance, Chart):
+            pk = instance.metric.object_id
+        cls._get_charts.invalidate(None, None, pk)
+
     def get_permissions(self):
         if self.request.method in SAFE_METHODS and not self.request.query_params.get(
             'key'
@@ -123,6 +138,7 @@ class DeviceMetricView(MonitoringApiViewMixin, GenericAPIView):
             )
         return response
 
+    @cache_memoize(CACHE_TIMEOUT, args_rewrite=get_charts_args_rewrite)
     def _get_charts(self, request, *args, **kwargs):
         ct = ContentType.objects.get_for_model(Device)
         return Chart.objects.filter(
@@ -134,7 +150,7 @@ class DeviceMetricView(MonitoringApiViewMixin, GenericAPIView):
             return {'data': self.instance.data}
         return {}
 
-    @cache_memoize(24 * 60 * 60, args_rewrite=get_device_args_rewrite)
+    @cache_memoize(CACHE_TIMEOUT, args_rewrite=get_device_args_rewrite)
     def get_object(self, pk):
         return super().get_object()
 
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -2,6 +2,7 @@ import logging
 import uuid
 from datetime import datetime
 
+from cache_memoize import cache_memoize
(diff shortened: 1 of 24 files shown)


pandafy:
The CI is failing due to a mismatch in formatting of timestamp. In the current implementation (master branch), the datetime object gets serialized when passed to celery. Due to this, the `time`  argument of `post_metric_write` signal is a string. When using UDP for writing timeseries database, the datetime object is not serialized (because the task is executed synchonously) and the formatting of timestamp differs. 

Moreover, if we pass a timestamp string to `Metric.write` method, the `post_metric_write` signal will use the exact string for `time` argument which leads to inconsistent results. 

I think we should format all incoming timestamps to the ISO format in `Metric.write` which also includes the timezone information.

openwisp_monitoring/device/api/views.py
```diff
@@ -57,15 +69,23 @@ class DeviceMetricView(MonitoringApiViewMixin, GenericAPIView):
     """
 
     model = DeviceData
-    queryset = (
-        DeviceData.objects.select_related('devicelocation')
```
nemesisdesign, line 61: aren't we using the `devicelocation` data to set the floorplan ID in the metrics?

openwisp_monitoring/device/base/models.py
```diff
@@ -258,13 +258,56 @@ def save_data(self, time=None):
...
+                continue
+            interface_name = interface.get('name')
+            wireless = interface.get('wireless', {})
+            if not wireless or wireless['mode'] != 'access_point':
+                continue
+            ssid = wireless.get('ssid')
+            clients = wireless.get('clients', [])
+            for client in clients:
+                # Save WifiClient
+                client_obj, created = WifiClient.objects.get_or_create(
```
nemesisdesign, line 284: This is also data which is queried often but shouldn't change as often, can you cache this too?
I think cache invalidation would be simple here, when save is called the cache can be invalidated.

openwisp_monitoring/device/api/views.py
pandafy: The `devicelocation` data is used in the celery task where the DeviceData object is queried from the database

https://github.com/openwisp/openwisp-monitoring/blob/d4bb639c55cc794d720e2d477ef5683a528fe136/openwisp_monitoring/device/tasks.py#L58-L65

nemesisdesign: ok :+1:

nemesisdesign (changes requested):
This looks on the right track to me and I do not see any evident flaw in the code.

I do see 2 areas which I think we can improve: fetching metrics and charts objects now is done each time, both when reading and writing. Charts and metric objects are not written often, I believe they're written only once when they are created. From time to time another metric or chart may be written when new data comes in, therefore I think caching the DB calls to fetch these objects will help considerably to speed up execution time, both for writing and reading metrics, and is worth the bit of extra effort, can you proceed with this?

openwisp_monitoring/device/api/views.py
```diff
@@ -106,16 +126,20 @@ def get(self, request, pk):
...
-            metric__object_id=args[0], metric__content_type=ct
+            metric__object_id=args[0], metric__content_type_id=ct.id
         ).select_related('metric')
 
     def _get_additional_data(self, request, *args, **kwargs):
         if request.query_params.get('status', False):
             return {'data': self.instance.data}
         return {}
 
+    @cache_memoize(24 * 60 * 60, args_rewrite=get_device_args_rewrite)
```
nemesisdesign, line 137: let's centralize the default timeout in a setting

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -151,9 +171,12 @@ def _get_or_create(cls, **kwargs):
             if lookup_kwargs.get('name'):
                 del lookup_kwargs['name']
             extra_tags = lookup_kwargs.pop('extra_tags', {})
-            metric = cls.objects.get(**lookup_kwargs)
+            metric_cache_key = get_or_create_metric_cache_key(**kwargs)
```
nemesisdesign, line 174: let's use another private method to fetch the metric object and add the memoize decorator on it

nemesisdesign (changes requested):

nemesisdesign (approved):

Issue #532, opened 2023-08-02
[feature] RADIUS Timeseries Charts


nemesifier:
A number of changes will be made to the OpenWISP RADIUS and OpenWISP Monitoring modules which will make it possible to store RADIUS session data in anonymised form (i.e. personal user data contained in RADIUS session details will be obfuscated with hashing techniques such as "sha1") and user registrations, this will allow a retention of statistical data beyond 12/18 months for analysis purposes by the user without violating privacy and data retention laws.

The data will be stored in the timeseries database (InfluxDB) according to the already adopted retention policy which can be changed upon request.

This feature must be disabled by default and implemented in a separate optional sub module of openwisp-radius (in a similar way as we have created an integration modules in network-topology).

The snapshotting of the radius accounting sessions into influxdb could be done each time a radius session is flagged as closed.
The snapshotting of total sign ups could be done daily at the end of each day.
The daily sign ups timeseries data could be stored whenever a new user signs up and completes the verification process, or it could be taken at the end of each day too, as long as the solution meets the requirements and it's scalable we can be flexible with the implementation details. 

## Time Period

It will be possible to change the period of the timeseries data by going back in time, the minimum period that can be viewed will be one day, the maximum period that can be viewed will be a year, it will be possible to select the dates of interest and the grouping of statistical data over a long period will be one day, i.e. each point on the graph will represent a specific date.

## CSV export

It will also be possible to export the data in CSV format, similar to how it already happens for the monitoring data of the devices.

The graphs that will be added are listed below and indicative graphical examples are provided.

## Permission to view statistics (read only)

The system will allow the definition of user profiles who will have read-only access to the administration interface who will be able to view the statistical data without making any changes.

## New daily sign up charts

![new-users](https://github.com/openwisp/openwisp-radius/assets/841044/ab1b43fb-3250-4e04-b500-331498f4af7d)

It will allow to obtain the following information:

- How many users have signed up each day over time
- Of the users who register every day, it will be possible to know the breakdown by identity verification method (eg: telephone number, credit card, SPID)
- The summary part below will allow you to know the total number of users who registered in the selected period (total users, users registered with telephone number, users registered with credit card, users who registered using SPID).

## Cumulative sign ups charts

![total-registered-users](https://github.com/openwisp/openwisp-radius/assets/841044/cf202cc3-bdf6-44f7-8ec2-03355f77e2ea)

It will allow to obtain the following information:

- How many total users are registered in the system on a given date
[...]

PR #538, opened 2023-08-25
[feature] Added organization filter to timeseries charts #532

pandafy:
- Added option to define "summary_query" for metrics
- Added functionality to handle "GROUP by tags" clause to InfluxDB client

Related to #532

### DIFF
--- a/openwisp_monitoring/db/backends/influxdb/client.py
+++ b/openwisp_monitoring/db/backends/influxdb/client.py
@@ -260,7 +260,7 @@ class DatabaseClient(object):
         if not len(result.keys()) or result.keys()[0][1] is None:
             return list(result.get_points())
         # Handles query which contains "GROUP BY TAG" clause
-        result_points = OrderedDict()
+        result_points = {}
         for (measurement, tag), group_points in result.items():
             tag_suffix = '_'.join(tag.values())
             for point in group_points:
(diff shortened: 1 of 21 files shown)


nemesifier (commented):
I see room for improvement here:

Screenshot from 2023-08-29 16-56-03

Instead of saying "organization filter", I think we should say something which indicates what we're currently viewing, I see these possible options:

- "All organizations" or "All available organizations" or "Organization: any"
- org.name or "Organization: org.name`

Or any other similar solution, what do you think?

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -671,6 +671,8 @@ def get_query(
         params = self._get_query_params(time, start_date, end_date)
         params.update(additional_params)
         params.update({'start_date': start_date, 'end_date': end_date})
+        if not params.get('organization_id') and self.config_dict.get('filter__all__'):
```
nemesifier, line 674: maybe we can use just `__all__` for simplicity and clarity to anyone familiar with python (I think we use similar constructs elsewhere)?

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -260,14 +260,14 @@ def get_list_query(self, query, precision='s'):
         if not len(result.keys()) or result.keys()[0][1] is None:
             return list(result.get_points())
         # Handles query which contains "GROUP BY TAG" clause
-        result_points = {}
+        result_points = OrderedDict()
```
nemesifier, line 263: AFAIK in recent versions of python dicts are ordered by default, isn't it?

nemesifier (commented):
Looks good, two observations below.

nemesifier (approved):

PR #539, opened 2023-08-27
[change] Copy points out of world map only when needed #531

### LINKED ISSUE #531: [change] Copy points out of world map only when needed
Follow up of #393 and #433.

We should trigger the copy of the points only when the view approaches the border of the map.
That way we can avoid hitting the performance penalty if the number of points on the map is high and they get copied for no added benefit. That is, we hit the performance penalty only when needed.

We shall add an event listener for the "mouseend" event of leaflet.js. This function would check the current bounds (what region of map is showing right now). If this region is outside of the default world map, we could add more points.

There is a caveat though, the points would not appear while panning. Only when the panning stops.

pandafy:
Closes #531

### DIFF
--- a/openwisp_monitoring/device/static/monitoring/js/device-map.js
+++ b/openwisp_monitoring/device/static/monitoring/js/device-map.js
@@ -235,7 +235,7 @@
                 map.on('moveend', event => {
                     let netjsonGraph = this;
                     let bounds = event.target.getBounds();
-                    if ((bounds._southWest.lng < -180) && !(netjsonGraph.westWorldFeaturesAppended)) {
+                    if (bounds._southWest.lng < -180 && !netjsonGraph.westWorldFeaturesAppended) {
                         let westWorldFeatures = window.structuredClone(netjsonGraph.data);
                         westWorldFeatures.features.forEach(element => {
                             element.geometry.coordinates[0] -= 360;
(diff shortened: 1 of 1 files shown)


openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -246,6 +232,27 @@
                 map.setMaxBounds(
                     L.latLngBounds(L.latLng(-90, -540), L.latLng(90, 540))
                 );
+                map.on('moveend', event => {
+                    let netjsonGraph = this;
+                    let bounds = event.target.getBounds();
+                    if ((bounds._southWest.lng < -180) && !(netjsonGraph.westWorldFeaturesAppended)) {
```
nemesifier, line 238:
```suggestion
                    if (bounds._southWest.lng < -180 && !netjsonGraph.westWorldFeaturesAppended) {
```
Wouldn't this work too?

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -246,6 +232,27 @@
...
+                        let westWorldFeatures = window.structuredClone(netjsonGraph.data);
+                        westWorldFeatures.features.forEach(element => {
+                            element.geometry.coordinates[0] -= 360;
+                        });
+                        netjsonGraph.utils.appendData(westWorldFeatures, netjsonGraph);
+                        netjsonGraph.westWorldFeaturesAppended = true;
+
+                    }
+                    if ((bounds._northEast.lng > 180) && !(netjsonGraph.eastWorldFeaturesAppended)) {
```
nemesifier, line 247: same here

nemesifier (changes requested):
Tested and is good, see my comments below.

openwisp_monitoring/device/static/monitoring/js/device-map.js
pandafy: I added extra parenthesis for readability.

nemesifier (approved):

PR #542, opened 2023-09-08
[fix] Fixed slow query for WifiSession dashboard pie chart #528

### LINKED ISSUE #528: [bug] Slow WiFi Session dashboard chart database query
The dashboard webpage fails to load when there are large number of devices and active WiFi Sessions.

We need to optimize the query for fetching count of active WiFi sessions. Currently, the following SQL query is getting executed: 
```sql
SELECT SUM("active") FROM (SELECT COUNT(CASE WHEN "device_monitoring_wifisession"."stop_time" IS NULL THEN %s ELSE NULL END) AS "active" FROM "device_monitoring_wifisession" GROUP BY "device_monitoring_wifisession"."id") subquery; 
```

The following is output of `ANALYZE EXPLAIN`: 
```
 Aggregate  (cost=204003.23..204003.24 rows=1 width=32) (actual time=80571.469..80571.472 rows=1 loops=1)
   ->  GroupAggregate  (cost=0.43..166125.92 rows=3030185 width=24) (actual time=493.846..80195.169 rows=3011150 loops=1)
         Group Key: device_monitoring_wifisession.id
         ->  Index Scan using device_monitoring_wifisession_pkey on device_monitoring_wifisession  (cost=0.43..120673.14 rows=3030185 width=24) (actual time=3.448..77936.727 rows=3011150 loops=1)
 Planning Time: 0.154 ms
 JIT:
   Functions: 7
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 35.183 ms, Inlining 0.000 ms, Optimization 80.170 ms, Emission 428.688 ms, Total 544.041 ms
 Execution Time: 80616.425 ms
(10 rows)
```

pandafy:
Closes #528

**Blockers**

- [ ]  https://github.com/openwisp/openwisp-utils/pull/355

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,7 @@ jobs:
       - name: Install openwisp-monitoring
         run: |
           pip install -U -I -e .
+          pip install --no-deps --no-cache-dir --force-reinstall --upgrade https://github.com/openwisp/openwisp-utils/tarball/dashboard-chart-filter
           pip uninstall -y django
           pip install ${{ matrix.django-version }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,6 @@ jobs:
       - name: Install openwisp-monitoring
         run: |
           pip install -U -I -e .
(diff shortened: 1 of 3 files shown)


nemesifier (commented):
This looks good :+1:, waiting for the dependent PR to be merged.

nemesifier (approved):

PR #545, opened 2023-09-20
[feature] Updated label for wireless interface clients

pandafy:
"Access Point MAC address" for wireless interface in "station" mode,
"Associated client MAC address" otherwise.


nemesifier (commented):
Can we take advantage to add to the schema the properties being added in https://github.com/openwisp/openwrt-openwisp-monitoring/pull/125/ ?

openwisp_monitoring/device/templates/admin/config/device/change_form.html
```diff
@@ -175,6 +175,22 @@ <h2>{% trans 'Interface status' %}: {{ interface.name }}</h2>
                       </div>
                   </div>
                   {% endif %}
+                  {% if interface.wireless.bitrate %}
+                  <div class="form-row">
+                      <label>{% trans 'Bitrate' %}:</label>
+                      <div class="readonly">
+                          {{ interface.wireless.bitrate }} KBits/s
```
nemesifier, line 182: can we do some preprocessing to show MBit/s which is probably more user friendly?

nemesifier (changes requested):
Looking good, one last comment below.

openwisp_monitoring/device/base/models.py
```diff
@@ -233,6 +233,11 @@ def _transform_data(self):
                             client['he'] = None
                     elif vht_enabled and client.get('he') is False:
                         client['he'] = None
+            # Convert bitrate from KBits/s to MBits/s
+            if wireless and wireless.get('bitrate'):
+                interface['wireless']['bitrate'] = round(
+                    interface['wireless']['bitrate'] / 1000, 1
+                )
```
nemesifier, line 240: I think this should be:

```python
round(bitrate / 1000.0, 1)
```

here's why:

```
>>> 65000 / 1000
65
>>> 65000 / 1000.
65.0
```

nemesifier (changes requested):

nemesifier (approved):

PR #547, opened 2023-09-26
[fix] Log info in retry mechanism when attempts fail

edoput:
Properly logs exceptions happening in wrapped code.

The `retry` wrapper is hiding exceptions happening in user's code.  This PR will
allow users to inspect the logs and see where the exception originated for tasks
that have been wrapped in the `retry` decorator.

### DIFF
--- a/openwisp_monitoring/utils.py
+++ b/openwisp_monitoring/utils.py
@@ -23,7 +23,7 @@ def retry(method):
             try:
                 return method(*args, **kwargs)
             except Exception as err:
-                logger.info(
+                logger.exception(
                     f'Error while executing method "{method.__name__}":\n{err}\n'
                     f'Attempt {attempt_no} out of {max_retries}.\n'
                 )


openwisp_monitoring/utils.py
```diff
@@ -22,10 +22,9 @@ def wrapper(*args, **kwargs):
         for attempt_no in range(1, max_retries + 1):
             try:
                 return method(*args, **kwargs)
-            except Exception as err:
-                logger.info(
-                    f'Error while executing method "{method.__name__}":\n{err}\n'
-                    f'Attempt {attempt_no} out of {max_retries}.\n'
+            except Exception as e:
+                logger.exception('Error while executing method "%s": %s. Attempt %n out of %n.',
+                    method.__name__, e, attempt_no, max_retries
```
nemesifier, line 27: can you please keep the f-string usage which is more readable?

nemesifier (changes requested):
Thanks @edoput, nice to see you here!

See my comment below.

Please also change the commit message to:

```
[fix] Log exception message during retry mechanism

Avoid swallowing the exception message during the retry mechanism.
```

openwisp_monitoring/utils.py
edoput: By using a template string though we get the string value even if we will not log the message. This way the computation of the message is delayed until the logging system decides that it should actually do the work of formatting the string.

I'm not sure how much runtime inspection the logger does but I guess it will also work (print a summary stack-trace) without having the current exception as an argument.

edoput: @nemesifier  I changed it back to using template strings as logger.exception does the runtime stack checking and can figure out the exception stacktrace to print.

openwisp_monitoring/utils.py
```diff
@@ -23,7 +23,7 @@ def wrapper(*args, **kwargs):
             try:
                 return method(*args, **kwargs)
             except Exception as err:
-                logger.info(
+                logger.exception(
```
nemesifier, line 26: This will log the stacktrace, but it will generate alerts on prod systems which we do not want.
Is the goal of this PR to show the stacktrace in the log message or just show the error?

Something like this should work:

```python
err_info = ''.join(traceback.format_exception(type(err), err, err.__traceback__))
logger.info(
  #...
  f'{err_info}'
  #...
)
```
``

nemesifier (changes requested):
@edoput check my comment below please

openwisp_monitoring/utils.py
edoput: The objective is to show both. Both exception and stack-trace are important for people that need to resolve an issue in their wrapped code. We can have a higher level default log level (logging.CRITICAL) be the default log level and that would solve alerts being generated? This can be user-configurable through the settings API?

I'm currently talking to a user that _is getting spammed_ with alerts but there is no actionable feedback on which part of the application (monitoring backend, protocol transport client, ...) is actually at fault.

nemesifier: ok, a solution similar to the one suggested above will show the stack trace.
In this particular case, we do not want to send alerts for errors that can be retried. We only want to send alerts when there is a problem we can fix. If there's a momentary network error we do not need to be alerted, logging as info is fine.

In the particular case of the user who is getting spammed, are you sure this is the piece of code which is causing the spamming of alerts? This code does not generate any alert unless the retry mechanism gives up.

nemesifier: If you see at the end of the function, if the max attempts are reached, the exception is raised, so there's no need to log exceptions here. If the problem is persistent, it will be raised and logged as error.

PR #553, opened 2023-10-27
additional error checking in ping script

clutch2sft:
Made command build "more correct"

Argument Splitting: Usually, each argument and the command itself are separate elements in the list. If any of the flags or options require values (like -c 5), those should usually also be separate elements in the list.

Implement try catch block in _command to catch any errors from the shell.

Returning the commands exit code.

Check exit code and if it is NOT zero(0)=all the pinged hosts are reachable -or- one(1)=Some hosts were unreachable. Then raise an error to improve (speed up) troubleshooting.

Checks:

- [X] I have manually tested the proposed changes

### DIFF
--- a/openwisp_monitoring/check/classes/ping.py
+++ b/openwisp_monitoring/check/classes/ping.py
@@ -88,14 +88,18 @@ class Ping(BaseCheck):
         command = [
             'fping',
             '-e',  # show elapsed (round-trip) time of packets
-            '-c %s' % count,  # count of pings to send to each target,
-            '-p %s' % interval,  # interval between sending pings(in ms)
-            '-b %s' % bytes_,  # amount of ping data to send
-            '-t %s' % timeout,  # individual target initial timeout (in ms)
+            '-c', str(count),  # count of pings to send to each target
+            '-p', str(interval),  # interval between sending pings (in ms)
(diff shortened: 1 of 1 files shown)


nemesifier (changes requested):
the file is getting entirely rewritten, I am not sure why, can you please fix that? We cannot perform a proper review if we do not see what exactly is getting changed.

clutch2sft:
> the file is getting entirely rewritten, I am not sure why, can you please fix that? We cannot perform a proper review if we do not see what exactly is getting changed.

I can see the individual changes in two commits in the commits tab.  If you can't see those there I can copy into this conversation.  Thanks!  Ggg

nemesifier (commented):
> > the file is getting entirely rewritten, I am not sure why, can you please fix that? We cannot perform a proper review if we do not see what exactly is getting changed.
> 
> I can see the individual changes in two commits in the commits tab. If you can't see those there I can copy into this conversation. Thanks! Ggg
Thanks for explaining.

The second commit seems to be causing issues, could you please remove that commit and add the change that was meant to be in that commit to the previous one using `git commig --amend`? That will update the first commit.

clutch2sft:
I think the commit issue is resolved.  Thanks.

PR #554, opened 2023-11-22
[change] Respect disabled organizations #552

### LINKED ISSUE #552: [change] Respect disabled organizations
When an org is disabled the monitoring module shall respect this by:

- set health status of devices to unknown, clear management ip field (we should seek to do with signals+celery task)
- stop collecting statistics
- stop performing any active check
- hide checks and alert settings in the device page

pandafy:
Closes #552

Checks:

### DIFF
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -58,6 +58,7 @@ class DeviceMonitoringConfig(AppConfig):
         DeviceLocation = load_model('geo', 'DeviceLocation')
         Metric = load_model('monitoring', 'Metric')
         Chart = load_model('monitoring', 'Chart')
+        Organization = load_model('openwisp_users', 'Organization')
 
         post_save.connect(
             self.device_post_save_receiver,
@@ -120,6 +121,11 @@ class DeviceMonitoringConfig(AppConfig):
             sender=DeviceLocation,
             dispatch_uid='post_delete_devicelocation_invalidate_devicedata_cache',
         )
+        post_save.connect(
+            self.organization_post_save_receiver,
+            sender=Organization,
+            dispatch_uid='post_save_organization_clear_management_ip_monitoring_status',
+        )
 
     @classmethod
     def device_post_save_receiver(cls, instance, created, **kwargs):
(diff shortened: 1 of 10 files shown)


openwisp_monitoring/device/apps.py
```diff
@@ -120,6 +121,11 @@ def connect_device_signals(self):
             sender=DeviceLocation,
             dispatch_uid='post_delete_devicelocation_invalidate_devicedata_cache',
         )
+        post_save.connect(
+            self.organization_post_save_receiver,
+            sender=Organization,
+            dispatch_uid='post_save_organization_clear_management_ip_monitoring_status',
```
nemesifier, line 127: Maybe this can be shortened a bit?

Eg: `post_save_organization_disabled_monitoring`.

openwisp_monitoring/device/apps.py
```diff
@@ -134,6 +140,13 @@ def device_post_delete_receiver(cls, instance, **kwargs):
         instance.checks.all().delete()
         instance.metrics.all().delete()
 
+    @classmethod
+    def organization_post_save_receiver(cls, instance, *args, **kwargs):
+        if instance.is_active is False:
+            from .tasks import organization_disabled_handler
```
nemesifier, line 146: is the import here necessary?

openwisp_monitoring/device/base/models.py
```diff
@@ -435,6 +435,25 @@ def is_metric_critical(metric):
                 return True
         return False
 
+    @classmethod
+    def organization_disabled_handler(cls, organization_id):
```
nemesifier, line 439: Why this name? What about `handle_disabled_organization` or anything like that?

openwisp_monitoring/device/tasks.py
```diff
@@ -63,3 +63,9 @@ def write_device_metrics(pk, data, time=None, current=False):
     except DeviceData.DoesNotExist:
         return
     device_data.writer.write(data, time, current)
+
+
+@shared_task(base=OpenwispCeleryTask)
+def organization_disabled_handler(organization_id):
```
nemesifier, line 69: similar here, why the `handler` suffix if we don't have anything like that in any of these tasks above?
Let's be consistent with the naming styles used.

nemesifier (changes requested):

openwisp_monitoring/device/apps.py
pandafy: Yes, because `openwisp_monitoring.device.tasks` imports `openwisp_monitoring.checks.tasks.perform_check` which in-turn imports the `ContentType` model globally.

openwisp_monitoring/device/tests/test_api.py
```diff
@@ -155,27 +155,6 @@ def test_400(self):
         )
         self.assertEqual(r.status_code, 400)
 
-    def test_200_none(self):
```
nemesifier, line 158: why is this test being removed?

nemesifier (approved):

PR #570, opened 2024-04-04
[change] Renamed celery monitoring tasks with leading underscores #569

### LINKED ISSUE #569: [change] Rename celery tasks _timeseries_write and _timeseries_batch_write
Given the following ROUTES setting:

```python
CELERY_TASK_ROUTES = {
        # timeseries writing
        'openwisp_monitoring.monitoring.tasks.*': {'queue': 'manager'},
}
``` 

We would expect that tasks are routed to the manager queue but the following two tasks are still routed to the default queue:

- `openwisp_monitoring.monitoring.tasks._timeseries_write`
- `openwisp_monitoring.monitoring.tasks._timeseries_batch_write`

This has unexpected consequences.

It may be caused by the leading underscore in the task name. If that's the case it would be appropriate to remove the leading underscore and make sure functions used internally are using the underscore.

pandafy:
Fixes #569

Checks:

### DIFF
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -25,7 +25,7 @@ from openwisp_utils.base import TimeStampedEditableModel
 
 from ...db import device_data_query, timeseries_db
 from ...monitoring.signals import threshold_crossed
-from ...monitoring.tasks import timeseries_write
+from ...monitoring.tasks import _timeseries_write
 from ...settings import CACHE_TIMEOUT
 from .. import settings as app_settings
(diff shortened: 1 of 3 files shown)


nemesifier (changes requested):
This is not what I meant.
I did not mean to change the name.

I meant to swap the names of `_timeseries_write` and `timeseries_write`.
I think the celery task should be named `timeseries_write` and the function used internally in the application should be named `_timeseries_write`, while now is the opposite.

nemesifier (approved):

PR #578, opened 2024-05-09
[change] Set device status to unknown when critical checks are disaled

### LINKED ISSUE #576: [change] Disabling ping check should probably set the health status to "unkown"
Disabling a check for any metric which is listed in `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS` (by defaults lists only ping), should put the device in unkown status.

If we don't do this, the device will show its status as "ok", even though ping checks are not performed anymore, and hence the system cannot be sure the device is really reachable. This will be confusing.

We shall add a data migration which can fix this retroactively for checks which have been already disabled.

Here's more details about how this should be implemented:

- In the device page, there's a "Checks" tab, from which ping checks can be disabled (remove the "active" flag and save).
  Checks can be also deleted, I think we should listen also for this kind of signal.
- We shall add a receiver function which listens to the change and delete signals of `Check`
  - The receiver shall evaluate the check to see whether this is a cheeck for one of the metrics listed in `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS`, if not, it shall return without doing anything, otherwise it proceeds
  - The next step is to get the device and flag its status to Unknown.

praptisharma28:
Implements and closes #576.

Checks:

### DIFF
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -44,11 +44,30 @@ class DeviceMonitoringConfig(AppConfig):
         self.connect_config_status_changed()
         self.connect_wifi_client_signals()
         self.connect_offline_device_close_wifisession()
+        self.connect_check_signals()
         self.device_recovery_detection()
         self.set_update_config_model()
         self.register_dashboard_items()
         self.register_menu_groups()
         self.add_connection_ignore_notification_reasons()
+
+    def connect_check_signals(self):
+        from django.db.models.signals import post_delete, post_save
+        from swapper import load_model
+
+        Check = load_model('check', 'Check')
+        DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+        post_save.connect(
+            DeviceMonitoring.handle_critical_metric,
+            sender=Check,
+            dispatch_uid='check_post_save_receiver',
+        )
+        post_delete.connect(
+            DeviceMonitoring.handle_critical_metric,
+            sender=Check,
+            dispatch_uid='check_post_delete_receiver',
+        )
 
     def connect_device_signals(self):
         from .api.views import DeviceMetricView
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -9,7 +9,7 @@ from dateutil.relativedelta import relativedelta
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete
 from django.dispatch import receiver
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -468,8 +468,6 @@ class AbstractDeviceMonitoring(TimeStampedEditableModel):
                 device_monitoring = cls.objects.get(device=instance.content_object)
                 if not instance.is_active or kwargs.get('signal') == post_delete:
                     device_monitoring.update_status('unknown')
-                elif kwargs.get('signal') == post_save and instance.is_active:
-                    device_monitoring.update_status('ok')
             except cls.DoesNotExist:
                 pass
 
--- a/openwisp_monitoring/device/base/models.py
+++ b/openwisp_monitoring/device/base/models.py
@@ -9,6 +9,7 @@ from dateutil.relativedelta import relativedelta
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
@@ -21,6 +22,7 @@ from pytz import timezone as tz
 from swapper import load_model
 
 from openwisp_controller.config.validators import mac_address_validator
+from openwisp_monitoring.device.settings import get_critical_device_metrics
 from openwisp_utils.base import TimeStampedEditableModel
 
 from ...db import device_data_query, timeseries_db
@@ -453,6 +455,23 @@ class AbstractDeviceMonitoring(TimeStampedEditableModel):
         cls.objects.filter(device__organization_id=organization_id).update(
             status='unknown'
         )
+
+    @classmethod
+    def _get_critical_metric_keys(cls):
+        return [metric['key'] for metric in get_critical_device_metrics()]
+
+    @classmethod
+    def handle_critical_metric(cls, instance, **kwargs):
+        critical_metrics = cls._get_critical_metric_keys()
+        if instance.check_type in critical_metrics:
+            try:
+                device_monitoring = cls.objects.get(device=instance.content_object)
+                if not instance.is_active or kwargs.get('signal') == post_delete:
+                    device_monitoring.update_status('unknown')
+                elif kwargs.get('signal') == post_save and instance.is_active:
+                    device_monitoring.update_status('ok')
+            except cls.DoesNotExist:
+                pass
 
 
 class AbstractWifiClient(TimeStampedEditableModel):
--- a/openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
+++ b/openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
@@ -1,0 +1,27 @@
+from django.db import migrations
+from swapper import load_model
+
+
+def update_critical_device_metric_status(apps, schema_editor):
+    Check = apps.get_model('check', 'Check')
+    # We need to load the real concrete model here, because we
+    # will be calling one of its class methods below
+    DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+    critical_metrics_keys = DeviceMonitoring._get_critical_metric_keys()
+
+    for check in Check.objects.filter(
+        is_active=False, check_type__in=critical_metrics_keys
+    ).iterator():
+        DeviceMonitoring.handle_critical_metric(check)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('device_monitoring', '0008_alter_wificlient_options'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            update_critical_device_metric_status, reverse_code=migrations.RunPython.noop
+        ),
+    ]
--- a/openwisp_monitoring/device/tests/test_models.py
+++ b/openwisp_monitoring/device/tests/test_models.py
@@ -598,12 +598,12 @@ class TestDeviceMonitoring(CreateConnectionsMixin, BaseTestCase):
         dm.refresh_from_db()
         self.assertEqual(dm.status, 'unknown')
         with self.subTest(
-            "Ensure status doesn't change on saving active critical check"
+            'Ensure status does not change on saving active critical check'
         ):
             ping_check_instance.is_active = True
             ping_check_instance.save()
             dm.refresh_from_db()
-            self.assertEqual(dm.status, 'ok')
+            self.assertEqual(dm.status, 'unknown')
 
     def test_saving_non_critical_check(self):
         Check = load_model('check', 'Check')
@@ -621,7 +621,7 @@ class TestDeviceMonitoring(CreateConnectionsMixin, BaseTestCase):
             non_critical_check.is_active = False
             non_critical_check.save()
         self.assertEqual(
-            handler.call_count, 0, "Signal should not be fired for non-critical check"
+            handler.call_count, 0, 'Signal should not be fired for non-critical check'
         )
         dm.refresh_from_db()
         self.assertEqual(dm.status, 'ok')
--- a/openwisp_monitoring/device/tests/test_models.py
+++ b/openwisp_monitoring/device/tests/test_models.py
@@ -578,6 +578,73 @@ class TestDeviceMonitoring(CreateConnectionsMixin, BaseTestCase):
         )
         return dm, ping, load, process_count
 
+    def test_disabling_critical_check(self):
+        Check = load_model('check', 'Check')
+        dm, ping, load, process_count = self._create_env()
+        ping_check_instance = Check.objects.create(
+            name='Ping Check',
+            check_type='ping',
+            content_object=dm.device,
+            params={},
+        )
+        dm.update_status('ok')
+        with catch_signal(health_status_changed) as handler:
+            ping_check_instance.is_active = False
+            ping_check_instance.save()
+        self.assertEqual(handler.call_count, 1)
+        call_args = handler.call_args[1]
+        self.assertEqual(call_args['instance'], dm)
+        self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+        with self.subTest(
+            "Ensure status doesn't change on saving active critical check"
+        ):
+            ping_check_instance.is_active = True
+            ping_check_instance.save()
+            dm.refresh_from_db()
+            self.assertEqual(dm.status, 'ok')
+
+    def test_saving_non_critical_check(self):
+        Check = load_model('check', 'Check')
+        dm, ping, load, process_count = self._create_env()
+        # Ensure initial status is 'ok'
+        dm.update_status('ok')
+        # Created a non-critical check
+        non_critical_check = Check.objects.create(
+            name='Configuration Applied',
+            check_type='non_critical',
+            content_object=dm.device,
+            params={},
+        )
+        with catch_signal(health_status_changed) as handler:
+            non_critical_check.is_active = False
+            non_critical_check.save()
+        self.assertEqual(
+            handler.call_count, 0, "Signal should not be fired for non-critical check"
+        )
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'ok')
+
+    def test_deleting_critical_check(self):
+        Check = load_model('check', 'Check')
+        dm, ping, load, process_m = self._create_env()
+        ping_check_instance = Check.objects.create(
+            name='Ping Check',
+            check_type='ping',
+            content_object=dm.device,
+            params={},
+        )
+        dm.update_status('ok')
+        with catch_signal(health_status_changed) as handler:
+            ping_check_instance.delete()
+        self.assertEqual(handler.call_count, 1)
+        call_args = handler.call_args[1]
+        self.assertEqual(call_args['instance'], dm)
+        self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+
     def test_status_changed(self):
         dm, ping, load, process_count = self._create_env()
         # check signal
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,11 +86,13 @@ jobs:
           SAMPLE_APP=1 coverage run --source=openwisp_monitoring runtests.py
           coverage run -a --source=openwisp_monitoring runtests.py
           TIMESERIES_UDP=1 coverage run -a --source=openwisp_monitoring runtests.py
+          coverage xml
 
       - name: Upload Coverage
         uses: coverallsapp/github-action@v2
         with:
           parallel: true
+          format: cobertura
           flag-name: python-${{ matrix.python-version }}-${{ matrix.django-version }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 


nemesifier (changes requested):
It looks well done @praptisharma28, but an important part is missing: the unit tests.
We need one test for the case in which a check is disabled, a test for case in which the check is saved but not being changed (to ensure the health status is not changed inadvertently as it seems to happen now with the current code) and another test for the case in which the check is deleted.
You could write only one test with multiple sub tests using the `subTest` method too, whatever works best according to the test code (see existing tests across the codebase as a reference).

openwisp_monitoring/device/tests/test_check_signals.py
```diff
@@ -0,0 +1,48 @@
...
+            name='Ping Check',
+            check_type=CHECK_CLASSES[0][0],
+            content_object=self.device,
+            params={},
+        )
+
+    def test_check_signals(self):
+        with self.subTest('Test disabling a critical check'):
+            self.ping_check.is_active = False
+            self.ping_check.save()
```
nemesifier, line 24: Let's assert that the `health_status_change` signal has been sent as expected.

Look for `catch_signal` in the codebase of this repo to see how to do that, it's also documented in the openwisp-utils README (please read it).

openwisp_monitoring/device/apps.py
```diff
@@ -26,10 +27,13 @@
 from .signals import device_metrics_received, health_status_changed
 from .utils import (
     get_device_cache_key,
+    handle_critical_check_change,
     manage_default_retention_policy,
     manage_short_retention_policy,
 )
 
+Check = load_model('check', 'Check')
```
nemesifier, line 35: You can't load this here, the app will break because of this. Observe how it's done in `connect_device_signals`.

openwisp_monitoring/device/apps.py
```diff
@@ -49,6 +53,19 @@ def ready(self):
         self.register_dashboard_items()
         self.register_menu_groups()
         self.add_connection_ignore_notification_reasons()
+        self.connect_check_signals()
```
nemesifier, line 56: move this after the last connect method call for consistency.

openwisp_monitoring/device/apps.py
```diff
@@ -49,6 +53,19 @@ def ready(self):
...
+
+    def connect_check_signals(self):
+        post_save.connect(
+            check_post_save_receiver,
```
nemesifier, line 60: Why not just point it to `DeviceData.handle_unknown_status_change` and have one class method there? Look at the class method receivers used in `connect_device_signals`.

openwisp_monitoring/device/tests/test_check_signals.py
```diff
@@ -0,0 +1,48 @@
+from django.test import TestCase
```
nemesifier, line 1: We have a file which contains some tests for `apps.py`, I think we can add a test there instead of creating a new test class just for this.

openwisp_monitoring/device/utils.py
```diff
@@ -1,9 +1,24 @@
...
+from .signals import device_status_unknown
 
 SHORT_RP = 'short'
 DEFAULT_RP = 'autogen'
 
+Device = load_model('config', 'Device')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+
+def handle_critical_check_change(check):
```
nemesifier, line 15: So all this code logic should go in a class method on the `DeviceData` model as suggested earlier.

openwisp_monitoring/device/utils.py
```diff
@@ -1,9 +1,24 @@
...
+    critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+    if check.object_id and check.metric.name in critical_metrics:
+        device = Device.objects.get(pk=check.object_id)
+        device.monitoring.update_status('unknown')
+        device_status_unknown.send(sender=DeviceMonitoring, instance=device.monitoring)
```
nemesifier, line 20: we can eliminate this signal because the `update_status` method will take care of sending the `health_status_changed` signal which can be used to know when the health status has moved to unknown.

openwisp_monitoring/device/utils.py
```diff
@@ -1,9 +1,24 @@
...
 SHORT_RP = 'short'
 DEFAULT_RP = 'autogen'
 
+Device = load_model('config', 'Device')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+
+def handle_critical_check_change(check):
+    critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+    if check.object_id and check.metric.name in critical_metrics:
```
nemesifier, line 17: `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS` contains a list of dictionaries, so you're comparing a string to a list of dictionaries, it will always return `False`

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,27 @@
...
+Device = load_model('config', 'Device')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+
+def update_device_status(apps, schema_editor):
+    critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+    for check in Check.objects.filter(is_active=False):
+        if check.object_id and check.metric.name in critical_metrics:
+            device = Device.objects.get(pk=check.object_id)
+            device.monitoring.update_status('unknown')
```
nemesifier, line 15: The last 3 lines here like a duplication of the logic of the receiver. Just call the same receiver function here, perform any adaptation if needed, let's not duplicate logic.

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,27 @@
...
+from swapper import load_model
+
+Check = load_model('check', 'Check')
+Device = load_model('config', 'Device')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+
+def update_device_status(apps, schema_editor):
+    critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+    for check in Check.objects.filter(is_active=False):
```
nemesifier, line 12: 1. let's use the `iterator()` method to avoid this line to potentially crash big databases
2. let's filter checks properly, eg: `Check.objects.filter(is_active=False, metric__key__in=[critical_metrics_key])`, you'll need to prepare `critical_metrics_key` beforehand as `OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS` is a list of dictionaries.

nemesifier (changes requested):
Needs more work. See my comments below.

openwisp_monitoring/device/apps.py
nemesifier: This comments hasn't been addressed

openwisp_monitoring/device/models.py
```diff
@@ -19,6 +20,21 @@ class Meta:
         proxy = True
         swappable = swappable_setting('device_monitoring', 'DeviceData')
 
+    def handle_unknown_status_change(self, instance, **kwargs):
+        critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+        if instance.metric.name in critical_metrics:
```
nemesifier, line 25: again, this will not work.

openwisp_monitoring/device/models.py
```diff
@@ -19,6 +20,21 @@ class Meta:
...
+            if kwargs.get('created', False):
+                self.update_status('unknown')
+            elif kwargs.get('sender', None).__name__ == 'Check':
+                self.update_status('unknown')
+
+    def handle_critical_check_change(cls, check):
+        critical_metrics = settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
+        device_data_instances = DeviceData.objects.filter(pk=check.object_id)
+        for instance in device_data_instances:
+            if check.metric.name in critical_metrics:
```
nemesifier, line 35: same here

openwisp_monitoring/device/models.py
```diff
@@ -19,6 +20,21 @@ class Meta:
...
+                instance.update_status('unknown')
```
nemesifier, line 36: why 2 methods? They seem mostly identical, I think we should aim to have one.

openwisp_monitoring/device/apps.py
```diff
@@ -49,6 +51,24 @@ def ready(self):
...
+        self.connect_check_signals()
+
+    def connect_check_signals(self):
+        from swapper import load_model
+
+        Check = load_model('check', 'Check')
+        DeviceData = load_model('device_monitoring', 'DeviceData')
+
+        @receiver(post_save, sender=Check)
+        def check_post_save_receiver(sender, instance, **kwargs):
```
nemesifier, line 63: why is this left here?

openwisp_monitoring/device/apps.py
```diff
@@ -49,6 +51,24 @@ def ready(self):
...
+            if not instance.is_active:
+                handle_critical_check_change(instance)
+
+        post_save.connect(
+            DeviceData.handle_unknown_status_change,
+            sender=Check,
+            dispatch_uid='check_post_save_receiver',
+        )
```
nemesifier, line 71: aren't you forgetting to connect the post_delete signal?

nemesifier (changes requested):

openwisp_monitoring/device/models.py
pandafy, line 36: We don't add any code directly to the Model class. We add all the logic to the AbstractModel, so it allows to extend the models. Please keep this in mind.

openwisp_monitoring/device/utils.py
```diff
@@ -1,9 +1,15 @@
...
+
 from ..db import timeseries_db
 from . import settings as app_settings
 
 SHORT_RP = 'short'
 DEFAULT_RP = 'autogen'
 
+Device = load_model('config', 'Device')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+DeviceData = load_model('device_monitoring', 'DeviceData')
```
pandafy, line 11: @praptisharma28 you are getting the `django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.` error because of these assignments here. 

This file is loaded in the global namespace in openwisp_monitoring/device/apps.py. Thus, the code is trying to load the model before Django app has been loaded. 

I don't see these models being used in this file. Moreover, this file does not look right to store the signal receiver. I would add the logic to a classmethod in AbstractDeviceMonitoring, similar to `handle_disabled_organization` https://github.com/openwisp/openwisp-monitoring/blob/998c81b4a6e7600d418db2fb96e6a975f72b45cc/openwisp_monitoring/device/base/models.py#L438-L455

praptisharma28: OKay, got the idea. Thanks Gagan.

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +454,20 @@ def handle_disabled_organization(cls, organization_id):
...
 
+    @classmethod
+    def handle_critical_metric(cls, metric):
+        """
+        Updates the device status to 'unknown' if the given metric
+        is a critical metric from the CRITICAL_DEVICE_METRICS setting.
+        """
+        DeviceData = load_model('device_monitoring', 'DeviceData')
+
+        if metric.name in [m['key'] for m in app_settings.CRITICAL_DEVICE_METRICS]:
```
pandafy, line 465: There is a property  `related_metrics` of DeviceMonitoring model which returns the related metrics of the device. We shall loop over that queryset and update critical metrics accordingly. 

https://github.com/openwisp/openwisp-monitoring/blob/998c81b4a6e7600d418db2fb96e6a975f72b45cc/openwisp_monitoring/device/base/models.py#L386-L393

pandafy: I would just do the following to get the critical metrics. 

```python 
qs = cls.related_metrics.filter(key__in=app_settings.CRITICAL_DEVICE_METRICS)
```

pandafy: I gave a terrible suggestion earlier. I had the wrong impression that we have a device object here because we are in the AbstractDeviceModel. 

Instead, we have a metric object here. I think, for the sake of readability we should pass `Check` object to the receiver here and then retrieve the related DeviceMonitoring and Metric object in the receiver. 

The `Check` model has a GenericForeignKey relation which includes the object_id of the device. We can get the DeviceMonitoring object like so  

```python
device_monitoring = cls.objects.get(device_id=check.object_id)
device_monitoring.update_status('unknown')
```

I think the this conditional suffices this requirement of the issue, so we don't need to perform another database query 

> Disabling a check for any metric which is listed in OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS (by defaults lists only ping), should put the device in unkown status.

```
if metric.name in [m['key'] for m in app_settings.CRITICAL_DEVICE_METRICS]:
```

praptisharma28: ``
    def connect_check_signals(self):
        from swapper import load_model

        Check = load_model('check', 'Check')
        DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')

        @receiver(post_save, sender=Check)
        def check_post_save_receiver(sender, instance, **kwargs):
            DeviceMonitoring.handle_critical_metric(instance)  

        @receiver(post_delete, sender=Check)
        def check_post_delete_receiver(sender, instance, **kwargs):
            DeviceMonitoring.handle_critical_metric(instance)  
 ``
 
 `` 
     @classmethod
    def handle_critical_metric(cls, check):
        """
        Updates the device status to 'unknown' if the given check
        is for a critical metric from the CRITICAL_DEVICE_METRICS setting
        and the check is not active.
        """

        critical_metrics = getattr(
            settings,
            'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS',
            [{'key': 'ping', 'field_name': 'reachable'}],
        )
        # Check if the check is associated with a critical metric
        if check.metric in [m['key'] for m in critical_metrics] and not check.is_active:
            try:
                device_monitoring = cls.objects.get(device_id=check.object_id)
                device_monitoring.update_status('unknown')
            except cls.DoesNotExist:
                pass
 ``

openwisp_monitoring/device/tests/test_apps.py
```diff
@@ -37,3 +40,54 @@ def test_device_map_template_registered(self):
                 'monitoring/js/device-map.js',
             ),
         )
+
+
+class TestCheckSignals(TestCase):
```
pandafy, line 45: The tests are failing because this class is not inheriting the test mixins which provide helper methods like `_create_device`.

Instead of adding a new test class just for these tests, I think we can move these tests to `TestDeviceMonitoring` in openwisp_monitoring/device/tests/test_models.py

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations
+from swapper import load_model
+
+Check = load_model('check', 'Check')
+DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+
+def update_device_status(apps, schema_editor):
+    critical_metrics_keys = [
+        metric['key'] for metric in settings.OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS
```
pandafy, line 11: You have to access the setting, just like you are doing it here 

https://github.com/openwisp/openwisp-monitoring/pull/578/files#diff-e23325d8b8273f2f23eca24901c37542e622ff22082652b203617ea769f98ba6R465

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +454,27 @@ def handle_disabled_organization(cls, organization_id):
             status='unknown'
         )
 
+    @classmethod
+    def handle_critical_metric(cls, check, **kwargs):
```
pandafy, line 458: @praptisharma28 your tests are failing because the signature of this method is wrong. The `post_save` signal along with other model signals in Django calls the signal receivers with keyword arguments. . Read the [Django documentation for the arguments sent with post_save signal](https://docs.djangoproject.com/en/5.0/ref/signals/#post-save) 

There's no `check` argument, but it called `instance` which references the model object. The test wouldn't fail like this, if you rename the `check`  argument to `instance`.  You would need to update the code logic similarly.

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +454,27 @@ def handle_disabled_organization(cls, organization_id):
...
+        and the check is not active.
+        """
+        from django.conf import settings
+
+        critical_metrics = getattr(
+            settings,
+            'OPENWISP_MONITORING_CRITICAL_DEVICE_METRICS',
+            [{'key': 'ping', 'field_name': 'reachable'}],
+        )
+        if check.check_type in [m['key'] for m in critical_metrics]:
```
nemesifier, line 471: The `check_type` value for ping checks is `openwisp_monitoring.check.classes.Ping`, so this if block will never be executed with this code.

We can get the critical metrics like:

```python
Metric.objects.filter(key__in=critical_metric_list, object_id=check.object_id, content_type=check.content_type)
```

But you need to build `critical_metric_list` by preprocessing the output of `get_critical_device_metrics` so the result is a list of strings (now it returns a list of dicts and it won't work).

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,33 @@
...
+        'check', 'Check'
+    )  # Ensure 'device' is the app where model is.
+    DeviceMonitoring = apps.get_model('device_monitoring', 'DeviceMonitoring')
+
+    critical_metrics_keys = [metric['key'] for metric in get_critical_device_metrics()]
+
+    for check in Check.objects.filter(
+        is_active=False, check_type__in=critical_metrics_keys
+    ).iterator():
+        DeviceMonitoring.handle_critical_metric(check)
```
nemesifier, line 17: the best will be to make sure we do not duplicate logic already being defined in `openwisp_monitoring/device/base/models.py`. Look for any similarities and create private methods in the models that can be used also here whenever possible.

nemesifier (changes requested):

openwisp_monitoring/device/apps.py
```diff
@@ -44,12 +45,36 @@ def ready(self):
...
+        from django.db.models.signals import post_delete, post_save
+        from swapper import load_model
+
+        Check = load_model('check', 'Check')
+        DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
+
+        @receiver(post_save, sender=Check)
+        @receiver(post_delete, sender=Check)
+        def update_device_status(sender, instance, **kwargs):
+            DeviceMonitoring.handle_critical_metric(instance)
```
nemesifier, line 65: This is redundant with the code below.

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
nemesifier: I believe this migration will fail, because `apps.get_model('device_monitoring', 'DeviceMonitoring')` does not return the real model, but a simulation generated from the Migration framework.

Try to rollback this migration, then set some critical checks to False in your dev env, then run the migration again. What happens? I think it's highly unlikely it will run properly.
We will likely have to load the real (aka "concrete") `DeviceMonitoring` model here.

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+from openwisp_monitoring.device.settings import get_critical_device_metrics
+
+
+def update_critical_device_metric_status(apps, schema_editor):
+    Check = apps.get_model('check', 'Check')
+    DeviceMonitoring = apps.get_model('device_monitoring', 'DeviceMonitoring')
+    critical_metrics_keys = [metric['key'] for metric in get_critical_device_metrics()]
```
nemesifier, line 9: There's some duplication here, this logic could be encapsulated in a method, eg: `critical_keys = DeviceMonitoring._get_critical_metric_keys()`

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +455,22 @@ def handle_disabled_organization(cls, organization_id):
             status='unknown'
         )
 
+    @classmethod
+    def handle_critical_metric(cls, instance, **kwargs):
+        from openwisp_monitoring.device.settings import get_critical_device_metrics
+
+        critical_metrics = [metric['key'] for metric in get_critical_device_metrics()]
```
nemesifier, line 462: once you follow the previous suggestions, you can change this to: `critical_keys = DeviceMonitoring._get_critical_metric_keys()`

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +455,22 @@ def handle_disabled_organization(cls, organization_id):
...
+        if instance.check_type in critical_metrics:
+            try:
+                device_monitoring = cls.objects.get(device=instance.content_object)
+                if not instance.is_active or kwargs.get('signal') == post_delete:
+                    device_monitoring.update_status('unknown')
+                    health_status_changed.send(
```
nemesifier, line 468: Isn't this already performed by the `update_status`? It seems redundant to me and hence I think it should be removed.

nemesifier (changes requested):
Good progress @praptisharma28, see my comments below.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
         )
         return dm, ping, load, process_count
 
+    def test_disabling_critical_check(self):
+
```
nemesifier, line 582: Remove this blank line and any other blank line in **all these tests that you added** to maintain consistency with the rest of the test of this document. Occasional blank lines are fine but if you need to add so many, you probably need to increase the font size and/or line height of your code editor.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+            ping_check_instance.save()
+        if handler.call_count > 0:
+            call_args = handler.call_args[1]
+            self.assertEqual(call_args['instance'], dm)
+            self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+        dm.update_status('ok')
+
+    def test_saving_active_critical_check(self):
```
nemesifier, line 606: I tried running this test on master and is not failing.
What are we trying to test here? Are we checking whether the new code affects normal operations?
In that case I think we can merge it with the previous tests: at the end of the previous test, just change any detail of the check, eg: the "name" and save again, then ensure the status hasn't changed and the signal hasn't been fired. This can be executed as a `subTest`.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+
+        ping_check_instance = Check.objects.create(
+            name='Ping Check',
+            check_type='ping',
+            content_object=dm.device,
+            params={},
+        )
+        dm.update_status('ok')
+
+        self.assertEqual(dm.status, 'ok')
```
nemesifier, line 594: This line is useless, please remove it.
We already trust `dm.update_status('ok')` to work because there's other tests which ensure that, so there's no reason to check again here. Let's avoid bloating the tests.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+        with catch_signal(health_status_changed) as handler:
+            ping_check_instance.is_active = False
+            ping_check_instance.save()
+        if handler.call_count > 0:
```
nemesifier, line 598: why this if statement? Please replace this with just asserting that it has been called once, then proceed with the other assertions below which look good.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+            call_args = handler.call_args[1]
+            self.assertEqual(call_args['instance'], dm)
+            self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+        dm.update_status('ok')
```
nemesifier, line 604: Remove this line, it doesn't make any sense here.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+        )
+        dm.update_status('ok')
+
+        self.assertEqual(dm.status, 'ok')
+        ping_check_instance.is_active = True
+        ping_check_instance.save()
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'ok')
+
+    def test_saving_non_critical_check(self):
```
nemesifier, line 625: I think we need to change this too.

The point of this test is to disable a non critical check, to ensure it doesn't affect the health status.

So, you should set the health status of the device to ok, then disable the check and ensure the status hasn't changed and the signal hasn't been fired.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,90 @@ def _create_env(self):
...
+            check_type='ping',
+            content_object=dm.device,
+            params={},
+        )
+        dm.update_status('ok')
+
+        with catch_signal(health_status_changed) as handler:
+            ping_check_instance.delete()
+
+        if handler.call_count > 0:
```
nemesifier, line 657: the same about the if statement I explained above applies here

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +455,25 @@ def handle_disabled_organization(cls, organization_id):
             status='unknown'
         )
 
+    @classmethod
+    def _get_critical_metric_keys(cls):
+        from openwisp_monitoring.device.settings import get_critical_device_metrics
```
nemesifier, line 460: Please move this to the top imports of the file, there's no good reason to import this here.

openwisp_monitoring/device/migrations/0009_update_device_status_for_disabled_critical_checks.py
```diff
@@ -0,0 +1,25 @@
+from django.db import migrations
+from swapper import load_model
+
+
+def update_critical_device_metric_status(apps, schema_editor):
+    Check = apps.get_model('check', 'Check')
+    DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
```
nemesifier, line 7: Add a comment above here explaining why we need the concrete model here. Eg:

```suggestion
    # We need to load the real concrete model here, because we 
    # will be calling one of its class methods below
    DeviceMonitoring = load_model('device_monitoring', 'DeviceMonitoring')
```

nemesifier (changes requested):
Good progress @praptisharma28.

See my comments below, I believe that once addressing those we can merge.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,73 @@ def _create_env(self):
...
+            ping_check_instance.is_active = False
+            ping_check_instance.save()
+        self.assertEqual(handler.call_count, 1)
+        call_args = handler.call_args[1]
+        self.assertEqual(call_args['instance'], dm)
+        self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+        with self.subTest(
+            "Ensure status doesn't change on saving active critical check"
```
nemesifier, line 601: we are using single quotes in this codebase, please maintain consistency

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,73 @@ def _create_env(self):
...
+            name='Configuration Applied',
+            check_type='non_critical',
+            content_object=dm.device,
+            params={},
+        )
+        with catch_signal(health_status_changed) as handler:
+            non_critical_check.is_active = False
+            non_critical_check.save()
+        self.assertEqual(
+            handler.call_count, 0, "Signal should not be fired for non-critical check"
```
nemesifier, line 624: same here

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +578,73 @@ def _create_env(self):
...
+        self.assertEqual(call_args['status'], 'unknown')
+        dm.refresh_from_db()
+        self.assertEqual(dm.status, 'unknown')
+        with self.subTest(
+            "Ensure status doesn't change on saving active critical check"
+        ):
+            ping_check_instance.is_active = True
+            ping_check_instance.save()
+            dm.refresh_from_db()
+            self.assertEqual(dm.status, 'ok')
```
nemesifier, line 606: This is so wrong @praptisharma28, the description you just wrote above says "Ensure status doesn't change on saving active critical check", but the status here is changing! This assertion must be `self.assertEqual(dm.status, 'unknown')`

openwisp_monitoring/device/base/models.py
```diff
@@ -454,6 +456,23 @@ def handle_disabled_organization(cls, organization_id):
...
+    @classmethod
+    def handle_critical_metric(cls, instance, **kwargs):
+        critical_metrics = cls._get_critical_metric_keys()
+        if instance.check_type in critical_metrics:
+            try:
+                device_monitoring = cls.objects.get(device=instance.content_object)
+                if not instance.is_active or kwargs.get('signal') == post_delete:
+                    device_monitoring.update_status('unknown')
+                elif kwargs.get('signal') == post_save and instance.is_active:
+                    device_monitoring.update_status('ok')
```
nemesifier, line 472: This is wrong. Only checks are responsible for knowing whether a device is ok, we can't know without checking.
Was it specified anywhere that we should set the status to ok when the change goes back to active? Why did you add this?

nemesifier (changes requested):

nemesifier (approved):

PR #580, opened 2024-05-16
New branch

praptisharma28:
Checks:


openwisp_monitoring/utils.py
```diff
@@ -14,6 +15,13 @@ def transaction_on_commit(func):
         transaction.on_commit(func)
 
 
+def _get_org(self):
```
pandafy, line 18: We don't require this. This is already available through test helper mixins

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +579,59 @@ def _create_env(self):
...
+            with catch_signal(health_status_changed) as handler:
+                self.ping_check.is_active = False
+                self.ping_check.save()
+            self.assertEqual(handler.call_count, 1)
+            call_args = handler.call_args[0]
+            self.assertEqual(call_args[0], self.device.monitoring)
+            self.assertEqual(call_args[1], 'unknown')
+            self.device.refresh_from_db()
+            self.assertEqual(self.device.monitoring.status, 'unknown')
+            self.device.monitoring.update_status('ok')
```
pandafy, line 607: These sub-tests look good. I would just un-indent the part for resetting the DeviceMonitoring.status. So, the code gets executed regardless the subTest passes or fails. 

This applies to all the sub-tests.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +579,59 @@ def _create_env(self):
...
+        self.device = self._create_device(organization=self._create_org())
+        self.ping_check = Check.objects.create(
+            name='Ping Check',
+            check_type=CHECK_CLASSES[0][0],
+            content_object=self.device,
+            params={},
+        )
+        self.device.monitoring.update_status('ok')
+
+    def test_check_signals(self):
```
pandafy, line 595: Here, I would call `_create_env`.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -578,6 +579,59 @@ def _create_env(self):
...
+
+        Check = load_model('check', 'Check')
+        self.device = self._create_device(organization=self._create_org())
+        self.ping_check = Check.objects.create(
+            name='Ping Check',
+            check_type=CHECK_CLASSES[0][0],
+            content_object=self.device,
+            params={},
+        )
+        self.device.monitoring.update_status('ok')
```
pandafy, line 593: The `setUp` method is executed before every test case, this is polluting the test envrironment and making other tests fail. We don't want to do this here. Use the `_create_env` method like we do in other tests to create the necessary objects.

PR #581, opened 2024-05-23
[fix] Fixed duplication of points when map bounds are crossed #575

### LINKED ISSUE #575: [bug] General map: points in the current view get duplicated when map bounds are crossed
Points in the current view get duplicated when map bounds are crossed.

monitoring-map-duplication

In the image above we can observe how points in the center get duplicated, see how they seem to be thicker after moving the map around a bit.

I think this was introduced either in https://github.com/openwisp/openwisp-monitoring/pull/539 or https://github.com/openwisp/openwisp-monitoring/pull/549.

pandafy:
Closes #575

**Blockers**

- [ ] https://github.com/openwisp/netjsongraph.js/pull/263


openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -236,22 +236,35 @@
...
+                            element => element.geometry.coordinates[0] <= 180
+                        )
                         westWorldFeatures.features.forEach(element => {
                             if (element.geometry) {
                                 element.geometry.coordinates[0] -= 360;
                             }
                         });
+                        // netjsonGraph.utils.appendData call the render method which
+                        // super imposes the data on the existing points on the map.
+                        netjsonGraph.leaflet.geoJSON.removeFrom(netjsonGraph.leaflet);
```
pandafy, line 250: The points were getting duplicated by just using appendData. I reviewed the code of netjsongraph.js and appendData do not call `netjsonGraph.leaflet.geoJSON.removeFrom(netjsonGraph.leaflet);` like `overrideData` does. 

https://github.com/openwisp/netjsongraph.js/blob/90799710ba2eec7d440a67e58d98452ce3f28aec/src/js/netjsongraph.update.js#L176-L184

I may be wrong, but we may need to replicate this in appendData as well.

nemesifier: I did not understand what you mean exactly with:

> I may be wrong, but we may need to replicate this in appendData as well.

You mean that we should patch that method in netjsongraph.js, or what?

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -236,22 +236,35 @@
...
                         netjsonGraph.westWorldFeaturesAppended = true;
 
                     }
                     if (bounds._northEast.lng > 180 && !netjsonGraph.eastWorldFeaturesAppended) {
                         let eastWorldFeatures = window.structuredClone(netjsonGraph.data);
+                        // Exclude the features that may be added for the West world map
+                        eastWorldFeatures.features = eastWorldFeatures.features.filter(
+                            element => element.geometry.coordinates[0] >= -180
+                        );
+                        window.console.log(eastWorldFeatures.features);
```
nemesifier, line 261: :eyes:?

nemesifier (changes requested):

nemesifier (approved):

PR #584, opened 2024-05-31
[monitoring] Adding influxDB 2.x version support #274

### LINKED ISSUE #274: [CI] Migrate CI build to influxdb 2.0

praptisharma28:
Fixes #274

Checks:


openwisp_monitoring/db/backends/__init__.py
```diff
@@ -57,7 +64,18 @@ def load_backend_module(backend_name=TIMESERIES_DB['BACKEND'], module=None):
...
-timeseries_db = load_backend_module(module='client').DatabaseClient()
+if '2' in TIMESERIES_DB['BACKEND']:
+    timeseries_db = load_backend_module(module='client').DatabaseClient(
+        bucket=TIMESERIES_DB['BUCKET'],
+        org=TIMESERIES_DB['ORG'],
+        token=TIMESERIES_DB['TOKEN'],
+        url=f"http://{TIMESERIES_DB['HOST']}:{TIMESERIES_DB['PORT']}",
+    )
+else:
+    timeseries_db = load_backend_module(module='client').DatabaseClient()
```
pandafy, line 79: Any backend specific configuration should be done in the backend class. Perform this operation in the `__init__` method of the backend class 

```
.DatabaseClient(
        bucket=TIMESERIES_DB['BUCKET'],
        org=TIMESERIES_DB['ORG'],
        token=TIMESERIES_DB['TOKEN'],
        url=f"http://{TIMESERIES_DB['HOST']}:{TIMESERIES_DB['PORT']}",
 ```

**Comment by praptisharma28** [comment]
June 18th weekly call summary:
- Need to get ping check in influxdb 2.0 at the server locally, so that health status like "CRITICAL", "OK" are visible too and not just "UNKNOWN".
- Need to get charts which right now says an error when I check for charts.
Note:  Celery beat and worker will be running all the time along with the server, when I am working on 8000 port locally.

**Review by nemesifier — CHANGES_REQUESTED** [review]
There's too many `if` statements in the code (eg: if influxdb1 or influxdb2).
This is not the way to go. The logic for influxdb1 or influxdb2 must be encapsulated in each respective timeseries DB backend, the rest of the code should just be adapted to call that logic.

Please have a look at the work which was done for the elasticsearch backend to get an idea:
https://github.com/openwisp/openwisp-monitoring/pull/164

File: openwisp_monitoring/db/backends/influxdb2/client.py [hunk 2]
```diff
@@ -0,0 +1,462 @@
...
+        from(bucket: "{bucket}")
+          |> range(start: {start}, stop: {stop})
+          |> filter(fn: (r) => r["_measurement"] == "ping")
+          |> filter(fn: (r) => r["_field"] == "loss" or r["_field"] == "reachable" or r["_field"] == "rtt_avg" or r["_field"] == "rtt_max" or r["_field"] == "rtt_min")
+          |> filter(fn: (r) => r["content_type"] == "config.device")
+          |> filter(fn: (r) => {device_filter})
+          |> aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)
+          |> yield(name: "mean")
+        '''
+        return query
```
pandafy, line 229: Please move this query to queries.py and use `get_query()` method to generate the final query.

pandafy (changes requested):
In our call today, @praptisharma28 walked me through the code and helped me understand the changes. In this process, we found some areas for improvement.

openwisp_monitoring/monitoring/base/models.py
```diff
@@ -674,29 +675,34 @@ def get_query(
         additional_params = additional_params or {}
         params = self._get_query_params(time, start_date, end_date)
         params.update(additional_params)
-        params.update({'start_date': start_date, 'end_date': end_date})
+        params.update({
+            'start_date': start_date, 
+            'end_date': end_date,
+            'measurement': self.config_dict.get('measurement', self.metric.key),
+            'field_name': fields or self.config_dict.get('field_name'),
```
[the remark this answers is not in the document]
praptisharma28: I used pdb, got:
```
(Pdb) params
{'field_name': None, 'key': 'test_metric', 'time': '2024-06-25', 'days': '7d', 'content_type': 'openwisp_users.user', 'object_id': '4f5f77d3-3b30-4eca-97c6-1301217a4edc', 'start_date': None, 'end_date': None, 'measurement': 'test_metric'}
```
But if I remove it, I lose on the summary of ```uptime``` chart. Charts like ```mostly reachable, unreachable, partially reachable``` are lost.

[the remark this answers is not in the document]
pandafy: Screenshot from 2024-07-02 16-53-14


There should be only one summary in the uptime chart.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,462 @@
...
+ 
+    def read(self, key, fields, tags, **kwargs):
+        extra_fields = kwargs.get('extra_fields')
+        since = kwargs.get('since', '-30d')  # Default to last 30 days if not specified
+        order = kwargs.get('order')
+        limit = kwargs.get('limit')
+        bucket = self.bucket
+
+        # Start building the Flux query
+        flux_query = f'from(bucket:"{bucket}")'
```
pandafy, line 156: @praptisharma28 why are you building the query like this here? Why aren't you using the queries defined in queries.py? 

This looks wrong. Charts for different metrics are handled differently. That's why there are individual queries for each chart.

pandafy: My bad, I got confused between the  `get_list_queries` and `read` method. The current implementation is okay.

pushpitkamboj:
Hi, i was going through previous GSOC projects and saw someone did worked on this same gsoc project 2 years ago, was not completed? How much of the project was completed back then?
getting context from this feels relevant for gsoc'26 monitoring project
@pandafy @nemesifier

PR #592, opened 2024-07-08
[fix] Fixed WifiSessionAdmin.stop_time not following project timezone

pandafy:
Checks:

### DIFF
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -8,6 +8,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.forms import ModelForm
 from django.templatetags.static import static
 from django.urls import resolve, reverse
+from django.utils import timezone
 from django.utils.formats import localize
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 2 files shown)


openwisp_monitoring/device/admin.py
```diff
@@ -378,7 +379,7 @@ def wps(self, obj):
     def get_stop_time(self, obj):
         if obj.stop_time is None:
             return mark_safe('<strong style="color:green;">online</strong>')
-        return localize(obj.stop_time)
```
pandafy, line 381: From the [Django docs](https://docs.djangoproject.com/en/5.0/topics/i18n/timezones/): 

> When support for time zones is enabled, Django stores datetime information in UTC in the database, uses time-zone-aware datetime objects internally, and translates them to the end user’s time zone in templates and forms.

The `localize` function formats the time according to the locale set in the project settings. Thus, it returns a string representation of the date (e.g. `8 Jul 2024, 2:52 p.m.`). 

When, USE_TZ is set to True, the code was formatting the UTC time to string instead of first converting the time to the set timezone in the setting. Thus, changing the timezone was having no change in the stop time field of WifiSession admin.

nemesifier (approved):

PR #603, opened 2024-07-26
[fix] Both 'rx_bytes' and 'tx_bytes' are present #595

### LINKED ISSUE #595: [bug] Code issue when writing traffic rx/tx
https://github.com/openwisp/openwisp-monitoring/blob/fb2814b6dd8213ddb45a8497cf41dfb10eee04a2/openwisp_monitoring/device/writer.py#L85

It should be:
```
          if (
                ifstats.get('rx_bytes') is not None
                and ifstats.get('tx_bytes') is not None
            ):
```

praptisharma28:
Fixes #595

## Checklist

- Both 'rx_bytes' and 'tx_bytes' are present in the statistics.
- Neither of them is None, which allows for zero values (as per the comment).

Closes #595

### DIFF
--- a/openwisp_monitoring/device/tests/__init__.py
+++ b/openwisp_monitoring/device/tests/__init__.py
@@ -86,10 +86,11 @@ class TestDeviceMonitoringMixin(CreateConfigTemplateMixin, TestMonitoringMixin):
         data = self._transform_wireless_interface_test_data(data)
         self.assertDictEqual(dd_data, data)
 
-    def create_test_data(self, no_resources=False):
+    def create_test_data(self, no_resources=False, data=None, assertions=True):
         o = self._create_org()
         d = self._create_device(organization=o)
-        data = self._data()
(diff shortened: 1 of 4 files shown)


nemesifier (commented):
The tests weren't useful, they weren't failing without the fix.

Also, I think the fix should be to make sure we write the interface stats as long as we get either `rx_bytes` or `tx_bytes`, while the patch proposed here would write only if both are present.

I pushed a real fix, please have a look, you can try running tests with:

```
./manage.py test openwisp_monitoring.device.tests.test_api.TestDeviceApi.test_missing_rx_bytes openwisp_monitoring.device.tests.test_api.TestDeviceApi.test_missing_tx_bytes
```

If you change the code in `openwisp_monitoring/device/writer.py` to reset it to the previous state, you should see both tests failing. That's how we do it.

nemesifier (approved):

PR #607, opened 2024-08-02
[feature] Added support for device deactivation #560

### LINKED ISSUE #560: [feature] Deactivating device should also disable monitoring
Depends on https://github.com/openwisp/openwisp-controller/issues/625

Deactivating a device from controller should also deactivate the monitoring. This should have similar outcome as deactivating organization which includes: 

- Disabling monitoring checks 
- Setting monitoring status to "UNKNOWN". Or, we can use a special status for this e,g, "DEACTIVATED" 
- Replying with 404 for monitoring requests

pandafy:
Closes #560

**Blockers**

- [x] https://github.com/openwisp/openwisp-controller/pull/840


nemesifier (changes requested):
What happens to the monitoring status of a device when it gets deactivated?
I was expecting to see some logic which would see some logic for the following:

- Making sure monitoring checks are not run
- Setting monitoring status to "UNKNOWN". Or, we can use a special status for this e,g, "DEACTIVATED" - I think "DEACTIVATED" is more explicit and preferable for UX
- Replying with 404 for monitoring requests

I would also add that activating a device must reverse this: checks are run again, hence monitoring status should automatically become either OK or CRITICAL and device metrics shall be accepted.

We need tu update the docs to reflect these changes.

Please rebase on the latest master and reformat code if necessary.

nemesifier (changes requested):
### Color

Let's make the color of deactivate status black (`#000`), both in the health status (which is now of the same color of `unknown`) and in the pie chart:

Screenshot from 2024-11-12 18-23-38

### Device details: hide checks

Once the device is deactivated, I think we shouldn't show the checks in the device details anymore:

Screenshot from 2024-11-12 18-35-58

### Cache invalidation

Let's ensure that once the device is deactivated, the monitoring API responds 404 correctly, during testing we noticed a cache invalidation issue.

nemesifier (changes requested):
Let's adjust the colors here as discussed:

Screenshot from 2024-11-13 12-14-40

nemesifier (changes requested):
During testing, I noticed that reactivating a device leaves the health status as "DEACTIVATED", I think we should change this, the only option we have is "UNKNOWN" unless we add a new one, but I think "UNKNOWN" is good for now, what do you think?

PR #641, opened 2025-02-21
[change] Set status to "problem" when device is sending metrics #566

### LINKED ISSUE #566: [feature] Handle cases in which devices are receiving metrics but not reachable via ping
There can be a conflicting situation in which a device is not reachable on the management IP but is sending metrics succesfully to the server.

Due to the recovery detection feature, this generates additional load on the server because as soon as metrics or checksum requests are received, the system schedules a ping because it belive it will be able to reach the device and hence set the status back to OK, but that won't happen.

If many devices are in this situation, the monitoring queue can grow indefinitely until consuming all the available memory, at that point the server will crash.

We need to devise a way to spot these situations and set the status to "PROBLEM".

In this case, the ping check should not set the status to CRITICAL even if it cannot ping, unless no metrics were received for more than 10 minutes.

The device recovery mechanism should not be triggered if the status of the device is not critical.

Maybe we could solve this by simply modifying the ping check to look whether the device has been receiving monitoring metrics before deciding to set the status to CRITICAL or PROBLEM.

pandafy:
## Checklist

Closes #566

## Description of Changes

If the threshold is crossed for critical metrics but the device is sending data for passive checks, then the monitoring status will change to 'problem' instead of 'critical'.


nemesifier (commented):
Please try to make the code which checks the related metrics reusable, so we can execute it when the device recovery detection mechanism is triggered

We need to change the definition of problem, which now is:
"problem" means the device is having issues but it's still reachable;

After these changes we can't say the device it's still reachable, but we can say that it's still (at least partially) communicating with the server.

I think this is a change and not a feature, I recommend the title to reflect the fact that we are focusing on setting the status in a way that keeps into account other metrics.

pandafy:
The device recovery mechanism triggers checks in the following scenarios
1. Device requests checksum 
https://github.com/openwisp/openwisp-monitoring/blob/f67b6337cb4ea187cfaee24051fbed6476ec4e4c/openwisp_monitoring/device/apps.py#L207-L211
2. Metrics are received for the device 
https://github.com/openwisp/openwisp-monitoring/blob/f67b6337cb4ea187cfaee24051fbed6476ec4e4c/openwisp_monitoring/device/apps.py#L207-L211
3. When the device connection working status is changed 
https://github.com/openwisp/openwisp-monitoring/blob/f67b6337cb4ea187cfaee24051fbed6476ec4e4c/openwisp_monitoring/device/apps.py#L274-L282

> Please try to make the code which checks the related metrics reusable, so we can execute it when the device recovery detection mechanism is triggered

We planned to update the device recovery detection mechanism so a device could recover from `critical` state, if it starts sending monitoring metrics

If we do this, then the operation in point 2, will set the device monitoring status to 'problem'. But, operations in point 1 and 3 will still not change the monitoring status of the device since those operations don't write any metrics.

docs/user/settings.rst
```diff
@@ -754,6 +754,34 @@ This setting allows you to configure the config check interval used by
...
+``OPENWISP_MONITORING_DATA_COLLECTED_CHECK_INTERVAL``
+-----------------------------------------------------
+
+============ =======
+**type**:    ``int``
+**default**: ``60``
+============ =======
+
+This setting allows you to configure the data collected check interval
+used by :ref:`data collected <date_collected_check>`. By default it is set
```
nemesifier, line 782: date_collected_check ?

docs/user/checks.rst
```diff
@@ -35,6 +35,16 @@ configuration status of a device changes, this ensures the check reacts
 quickly to events happening in the network and informs the user promptly
 if there's anything that is not working as intended.
 
+.. _data_collected_check:
+
+Data Collected
```
nemesifier, line 40:
```suggestion
Monitoring Data Collected
```

I would just change the label but keep the internal naming as is for now.

docs/user/checks.rst
```diff
@@ -35,6 +35,16 @@ configuration status of a device changes, this ensures the check reacts
...
+--------------
+
+This check ensures that the :doc:`openwisp-monitoring agent
+</openwrt-monitoring-agent/index>` is running and sending passive metrics
```
nemesifier, line 44: "sending passive metrics" doesn't make sense IMHO: the agent sends metrics, the server receives it, so the server passively receives them, but the agent actively sends them, so saying the agent sends passive metrics seems contradictory to me.

openwisp_monitoring/device/base/models.py
```diff
@@ -384,12 +386,74 @@ def related_metrics(self):
...
+            and (
+                unhealthy_critical_metrics
+                == self.related_metrics.filter(
+                    key__in=self._get_critical_metric_keys(),
+                ).count()
+            )
+        ):
+            # All critical metrics are unhealthy
+            status = 'critical'
+        return status
```
nemesifier, line 449: I think the previous version of this was ok, except that it had to take into account for multiple metrics instead of just one.
This version is a lot more complex, are you 100% sure it's necessary?

openwisp_monitoring/device/base/models.py
```diff
@@ -399,21 +463,8 @@ def threshold_crossed(sender, metric, alert_settings, target, first_time, **kwar
             monitoring = target.monitoring
         except DeviceMonitoring.DoesNotExist:
             monitoring = DeviceMonitoring.objects.create(device=target)
-        status = 'ok' if metric.is_healthy else 'problem'
-        related_status = 'ok'
-        for related_metric in monitoring.related_metrics.filter(is_healthy=False):
-            if monitoring.is_metric_critical(related_metric):
-                related_status = 'critical'
-                break
```
nemesifier, line 407: if you remember our discussion, we said that we would have to avoid breaking here in order to set the status to critical only when all the critical metrics are unalthy.

nemesifier (changes requested):

openwisp_monitoring/device/base/models.py
pandafy: These are removed lines. The logic is added in `get_monitoring_status()`.

openwisp_monitoring/device/base/models.py
pandafy: This conditional ensures the following scenarios: 
1. If all critical metrics of the device is unhealthy, but the device is receiving passive metrics, then the status will be `problem` 
2. It only takes into account critical metrics for which the device has data. E.g. a new metric `data_collected` is added, but the device does not have any data for this metric, then only `Ping` check will be used to determine critical status.

nemesifier (commented):
We discussed that we have to update the definition of critical and problem in both the help_text and the docs, based on our discussions the gist is the following

- critical: the device is not communicating with the server at all
- problem: the device is either communicating with the server only partially or is experiencing some issue 

### Existing devices

Is the data collected check going to be created automatically for existing devices?

### Tests

I would leave existing tests which were designed for ping being the only critical metric unchanged and I would add a new set of tests which deal with the new scenario, making sure to name them properly so that the difference it's super obvious.

pandafy:
> Is the data collected check going to be created automatically for existing devices?

Yes, there's a [data migration](https://github.com/openwisp/openwisp-monitoring/blob/e71853e955489cfc794d3e49ba16bbe7864b7aa1/openwisp_monitoring/check/migrations/0010_create_data_collected_check.py).

openwisp_monitoring/device/base/models.py
```diff
@@ -384,12 +385,22 @@ def related_metrics(self):
             content_type__app_label='config',
         )
 
+    @classmethod
+    def get_active_metrics(cls):
```
nemesifier, line 389: A docstring here would be good, eg: returns a list of active metrics (Metric.key).

nemesifier (changes requested):
Good progress. I think the text in the docs needs more improvement, see my comments below.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -835,9 +830,9 @@ def test_status_critical_all_critical_metrics_unhealthy(self):
         dm.refresh_from_db()
         self.assertEqual(dm.status, 'critical')
 
-    def test_critical_metric_set_status_problem(self):
+    def test_unhealthy_critical_metric_set_status_problem(self):
         """
-        Tests scenario where critical metric sets status to 'problem'
+        Tests scenario where unhealthy critical metric sets status to 'problem'
```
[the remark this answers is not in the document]
pandafy: `Tests` here is a verb, i.e. `This test tests scenario where...` which is similar to `Returns critical metrics of the device`. 

I don't have strong opinion on this. 

According to docstring conventions, the first line of the docstring has to be one-liner summary.

openwisp_monitoring/device/tests/test_models.py
```diff
@@ -830,13 +830,13 @@ def test_status_critical_all_critical_metrics_unhealthy(self):
...
+    def test_single_unhealthy_critical_metric_set_status_problem(self):
         """
-        Tests scenario where unhealthy critical metric sets status to 'problem'
+        Tests that a single unhealthy critical metric sets the device status to 'problem'.
 
-        If threshold of critical metrics is crossed, but the device is
-        receiving data for other metrics, then the device status
-        will change to 'problem' instead of 'critical'.
+        If the threshold of a critical metric (ping) is exceeded, but the device is
+        still sending data for monitoring metrics, it's status should change to 'problem'
```
nemesifier, line 838: it's > its

not sure why github doesn't allow me to send code suggestions

docs/user/settings.rst
```diff
@@ -624,14 +624,24 @@ Automatically created charts.
...
+device's health status changes to ``CRITICAL``.
 
-By default, if devices are not reachable by pings they are flagged as
-``CRITICAL``.
+.. note::
+
+    If :ref:`openwisp_monitoring_auto_data_collected_check` is enabled,
+    then :ref:`monitoring data collected
+    <monitoring_data_collected_check>` metric is automatically added to
+    this list. Thus, the default value of the setting becomes
```
nemesifier, line 637: can't we just state the default value is the one below, without mentioning the rest, hence simplifying the text for the end-user?

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -349,6 +349,41 @@ def _get_access_tech():
...
     },
+    'data_collected': {
+        'label': _('Monitoring Data Collected'),
+        'name': 'Monitoring Data Collected',
+        'key': 'data_collected',
+        'field_name': 'data_collected',
+        'alert_settings': {'operator': '<', 'threshold': 1, 'tolerance': 30},
+        'notification': {
+            'problem': {
+                'verbose_name': 'Monitoring Data Collected PROBLEM',
```
nemesifier, line 360: shouldn't this be flagged as translatable?

openwisp_monitoring/monitoring/configuration.py
```diff
@@ -349,6 +349,41 @@ def _get_access_tech():
...
+                    '[{site.name}] PROBLEM: {notification.target} {notification.verb}'
+                    ' monitoring data'
+                ),
+                'message': _(
+                    'The [{notification.target}]({notification.target_link})'
+                    ' device {notification.verb} monitoring data.'
+                ),
+            },
+            'recovery': {
+                'verbose_name': 'Monitoring Data Collected RECOVERY',
```
nemesifier, line 373: Same

nemesifier (commented):
Looks good! We need to find out why all of a sudden tests are failing randomly.

nemesifier (approved):

openwisp_monitoring/check/migrations/0010_create_data_collected_check.py
```diff
@@ -0,0 +1,44 @@
...
+from openwisp_monitoring.check.settings import AUTO_DATA_COLLECTED_CHECK
+from openwisp_monitoring.check.tasks import auto_create_check
+
+
+def create_data_collected_check(apps, schema_editor):
+    if AUTO_DATA_COLLECTED_CHECK:
+        ContentType = apps.get_model('contenttypes', 'ContentType')
+        Check = apps.get_model('check', 'Check')
+        Device = apps.get_model('config', 'Device')
+        for device in Device.objects.all():
```
nemesifier, line 13:
```suggestion
        for device in Device.objects.iterator():
```

nemesifier (changes requested):

nemesifier (approved):

PR #647, opened 2025-03-12
[fix:tests] Fixed flaky selenium tests

nemesifier:
## Checklist

## Description of Changes

Added more waits before trying to click on elements. Wait for loading spinner to disappear before clicking save.


nemesifier (commented):
**Update on what I learnt today**:

- stop hiding dust under the carpet: if tests are failing most of the times it likely indicates an underlying issue or something we haven't understood about selenium or the driver in use, each driver has its peculiarities
    - geckodriver doesn't seem to play well with the loading overlay JS fadeout mechanism, it belives is not visible anymore and hence proceeds to click(), only to fail because the button is still covered. In this case turning off animations or explicitly hiding the overlay in my opinion is an acceptable solution
    - I don't see a reason to keep code which automatically accepts "unsaved changes" alerts, I think this is hiding dust under the carpet
    - I am also not a fan of accepting/silencing errors and in the monitoring PR I was finally able to avoid it once I understood the root cause (which is that the application is not ready to proceed while the web driver goes very fast)
- I don't think we should run selenium tests for `SAMPLE_APP=1` or other test runs like `UDP=1` in monitoring, because those runs are focused on testing backend features rather than frontend features
- we must be strict with the enforcement of best practices in selenium tests (use helper methods and add more helper methods if we find ourselves repeating  the same actions in different modules)
- **in general most problems seem to be related to the fast pace of requests and actions of automated tests, which is something the application is not designed for, if helper methods don't suffice, we need to either add explicit waits (eg: `time.sleep(0.5)` or remove anything that annoys automated testing and is not essential (like animations)**

PR #651, opened 2025-03-24
[change] Only trigger critical checks on device recovery #642

### LINKED ISSUE #642: [bug] Recovery detection mechanism triggers unwanted active checks
The recovery detection mechanism triggers all the checks: config check, iperf and max/min wifi clients. This is undesirable and can lead to inefficiency on busy installations which use these checks.

https://github.com/openwisp/openwisp-monitoring/blob/f67b6337cb4ea187cfaee24051fbed6476ec4e4c/openwisp_monitoring/device/tasks.py#L31

This is the entire logic of the celery task:

https://github.com/openwisp/openwisp-monitoring/blob/f67b6337cb4ea187cfaee24051fbed6476ec4e4c/openwisp_monitoring/device/tasks.py#L15-L38

Ideally we want only run checks which are related to critical metrics. The problem is that it seems we don't have an easy way to filter by checks which write critical metrics.
If we don't want to change the DB structure to track metrics which are related to a check, probably we can get away with editing the critical metric settings to allow an optional check key which points to the check class:

```python
critical_metrics = get_settings_value(
        'CRITICAL_DEVICE_METRICS',
        [
            {
                'key': 'ping',
                'field_name': 'reachable',
                'check': 'openwisp_monitoring.check.classes.Ping',
            }
        ],
    )
```

Then we loop over these critical metrics, extract the check classes in a `critical_checks` list and change the queryset like:

```python
critical_checks 
checks = device.checks.filter(is_active=True, check_type__in=critical_checks).only('id').values('id')
```
Do we have to add a composite index on both `is_active` and `check_type` to ensure lookup efficiency here? Not sure.

It may also be worth it to rename the functions/methods, eg: `trigger_device_critical_checks` or similar.

pandafy:
## Checklist

Closes #642


openwisp_monitoring/device/base/models.py
```diff
@@ -398,6 +398,19 @@ def get_active_metrics(cls):
             cls._active_metrics = active_metrics
         return cls._active_metrics
 
+    @classmethod
+    def get_critical_checks(cls):
```
pandafy, line 402: We can easily loop over the `CRITICAL_DEVICE_METRICS` list in the device/settings.py. 

I chose to add a class method, to make the methods more consistent. Now, we have two classmethod for the DeviceMonitoring model: 

- DeviceMonitoring.get_active_metrics()
- DeviceMonitoring.get_critical_checks()


What do you think @nemesifier? Shall we keep this or move the logic to app settings?

openwisp_monitoring/check/base/models.py
```diff
@@ -49,6 +49,7 @@ class AbstractCheck(TimeStampedEditableModel):
     class Meta:
         abstract = True
         unique_together = ('name', 'object_id', 'content_type')
+        indexes = [models.Index(fields=['content_type', 'object_id', 'is_active'])]
```
pandafy, line 52: # Index Optimization Summary

**TL;DR**: Added an index on `['content_type', 'object_id', 'is_active']` to optimize queries on active device checks. This index significantly improved query performance by enabling an index scan instead of a bitmap heap scan.

## Methodology

- Created 3000 rows in the `check_check` table for testing.
- Tested these indexing strategies:
  1.  No additional index
  2.  Index on `['check_type', 'is_active']`
  3.  Index on `['content_type', 'object_id', 'check_type', 'is_active']`
  4.  Final: Index on `['content_type', 'object_id', 'is_active']`
- Compared query execution plans using Django's `.explain()` with PostgreSQL database.

## Results

**No index / default:**

``` sql
Bitmap Heap Scan on check_check
  Recheck Cond: (object_id = '...')
  Filter: (is_active AND check_type IN (...) AND content_type_id = 24)
  -> Bitmap Index Scan on check_check_object_id_c8edcb05_like
```

**Index on \`\`check_type\`\`, \`\`is_active\`\`:**

``` sql
Bitmap Heap Scan on check_check
  Recheck Cond: (object_id = '...')
  Filter: (is_active AND check_type IN (...) AND content_type_id = 24)
  -> Bitmap Index Scan on check_check_object_id_c8edcb05_like
```

**Index on \`\`content_type\`\`, \`\`object_id\`\`, \`\`check_type\`\`, \`\`is_active\`\`:**

``` sql
Index Scan using check_check_content_b49654_idx on check_check
  Index Cond: (content_type_id = 24 AND object_id = '...' AND is_active = true)
  Filter: (check_type IN (...))
```

**Final choice – Index on \`\`content_type\`\`, \`\`object_id\`\`, \`\`is_active\`\`:**

``` sql
Index Scan using check_check_content_b49654_idx on check_check
  Index Cond: (content_type_id = 24 AND object_id = '...' AND is_active = true)
  Filter: (check_type IN (...))
```

## Conclusion

The chosen index uses the most selective fields, enabling efficient index scans while keeping the index size small. It balances performance and simplicity better than wider indexes including `check_type`.

pandafy:
The CI is failing because of this bug in setuptools. 

https://github.com/pypa/setuptools/issues/4910

openwisp_monitoring/device/tasks.py
```diff
@@ -28,7 +29,11 @@ def trigger_device_checks(pk, recovery=True):
     except ObjectDoesNotExist:
         logger.warning(f'The device with uuid {pk} has been deleted')
         return
-    check_ids = list(device.checks.filter(is_active=True).values_list('id', flat=True))
+    check_ids = list(
+        device.checks.filter(
+            is_active=True, check_type__in=device.monitoring.get_critical_checks()
```
nemesifier, line 34: Does this query trigger the index use too?

openwisp_monitoring/device/base/models.py
nemesifier: Looks good :+1:

openwisp_monitoring/check/base/models.py
nemesifier: Great, thanks for the analysis!

openwisp_monitoring/device/tasks.py
pandafy: Yes, I did the analysis for this query. 

```sql
Index Scan using check_check_content_b49654_idx on check_check
  Index Cond: (content_type_id = 24 AND object_id = '...' AND is_active = true)
  Filter: (check_type IN (...))
```

nemesifier (approved):

PR #656, opened 2025-05-05
[docs] Added documentation for default alert settings #611

### LINKED ISSUE #611: [docs] Add information about default alert settings
We need a section which explains the default alerts and their default settings. It seems to me that this is missing now.

pandafy:
## Checklist

- [ ] I have manually tested the changes proposed in this pull request.

Closes #611


nemesifier (changes requested):
Looks good, let's refine a bit.

Let's be more explicit in mentioning which alerts are enabled by default and which one aren't, I don't see this information now. What about the settings to automatically create the checks? I think they should be mentioned here.

Don't we also have built-in alerts for the iperf check? It should be possible to set alert settings for the iperf metrics right? It could be worth a mention.

Please also add links to this page or sections of this page from other key pages, it's worth pointing to this piece of docs (eg: in the iperf page or in the wifi min/max docs).

pandafy:
> Don't we also have built-in alerts for the iperf check? It should be possible to set alert settings for the iperf metrics right? It could be worth a mention.

No, the `iperf3` check does not have any default notification. https://openwisp.io/docs/dev/monitoring/user/device-checks-and-alert-settings.html explains how to add that. 

> What about the settings to automatically create the checks? I think they should be mentioned here.

Wouldn't this be redundant? I have added notes to mention that the check should be enabled for receiving the alert. 

Screenshot from 2025-05-06 16-15-04

nemesifier (approved):

nemesifier:
Great, thanks!

PR #669, opened 2025-05-21
[fix] Don't evaluate threshold for historical data #666

### LINKED ISSUE #666: [bug] Do not send alert for old metric data
**Description:**
The OpenWrt monitoring agent stores metrics data locally when it is unable to establish a connection with OpenWISP. Once the connection is restored, the agent uploads all the stored data retroactively.

This causes OpenWISP Monitoring to process and store historical data that is no longer current. If any of these old metric values exceed the configured threshold, the system triggers alerts as if the condition is recent. This behavior leads to misleading and unnecessary notifications because the device may have already recovered during the period of disconnection.

**Expected Behavior:**
Metric.check_threshold() should only evaluate and trigger alerts for recent data points. Alerts should not be sent for stale or historical data.

**Suggested Solution:**
Implement logic to skip threshold checks for data points older than a certain time period. We need to decide on this time period.I am thinking, it can be **twice the tolerance time period**.

pandafy:
## Checklist

Fixes #666


openwisp_monitoring/monitoring/base/models.py
```diff
@@ -942,6 +946,15 @@ def operator(self):
             return self.config_dict['operator']
         return self.custom_operator
 
+    def is_historical_data(self, time):
+        """
+        Data older than 1 hour is considered historical data.
+        """
+        if not time or not isinstance(time, datetime):
+            raise ValueError('Invalid time value')
+        recent_time = timezone.now() - timedelta(minutes=60)
```
pandafy, line 955: I have chosen 1 hour arbitrarily. We can change this if needed.

PR #677, opened 2025-06-05
[feature] Add dedicated map page

### LINKED ISSUE #561: [feature] Add a dedicated page for showing devices on the map
The goal is to add a dedicated view for viewing devices on the map. We have such a map on the dashboard, but it's size constraints (due to other UI elements) affects it's usability. 

A dedicated page for using viewing all the devices on map would elevate such UX constraints. 

This is a rough idea of what we want to achieve

Screenshot from 2024-02-06 15-55-48

dee077:
## Checklist

Closes #561.

## Description of Changes
- Added a minimal Template to render a new map page
- Added a view to set the route to /map to render the newly added template.
- Added a map to the side menu, redirecting it to `/map`. 

## To do
- Structure the code properly
- Add js to include the `Netjsongraph` map in the newly added `map_page.html` template.


openwisp_monitoring/device/admin.py
```diff
@@ -573,3 +574,48 @@ def has_delete_permission(self, request, obj=None):
...
+
+def _patched_get_urls():
+    custom_urls = [
+        path(
+            "map/",
+            admin_site.admin_view(DeviceMapView.as_view()),
+            name="monitoring_device_map",
+        )
+    ]
+    return custom_urls + _original_get_urls()
```
pandafy: I am not sure about monkey patching the AdminSite instance. Maybe, we can add this URL to device ModelAdmin? Alternatively, we can add this in urls.py but wrap it in `admin_site.admin_view()` so it has the required checks for authentication and permission. 

Talking about permissions: who can view this map? Do we need to create special permission objects for this?

IMO. users which can view device should be able to view this map. 


@nemesifier what do you think?

nemesifier: I agree with everything you wrote @pandafy.

Regarding permissions to view the map, let's make sure it's consistent with what we show in the dashboard.
I am not sure if we do any permission checking now.

pandafy: > Regarding permissions to view the map, let's make sure it's consistent with what we show in the dashboard.
> I am not sure if we do any permission checking now.

We are not doing any special permission checks for the dashboard map.

nemesifier:
As agreed on the [GSoC-25 matrix chat](https://matrix.to/#/#gsoc2025-openwisp/general-map:matrix.org), we are closing this for now, but @dee077 will reopen it asap when other dependencies to implement the functionalities needed here will be ready (netjsongraph.js).

PR #680, opened 2025-06-12
[feature] Add floorplan map above dashboard map #564

### LINKED ISSUE #564: [feature] Allow to drill down to indoor map (floor plan) from map
We need to allow viewing indoor map (floorplan) for locations directly from the map. We should also show all the floors available for that location, and show floorplan for the selected floor.

dee077:
## Checklist

Closes #564.

## Description of Changes

- A floorplan button was added below the device table.
- On click, a new container is created, and the floorplan map is rendered using `Netjsongraph`.
- Currently using static sample data from Netjsongraph's examples.
- Will integrate dynamic data fetching via REST API and render on AJAX success.

## Screenshot
[Screencast from 2025-06-13 00-28-12.webm](https://github.com/user-attachments/assets/a6f61243-ff94-4c14-bb49-10577f89c36e)


openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -29,16 +32,31 @@
...
+}
+.floorplan-btn:hover {
+  background-color: #000000;
+  color: #fff;
+}
 #device-map-container {
   margin: 50px auto;
   position: relative;
-  height: 300px;
+  height: 80vh;
```
pandafy: Why are we increasing the height of the dashboard map?

pandafy: If this is only for testing, please add a TODO comment to remind us to remove this before merging.

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -29,16 +32,31 @@
...
+  color: #fff;
+  border-radius: 5px;
+  text-decoration: none;
+  cursor: pointer;
+  font-size: 14px;
+}
+.floorplan-btn:hover {
+  background-color: #000000;
+  color: #fff;
+}
```
pandafy: Please don't add any new styles for button, especially related to font, colors, etc. We already have a `button` class in openwisp-utils. Use it instead. 

https://github.com/openwisp/openwisp-utils/blob/8e8f29887d79f7ce8b30a29c338a47ba38bf4447/openwisp_utils/admin_theme/static/admin/css/openwisp.css#L542C1-L566C2

Feel free to add customization related to padding, margin, etc.

openwisp_monitoring/device/api/serializers.py
```diff
@@ -154,3 +156,22 @@ class Meta:
             "stop_time",
             "modified",
         ]
+
+
+class MonitoringFloorplanCoordinatesSerializer(FloorplanCoordinatesSerializer):
+    status = serializers.CharField(source="content_object.monitoring.status")
+
+    class Meta:
+        model = DeviceLocation
```
pandafy: If you extend the Meta class of `FloorplanCoordinatesSerializer`, then you won't need to re-define model here. 
```suggestion
    class Meta(FloorplanCoordinatesSerializer.Meta):
```

openwisp_monitoring/device/api/serializers.py
```diff
@@ -154,3 +156,22 @@ class Meta:
             "stop_time",
             "modified",
         ]
+
+
+class MonitoringFloorplanCoordinatesSerializer(FloorplanCoordinatesSerializer):
```
pandafy: Nitpick: I would move this just after `DeviceMonitoringLocationSerializer`, to group similar functionalities together in the code.

openwisp_monitoring/device/api/urls.py
```diff
@@ -40,4 +40,9 @@
         views.wifi_session_detail,
         name="api_wifi_session_detail",
     ),
+    path(
+        "api/v1/monitoring/location/<str:pk>/floorplan_coordinates/",
+        views.floorplan_coordinates_list,
+        name="floorplan_coordinates_list",
+    ),
```
pandafy: **Thinking out loud**

We have this **existing** API endpoint which returns all the devices on a location 

```
/api/v1/monitoring/location/<pk>/device/
```

Thus, we can call this endpoint

```
/api/v1/monitoring/location/<pk>/floorplan-devices/
```

pandafy: @dee077 don't make this change yet. This is up for discussion. It is a small change, we can also do it later.

pandafy (changes requested):
@dee077 we don't need for https://github.com/openwisp/netjsongraph.js/pull/381 to get merged. You can build your branch of netjsongraph.js locally and copy the netjsongraph.min.js in openwisp-monitoring. Please do this in a separate commit. 

https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/static/monitoring/js/lib/netjsongraph.min.js

While testing, I noticed a bug which changed the width of the dashboard map container. @cestercian is already working on a fix in #335. You can either override the required CSS here temporarily till that PR is merged. 

This worked for me 
```diff
diff --git a/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css b/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
index 05cd643..c23870d 100755
--- a/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
+++ b/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
@@ -9,7 +9,7 @@
   border: 1px solid #ccc;
   color: #6d6357;
   line-height: 20px;
-  max-width: 400px;
+  /* max-width: 400px; */
   min-width: 200px;
   padding: 10px 15px;
 }
```

On narrower screens, the overlay shows a horizontal scroll bar. This is inconsistent with the handling of map on narrow screens. The horizontal scroll should not appear.  The overlay of the floorplan widget, should automatically adjust to the dimensions of the map container. 

Screenshot from 2025-06-23 18-37-33

I understand that right now we are focusing on the API changes. So, I am not pointing out changes required in the UI. We aim to develop a similar UI to the mockups.

openwisp_monitoring/device/api/urls.py
nemesifier: I believe the resource in this case it's the indoor coordinates of the devices, therefore we could use `/api/v1/monitoring/location/<pk>/indoor-coordinates/`, I think it's ok to omit the word "device" due to the fact that it's implicit as there's no other object which can be placed on indoor maps.

The convention is to use the singular name, unless the singular name is also plural as in this case (coordinates).

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -106,9 +109,9 @@
                                     </tbody>
                                 </table>
                                 ${pagination}
+                                <button class="default-btn floorplan-btn">Floorplan</button>
```
pandafy: Let's only show this button if the location has an indoor co-ordinate. Do you think it is possible to do that?

pandafy: Alternatively, we can render the floorplan button disabled and add a tooltip that the location does not have any indoor devices.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,209 @@
...
+      selectedIndex === floors.length - 1,
+    );
+
+    $(".floor-btn").removeClass("active selected");
+    $('.floor-btn[data-index="' + selectedIndex + '"]').addClass(
+      "active selected",
+    );
+  }
+
+  function bindNavigationHandlers(data) {
```
pandafy: Screenshot from 2025-07-01 16-34-51

I added two floors in a location, 0 and 19. I observed two issues in the above screenshot

1. The floor selection widget is very far from the floorplan overlay. Let's decrease the spacing there such that it does not change on different screen sizes. 
2. Floor 19 comes below floor 0. I was expected it the other way: 
```
[19]
[0]
```

pandafy (changes requested):
@dee077 whenever you create a PR which depends on a different PR in another module, please update the requirements.txt file like below so installing openwisp-monitoring using pip would automatically install the correct version of openwisp-controller. 

```diff 
diff --git a/requirements.txt b/requirements.txt
index 6f4b6b0..d1c706f 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
-openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/1.2
+openwisp-controller @ https://github.com/dee077/openwisp-controller/tarball/feature/828-api-indoor-map-coordinates
 influxdb~=5.3.2
 django-nested-admin~=4.1.1
 python-dateutil>=2.7.0,<3.0.0
```

Without this, CI would fail. We didn't encounter this yet since we don't have CI enabled on the `gsoc25-map` branch yet. 

I noticed an issue in the UI
**Steps to replicate**
1. Open a floorplan and interact with the floorplan (zoom in/zoom out)
2. Close the floorplan 
3. Zoom out the geographic map


https://github.com/user-attachments/assets/13cfe5d0-f18a-4ec9-b520-2bf1114db656

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -1,6 +1,16 @@
...
+  const colors = {
+    ok: "#267126",
+    problem: "#ffb442",
+    critical: "#a72d1d",
+    unknown: "#353c44",
+    deactivated: "#0000",
+  };
+  function getColor(status) {
+    return colors[status] || colors.unknown;
+  }
```
pandafy: We already have the colors defined for the dashboard map. Is there a way we can centralize the colors? Can you look into using CSS/JS variables?

openwisp_monitoring/device/api/views.py
```diff
@@ -246,13 +246,32 @@ class MonitoringGeoJsonLocationList(GeoJsonLocationList):
...
+
+    def get_available_floors(self):
+        location_id = self.request.parser_context["kwargs"].get("pk")
+        floors = (
+            DeviceLocation.objects
+            .filter(location__id=location_id, floorplan__isnull=False)
+            .values_list("floorplan__floor", flat=True)
+            .distinct()
+            .order_by("floorplan__floor")
+        )
```
pandafy: This will bypass the filtering done by the `MonitoringLocationDeviceList` view. Is there a way to access the queryset of the view here?

dee077: Updated the logic!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Will work on it when completed with the endpoint integration

openwisp_monitoring/device/api/views.py
```diff
@@ -246,32 +246,26 @@ class MonitoringGeoJsonLocationList(GeoJsonLocationList):
...
+        return qs
+
+    def list(self, request, *args, **kwargsf):
+        qs = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(qs)
+        serializer = self.get_serializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        response.data['has_floorplan'] = self.get_has_floorplan(qs)
+        return response
+
```
pandafy: Can we rewrite this as the following?

```
def list(self, request, *args, **kwargs): 
    response = super().list(request, *args, **kwargs)
    if response.status_code == 200: 
        response.data['has_floorplan'] = self.get_has_floorplan(qs)
    return response
```

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -211,13 +206,23 @@
...
     $.ajax({
-      url,
+      url: `${url}?${floor}`,
       method: "GET",
       dataType: "json",
       xhrFields: { withCredentials: true },
-      success: (data) => onAjaxSuccess(data),
+      success: (data) => {
+        // data append
+        let allResults;
```
pandafy: This is not appending data, is it?

dee077: This is a work in progress!

openwisp_monitoring/device/api/views.py
dee077: Updated!

openwisp_monitoring/device/api/serializers.py
```diff
@@ -46,6 +48,24 @@ class MonitoringLocationDeviceSerializer(LocationDeviceSerializer):
...
+            "content_object_id",
+            "floorplan_id",
+            "device_name",
+            "status",
+            "mac_address",
+            "floor_name",
+            "floor",
+            "image",
+            "location",
+        ]
```
pandafy: You don't need to redefine this here. Just do it like `MonitoringLocationDeviceSerializer`. 

Check the output of `MonitoringLocationDeviceList` view: 

```json 
{
            "id": "ec513d52-284a-43a0-ba8b-229845f2c1c1",
            "admin_edit_url": "http://127.0.0.1:8000/admin/config/device/ec513d52-284a-43a0-ba8b-229845f2c1c1/change/",
            "monitoring": {
                "status": "critical",
                "status_label": "critical"
            },
            "created": "2024-08-27T09:55:58+02:00",
            "modified": "2025-07-03T18:22:13.080081+02:00",
            "name": "04-pandafy-vm",
            "mac_address": "08:00:27:7B:89:FC",
            "key": "0435e3fa6bd52f71057a798c9a8d755c",
            "model": "innotek GmbH VirtualBox",
            "os": "OpenWrt 21.02.2 r16495-bf0c965af0",
            "system": "Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz",
            "notes": "",
            "last_ip": "192.168.56.2",
            "management_ip": "192.168.56.2",
            "hardware_id": null,
            "_is_deactivated": false,
            "organization": "57197e42-b7a9-4342-b1ef-672d7fd6ed59",
            "group": null
        }
```

The monitoring status is present in a dictionary with two keys `status` and `status_label`. Let's make this serializer consistent with that. 

Moreover, we allow to override the `status_label` using [OPENWISP_MONITORING_HEALTH_STATUS_LABELS
](https://openwisp.io/docs/dev/monitoring/user/settings.html#openwisp-monitoring-health-status-labels), so if we only provide the status value in the response, we won't be able to render the proper value in the frontend.

openwisp_monitoring/device/api/serializers.py
```diff
@@ -12,6 +13,7 @@
 DeviceMonitoring = load_model("device_monitoring", "DeviceMonitoring")
 DeviceData = load_model("device_monitoring", "DeviceData")
 Device = load_model("config", "Device")
+DeviceLocation = load_model("geo", "DeviceLocation")
```
pandafy, line 16: This is no longer used, let's remove it.

openwisp_monitoring/device/api/views.py
```diff
@@ -248,11 +251,60 @@ class MonitoringLocationDeviceList(LocationDeviceList):
...
+        serializer = self.get_serializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        if response.status_code == 200:
+            response.data['has_floorplan'] = self.get_has_floorplan(qs)
+        return response
 
 
 monitoring_location_device_list = MonitoringLocationDeviceList.as_view()
 
+# Todo: Remove this
```
pandafy, line 260: Thanks for this!

openwisp_monitoring/device/api/views.py
```diff
@@ -248,11 +251,60 @@ class MonitoringLocationDeviceList(LocationDeviceList):
...
+    )
+    serializer_class = MonitoringIndoorCoordinatesSerializer
+
+    def list(self, request, *args, **kwargs):
+        qs = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(qs)
+        serializer = self.get_serializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        if response.status_code == 200:
+            response.data['floors'] = self.get_available_floors()
```
pandafy: Why don't we just do this in the controller? Let's keep this view lean and only add logic here which is relevant to openwisp-monitoring.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,261 @@
...
+
+  async function openFloorPlan(url) {
+    await fetchData(url);
+
+    selectedIndex = floors.indexOf(currentFloor) || 0;
+    const maxStart = Math.max(0, floors.length - NAV_WINDOW_SIZE);
+    navWindowStart = Math.min(
+      Math.floor(selectedIndex / NAV_WINDOW_SIZE) * NAV_WINDOW_SIZE,
+      maxStart
+    );
```
pandafy: Please add a comment about what this code does.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,261 @@
...
+    await showFloor(url, currentFloor);
+  }
+
+  function fetchData(url, floor = null) {
+    const reqUrl = floor ? `${url}?floor=${floor}` : url;
+    return new Promise((resolve, reject) => {
+      if (floor && allResults.nodes.some((n) => n.floor === floor)) {
+        resolve();
+        return;
+      }
```
pandafy: Can you help me understand what does this code do?

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,261 @@
...
+          if (!floor) floors = data.floors;
+          allResults = {
+            ...allResults,
+            nodes: [...allResults.nodes, ...data.results],
+          };
+          if (!currentFloor && data.results.length) {
+            currentFloor = data.results[0].floor;
+          }
+          if (data.next) {
+            await fetchData(data.next, floor);
```
pandafy: The `next` URL should automatically have the `floor` query param set.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,261 @@
...
+    const maxStart = Math.max(0, floors.length - NAV_WINDOW_SIZE);
+    navWindowStart = Math.min(
+      Math.floor(selectedIndex / NAV_WINDOW_SIZE) * NAV_WINDOW_SIZE,
+      maxStart
+    );
+
+    const $floorPlanContainer = createFloorPlanContainer();
+    const $floorNavigation = createFloorNavigation();
+
+    $("#device-map-container").append($floorPlanContainer);
```
pandafy, line 33: `#device-map-container` this container is used by the netjsongraph.js library in device-map.js. Instead of appending to this element, can we add the floorplan after this?

I wonder if this causes the issue with the device map.

dee077: Yes I have tried to add it in diffrent divs and even directly to body issue still persists

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: It checks that the floor is passed and there exists a value in all results, which is equal to that passed floor. 
so that we can return early dont need the make an api call

openwisp_monitoring/device/api/views.py
dee077: Sure!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
pandafy: What is the benefit of does this? Don't we trust the API to return floors on the same floor?

openwisp_monitoring/device/static/monitoring/js/floorplan.js
pandafy: okay!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: If someone clicks on the same floor again, it checks if the floor is present in allResults. If it is, then it means it has already been fetched, so there is no need to make the API call again. Therefore, we return early.

pandafy: Sweet! Let's add a comment here.

dee077: Sure!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Added!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/api/serializers.py
dee077: Updated!

openwisp_monitoring/device/api/urls.py
dee077: Updated to `indoor-coordinates`

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated! Showing button only if there exists a floorplan

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/api/serializers.py
dee077: Removed!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/api/serializers.py
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Added Comment with an example let me know if it is too much

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/api/serializers.py
```diff
@@ -46,6 +48,15 @@ class MonitoringLocationDeviceSerializer(LocationDeviceSerializer):
     monitoring = DeviceMonitoringLocationSerializer()
 
 
+class MonitoringIndoorCoordinatesSerializer(IndoorCoordinatesSerializer):
+    monitoring = monitoring = DeviceMonitoringLocationSerializer(
```
pandafy, line 52:
```suggestion
    monitoring = DeviceMonitoringLocationSerializer(
```

openwisp_monitoring/device/api/views.py
```diff
@@ -248,12 +251,53 @@ class MonitoringLocationDeviceList(LocationDeviceList):
     serializer_class = MonitoringLocationDeviceSerializer
 
     def get_queryset(self):
-        return super().get_queryset().select_related("monitoring").order_by("name")
+        qs = super().get_queryset().select_related("monitoring").order_by("name")
+        return qs
+
+    def get_has_floorplan(self, qs):
+        qs = qs.filter(devicelocation__floorplan__isnull=False).exists()
+        return qs
```
pandafy: Didn't we decide to move all floorplan logic to controller? Is this a leftover?

openwisp_monitoring/device/api/views.py
```diff
@@ -248,12 +251,53 @@ class MonitoringLocationDeviceList(LocationDeviceList):
     serializer_class = MonitoringLocationDeviceSerializer
 
     def get_queryset(self):
-        return super().get_queryset().select_related("monitoring").order_by("name")
+        qs = super().get_queryset().select_related("monitoring").order_by("name")
+        return qs
```
pandafy: Why this change?

openwisp_monitoring/device/api/views.py
```diff
@@ -248,12 +251,53 @@ class MonitoringLocationDeviceList(LocationDeviceList):
...
+
+    def list(self, request, *args, **kwargsf):
+        qs = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(qs)
+        serializer = self.get_serializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        if response.status_code == 200:
+            response.data["has_floorplan"] = self.get_has_floorplan(qs)
+        return response
```
pandafy:
```suggestion
    def list(self, request, *args, **kwargs):
        response = super().list(self, request, *args, **kwargs)
        if response.status_code == 200:
            response.data["has_floorplan"] = self.get_has_floorplan(qs)
        return response
```

I'd also move this logic to the controller.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -6,13 +6,17 @@
   const localStorageKey = "ow-map-shown";
   const mapContainer = $("#device-map-container");
   const statuses = ["critical", "problem", "ok", "unknown", "deactivated"];
-  const colors = {
+  window.STATUS_COLORS = {
```
pandafy:
```suggestion
  window._owGeoMapConfig.STATUS_COLORS = {
```

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,262 @@
...
+      },
+
+      prepareData(data) {
+        data.nodes.forEach((node) => {
+          // To hide DeviceLocation id in tooltip
+          node.id = "";
+          node.properties = {
+            ...node.properties,
+            name: node.device_name,
+            status: "critical",
```
pandafy: Why do we still need to hard-code this here? I think we are at a good point where we can use the real value from the API response.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,262 @@
...
+
+  async function showFloor(url, floor) {
+    $("#floorplan-navigation .floor-btn")
+      .removeClass("active selected")
+      .filter(`[data-floor="${floor}"]`)
+      .addClass("active selected");
+
+    await fetchData(url, floor);
+
+    const nodesThisFloor = allResults.nodes.filter((n) => n.floor === floor);
```
pandafy: Do we merge all data in `allResults`? Ideally. we should keep the data in some kind of map 

```
{
  1: []
  2" []
}
```

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,262 @@
...
+        url: reqUrl,
+        method: "GET",
+        dataType: "json",
+        xhrFields: { withCredentials: true },
+        success: async (data) => {
+          if (!floor) floors = data.floors.sort((a, b) => b - a);
+          allResults = {
+            ...allResults,
+            nodes: [...allResults.nodes, ...data.results],
+          };
```
pandafy: This is bad! We get filtered data from the API, then we merge it with everything to filter again later? Let's improve this logic and maintain data for each floor separately.

pandafy (changes requested):

openwisp_monitoring/device/api/views.py
dee077: Did make this in indoor coordinates as this is a different endpoint
Now moved it to controller!

openwisp_monitoring/device/api/views.py
dee077: Removed the changes!

openwisp_monitoring/device/api/views.py
dee077: Updated in controller!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Make allResults as
```
{
-1: [list of all devices of floor -1],
0:  [list of all devices of floor 0],
1:  [list of all devices of floor 1],
}
```

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated the logic
```
if (!allResults[floor]) allResults[floor] = []
allResults[floor] = [...allResults[floor], ...data.results];
```

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -54,11 +54,14 @@
...
-            ...allResults,
-            nodes: [...allResults.nodes, ...data.results],
-          };
+          // To make this run only one time as only in the first call floor will not be provided
+          // And sort them in decresing order so that neagtive floor show at bottom and postove at top in floor navigation
+          if (!floor){
+            floors = data.floors.sort((a, b) => b - a);
+            floor = data.results[0].floor
+          }             
+          if (!allResults[floor]) allResults[floor] = []
```
pandafy, line 63: I would add a comment here.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -251,7 +253,14 @@
...
           map.fitBounds(bnds);
           map.setMaxBounds(bnds);
+          map.setView([0,0], 0)
         };
+        // Todo: Explore some better approach if exists
+        const fullScreen = $('#floorplan-container .leaflet-control-fullscreen-button');
+        fullScreen.off("click.zoomOnFullscreen"); // avoid duplicates
+        fullScreen.on("click.zoomOnFullscreen", () => {
+            map.setZoom(map.getZoom() + 1);
+        });
```
pandafy: Can you add a comment explaining what we are trying to achieve here?

pandafy: https://github.com/user-attachments/assets/87e6c1ba-1f1b-4b63-b572-8611b6244527

The zoom level changes drastically when entering/exiting the fullscreen mode.

dee077:
Closing as #688 is built on top of this PR

PR #688, opened 2025-07-18
[feature] Improve device popup UI and filtering #662

### LINKED ISSUE #662: [feature] Update Ui of pop up shown to list deivces on geo map
## Feature Description
Update the ui for the pop-up shown to list devices above the geo map as presented by ui/ux team in [Figma](https://www.figma.com/design/0eaWAMSiEAgfHMIjIVRv2F/Open-WISP-Bridgebeam?t=I6mHbSTDjUu6dCxm-0)

- Should have a filter option to get any specific device by name or MAC address.
- Have a filter based on the monitoring health status.

**Sample Image:**

Image

dee077:
## Checklist

Closes #662.

## Description of Changes
- Improved the ui accoding to the [figma design](https://www.figma.com/design/0eaWAMSiEAgfHMIjIVRv2F/Open-WISP-Bridgebeam?node-id=1087-27738&t=3N4CgFs7GfLu6Y8d-0)
- Added filter based on device name, mac address, and monitoring health status.

## Pending:
- [ ] The floorplan icon is not showing on the Switch to floor plan button.
- [ ] Correct filtering logic to filter with device name, mac address, and monitoring health status. 

## Screenshot

image

### DIFF
--- a/openwisp_monitoring/device/api/filters.py
+++ b/openwisp_monitoring/device/api/filters.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
 from django_filters import rest_framework as filters
 from swapper import load_model
@@ -21,6 +22,27 @@ class WifiSessionFilter(FilterDjangoByOrgManaged):
             "start_time": ["exact", "gt", "gte", "lt", "lte"],
             "stop_time": ["exact", "gt", "gte", "lt", "lte", "isnull"],
         }
+
+
+class DeviceFilter(filters.FilterSet):
+    search = filters.CharFilter(method="filter_search")
+    status = filters.CharFilter(method="filter_by_status")
+
+    def filter_search(self, queryset, name, value):
+        value = value.strip()
+        return queryset.filter(
+            Q(name__icontains=value) | Q(mac_address__icontains=value)
+        )
+
+    def filter_by_status(self, qs, name, value):
+        statuses = [s.strip() for s in value.split(",") if s.strip()]
+        if statuses:
+            return qs.filter(monitoring__status__in=statuses)
+        return qs
+
+    class Meta:
+        model = Device
+        fields = []
 
 
 class MonitoringDeviceFilter(OrganizationManagedFilter):
--- a/openwisp_monitoring/device/api/views.py
+++ b/openwisp_monitoring/device/api/views.py
@@ -39,6 +39,7 @@ from ..schema import schema
 from ..signals import device_metrics_received
 from ..tasks import write_device_metrics
 from .filters import (
+    DeviceFilter,
     MonitoringDeviceFilter,
     MonitoringNearbyDeviceFilter,
     WifiSessionFilter,
@@ -246,6 +247,8 @@ monitoring_geojson_location_list = MonitoringGeoJsonLocationList.as_view()
 
 class MonitoringLocationDeviceList(LocationDeviceList):
     serializer_class = MonitoringLocationDeviceSerializer
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = DeviceFilter
 
     def get_queryset(self):
         return super().get_queryset().select_related("monitoring").order_by("name")
--- a/openwisp_monitoring/device/static/monitoring/css/device-map.css
+++ b/openwisp_monitoring/device/static/monitoring/css/device-map.css
@@ -17,6 +17,87 @@
   font-weight: bold;
   text-transform: uppercase;
 }
+.paginator {
+  padding-bottom: 0;
+}
+.leaflet-popup-content {
+  display: flex;
+  min-width: fit-content;
+  padding: 11px 0 11px 4px;
+}
+.map-detail {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  box-sizing: border-box;
+}
+.map-detail h2 {
+  margin: 10px 0;
+  font-size: 20px;
+}
+.input-container {
+  margin: 10px 0;
+  position: relative;
+  display: inline-block;
+  width: 100%;
+  display: flex;
+}
+.input-container::before {
+  content: "";
+  position: absolute;
+  top: 46%;
+  left: 0.6rem;
+  transform: translateY(-54%);
+  width: 1.1rem;
+  height: 1.1rem;
+  background-color: #717680;
+  mask-image: url("Icon.svg");
+  -webkit-mask-image: url("Icon.svg");
+}
+.map-detail input {
+  outline: none;
+  width: 100%;
+  padding: 8px 12px;
+  padding-left: 2rem;
+  box-sizing: border-box;
+  border-radius: 6px;
+  font-size: 18px;
+}
+.label-container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  margin: 10px 0;
+  gap: 10px;
+}
+.status-filter {
+  cursor: pointer;
+}
+.health-status {
+  border: 1px solid;
+  padding: 3px 10px;
+  border-radius: 12px;
+  white-space: nowrap;
+  font-size: 13px;
+}
+.table-container {
+  max-height: 120px;
+  width: 400px;
+  overflow-y: auto;
+  min-width: 300px;
+  box-sizing: border-box;
+  margin: 10px 0;
+}
+.table-container table {
+  width: 100%;
+}
+.no-devices {
+  text-align: center;
+  font-weight: bold;
+}
+.remove-icon {
+  font-weight: bold;
+}
 .map-detail .paginator a {
   padding: 3px 7px;
   margin: 0 3px;
@@ -29,10 +110,19 @@
   background-color: #000000;
   color: #fff;
 }
+.floorplan-btn {
+  padding: 12px 25px;
+  margin: 10px 0;
+  width: fit-content;
+  font-size: 18px;
+  border-radius: 8px;
+  align-self: center;
+}
 #device-map-container {
   margin: 50px auto;
   position: relative;
-  height: 300px;
+  /* Todo: Remove this added only for testing */
+  height: 85vh;
   width: 90%;
   border: 1px solid rgba(0, 0, 0, 0.3);
 }
--- a/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
+++ b/openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
@@ -9,9 +9,11 @@
   border: 1px solid #ccc;
   color: #6d6357;
   line-height: 20px;
-  max-width: 400px;
+  /* max-width: 400px;
   min-width: 200px;
-  padding: 10px 15px;
+  padding: 10px 15px; */
+  width: 100%;
+  padding: 0;
 }
 #graphChartContainer {
   width: 100%;
--- a/openwisp_monitoring/device/static/monitoring/js/device-map.js
+++ b/openwisp_monitoring/device/static/monitoring/js/device-map.js
@@ -62,60 +62,191 @@
         withCredentials: true,
       },
       success: function (data) {
-        let html = "",
-          device;
-        for (let i = 0; i < data.results.length; i++) {
-          device = data.results[i];
-          html += `
-                            <tr>
-                                <td><a href="${device.admin_edit_url}">${device.name}</a></td>
-                                <td>
-                                    <span class="health-status health-${device.monitoring.status}">
-                                        ${device.monitoring.status_label}
-                                    </span>
-                                </td>
-                            </tr>`;
-        }
-        let pagination = "",
-          parts = [];
-        if (data.previous || data.next) {
-          if (data.previous) {
-            parts.push(
-              `<a class="prev" href="#prev" data-url="${data.previous}">&#8249; ${gettext("previous")}</a>`,
-            );
-          }
-          if (data.next) {
-            parts.push(
-              `<a class="next" href="#next" data-url="${data.next}">${gettext("next")} &#8250;</a>`,
-            );
-          }
-          pagination = `<p class="paginator">${parts.join(" ")}</div>`;
-        }
+        let devices = data.results;
+        const uniqueStatus = Array.from(
+          new Set(devices.map((d) => d.monitoring.status)),
+        );
+        let statusFilter = "";
+        uniqueStatus.forEach((status) => {
+          const label = gettext(status);
+          statusFilter += `
+             <span 
+                class="health-status health-${status} status-filter" 
+                data-status="${status}"
+              >
+                ${label}
+              </span>
+            `;
+        });
+        const has_floorplan = data.has_floorplan;
+        const floorplan_btn = has_floorplan
+          ? `<button class="default-btn floorplan-btn">
+          <span class="ow-floor icon"></span>  Switch to Floor Plan
+        </button>`
+          : "";
         layer.bindPopup(`
-                            <div class="map-detail">
-                                <h2>${layer.feature.properties.name} (${data.count})</h2>
-                                <table>
-                                    <thead>
-                                        <tr>
-                                            <th>${gettext("name")}</th>
-                                            <th>${gettext("status")}</th>
-                                        </tr>
-                                    </thead>
-                                    <tbody>
-                                        ${html}
-                                    </tbody>
-                                </table>
-                                ${pagination}
-                            </div>`);
+                          <div class="map-detail">
+                            <h2>${layer.feature.properties.name} (${data.count})</h2>
+                            <div class="input-container">
+                              <input id="device-search" placeholder="Search for devices" />
+                            </div>
+                            <div class="label-container">
+                              ${statusFilter}
+                            </div>
+                            <div class="table-container">
+                              <table>
+                                <thead>
+                                    <tr>
+                                        <th>${gettext("name")}</th>
+                                        <th><span class ="health-status-heading">${gettext(
+                                          "status",
+                                        )}</span></th>
+                                    </tr>
+                                </thead>
+                                <tbody>
+                                    
+                                </tbody>
+                              </table>
+                              <div class="ow-loading-spinner" style="display: none; position: absolute; bottom: 10px; left: 50%; transform: translateX(-50%);"></div>
+                            </div>
+                            ${floorplan_btn}
+                          </div>`);
         layer.openPopup();
 
-        // bind next/prev buttons
         let el = $(layer.getPopup().getElement());
-        el.find(".next").click(function () {
-          loadPopUpContent(layer, $(this).data("url"));
+        function renderRows(devices) {
+          if (devices.length === 0) {
+            el.find("tbody").html(`
+              <tr>
+                <td class="no-devices">
+                  ${gettext("No devices found!")}
+                </td>
+              </tr>
+            `);
+            return;
+          }
+          const rows = devices
+            .map(
+              (device) => `
+            <tr>
+                <td><a href="${device.admin_edit_url}">${device.name}</a></td>
+                <td>
+                    <span class="health-status health-${device.monitoring.status}">
+                        ${device.monitoring.status_label}
+                    </span>
+                </td>
+            </tr>
+          `,
+            )
+            .join("");
+          el.find("tbody").html(rows);
+        }
+        renderRows(devices);
+        el.find("#device-search").on("input", function (e) {
+          const q = e.target.value.toLowerCase().trim();
+          if (!q) {
+            renderRows(devices);
+          } else {
+            // const filtered = devices.filter((d) => {
+            //   return (
+            //     d.name.toLowerCase().includes(q) ||
+            //     d.mac_address.toLowerCase().includes(q)
+            //   );
+            // });
+            // renderRows(filtered);
+            $.ajax({
+              dataType: "json",
+              url: url + "&search=" + q,
+              xhrFields: { withCredentials: true },
+
+              success(data) {
+                devices = data.results;
+                nextUrl = data.next;
+                renderRows(devices);
+              },
+              error() {
+                console.error("Could not load more devices from", url);
+                nextUrl = url;
+              },
+            });
+          }
         });
-        el.find(".prev").click(function () {
-          loadPopUpContent(layer, $(this).data("url"));
+        let activeStatuses = [];
+        el.find(".status-filter").on("click", function (e) {
+          e.stopPropagation();
+          const $badge = $(this);
+          const status = $badge.data("status");
+          const label = gettext(status);
+
+          if ($badge.hasClass("active")) {
+            $badge.removeClass("active").html(label);
+            activeStatuses = activeStatuses.filter((s) => s !== status);
+          } else {
+            $badge
+              .addClass("active")
+              .html(`${label} <span class="remove-icon">&times;</span>`);
+            activeStatuses.push(status);
+          }
+
+          let toShow;
+          if (activeStatuses.length === 0) {
+            toShow = devices;
+          } else {
+            $.ajax({
+              dataType: "json",
+              url: url + "&status=" + label,
+              xhrFields: { withCredentials: true },
+
+              success(data) {
+                devices = data.results;
+                nextUrl = data.next;
+                renderRows(devices);
+              },
+              error() {
+                console.error("Could not load more devices from", url);
+                nextUrl = url;
+              },
+            });
+          }
+        });
+        let nextUrl = data.next;
+        let isLoading = false;
+
+        function loadMoreDevices() {
+          if (!nextUrl || isLoading) return;
+
+          isLoading = true;
+
+          const $spinner = el.find(".table-container .ow-loading-spinner");
+          $spinner.show();
+
+          const url = nextUrl;
+          nextUrl = null;
+
+          $.ajax({
+            dataType: "json",
+            url,
+            xhrFields: { withCredentials: true },
+
+            success(newData) {
+              devices = devices.concat(newData.results);
+              nextUrl = newData.next;
+              renderRows(devices);
+            },
+            error() {
+              console.error("Could not load more devices from", url);
+              nextUrl = url;
+            },
+            complete() {
+              isLoading = false;
+              $spinner.hide();
+            },
+          });
+        }
+        el.find(".table-container").on("scroll", function () {
+          if (this.scrollTop + this.clientHeight >= this.scrollHeight - 10) {
+            loadMoreDevices(nextUrl);
+          }
         });
 
         loadingOverlay.hide();
--- a/openwisp_monitoring/device/static/monitoring/css/Icon.svg
+++ b/openwisp_monitoring/device/static/monitoring/css/Icon.svg
@@ -1,0 +1,3 @@
+<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M16.5 16.5L12.875 12.875M14.8333 8.16667C14.8333 11.8486 11.8486 14.8333 8.16667 14.8333C4.48477 14.8333 1.5 11.8486 1.5 8.16667C1.5 4.48477 4.48477 1.5 8.16667 1.5C11.8486 1.5 14.8333 4.48477 14.8333 8.16667Z" stroke="#717680" stroke-width="1.66667" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>


openwisp_monitoring/device/api/filters.py
```diff
@@ -23,6 +25,25 @@ class Meta:
         }
 
 
+class DeviceFilter(filters.FilterSet):
+    search = filters.CharFilter(method="filter_search")
+
```
pandafy:
```suggestion
```

openwisp_monitoring/device/api/views.py
```diff
@@ -246,6 +247,8 @@ class MonitoringGeoJsonLocationList(GeoJsonLocationList):
 
 class MonitoringLocationDeviceList(LocationDeviceList):
     serializer_class = MonitoringLocationDeviceSerializer
+    filter_backends = [DjangoFilterBackend]
```
pandafy, line 253: **Thinking out loud**

Is this the right place to add filters? Shall we move the filters to openwisp-controller? 

Let's discuss this with @nemesifier!

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -1,22 +1,116 @@
...
+}
+.health-status {
+  padding: 3px 10px;
+  border-radius: 12px;
+  white-space: nowrap;
+  font-size: 12px;
+}
+.table-container {
+  height: 120px;
+  width: 400px;
```
pandafy: In the figma design, the width of the whole popup is of 400px. You have set this here. Thus will margin and padding, it becomes 450px. 

However, the popup does not look good with 400px width and less height. Let's discuss this in today's call.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +61,177 @@
...
+          const label = gettext(status_label);
+          statusFiltersBtn += `
+            <span 
+              class="health-status health-${status} status-filter" 
+              data-status="${status}"
+            >
+              ${label}
+            </span>
+          `;
+        });
```
pandafy: What if a deactivated object exists on the last page? In this way, we won't show a filter for the "deactivated" devices?

openwisp_monitoring/device/static/monitoring/css/device-map.css
pandafy, line 1: The status column has center text alignment in the figma design, while the UI has left alignment.

pandafy, line 1: The padding of the leaflet pop-up is not according to the figma desgin.

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,110 @@
...
+}
+.floor-icon {
+  display: inline-block;
+  width: 25px;
+  height: 25px;
+  background-color: white;
+  mask-repeat: no-repeat;
+  margin-right: 16px;
+}
+.floorplan-btn {
```
pandafy, line 195: In the figma desgin, we have more spacing on the top of the button.

openwisp_monitoring/device/api/filters.py
dee077: Updated!

openwisp_monitoring/device/api/views.py
dee077: If we add this filter to the controller, it would only apply to the search filter. To add filter by status in the monitoring view, we’d need to override it here in monitoring so we end up having one filter in controller and one in monitoring.

**Let me know if it is required to move it to the controller.**

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated the logic
Initialized a variable with the default status keys.
```
const statusLabelsMap = {
          ok: "ok",
          problem: "problem",
          critical: "critical",
          unknown: "unknown",
          deactivated: "deactivated",
        };
```
Then updated the keys of the object to the status labels present in the data fetched from the endpoint.
Then create the status <span> with the updated label and text inside it

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated the logic
Send the health status labels from the app settings to the window object uses it from there

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,273 @@
...
+        return;
+      }
+      $.ajax({
+        url: reqUrl,
+        method: "GET",
+        dataType: "json",
+        xhrFields: { withCredentials: true },
+        success: async (data) => {
+          // To make this run only one time as only in the first call floor will not be provided
+          // And sort them in decresing order so that neagtive floor show at bottom and postove at top in floor navigation
```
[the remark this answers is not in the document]
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -1,44 +1,195 @@
...
+}
+.floor-icon {
+  display: inline-block;
+  width: 25px;
+  height: 25px;
+  background-color: white;
+  mask-repeat: no-repeat;
+  margin-right: 16px;
 }
 #device-map-container {
```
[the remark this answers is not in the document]
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -1,44 +1,195 @@
...
+  justify-content: center;
+}
+.floor-icon {
+  display: inline-block;
+  width: 25px;
+  height: 25px;
+  background-color: white;
+  mask-repeat: no-repeat;
+  margin-right: 16px;
 }
```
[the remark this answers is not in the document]
dee077: Updated!

openwisp_monitoring/device/api/views.py
```diff
@@ -254,6 +260,33 @@ def get_queryset(self):
 monitoring_location_device_list = MonitoringLocationDeviceList.as_view()
 
 
+# Todo: Remove this
+class TestPagination(ListViewPagination):
+    page_size = 2
+
+
+class MonitoringIndoorCoordinatesList(IndoorCoordinatesList):
+    pagination_class = TestPagination
```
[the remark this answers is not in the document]
dee077: Removed!

openwisp_monitoring/device/api/serializers.py
```diff
@@ -46,6 +48,15 @@ class MonitoringLocationDeviceSerializer(LocationDeviceSerializer):
     monitoring = DeviceMonitoringLocationSerializer()
 
 
+class MonitoringIndoorCoordinatesSerializer(IndoorCoordinatesSerializer):
+    monitoring = monitoring = DeviceMonitoringLocationSerializer(
```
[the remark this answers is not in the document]
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,273 @@
...
+          if (!currentFloor && data.results.length) {
+            currentFloor = data.results[0].floor;
+          }
+          if (data.next) {
+            await fetchData(data.next);
+          }
+          resolve();
+        },
+        error: () => {
+          alert("Error loading floorplan coordinates.");
```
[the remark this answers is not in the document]
dee077: I don't think there is a function named `showErrorNotification`, there are multiple places where alert is used, should use toast form openwisp notification instead?

openwisp_monitoring/device/api/views.py
pandafy: Okay, let's keep it here.

openwisp_monitoring/device/apps.py
```diff
@@ -376,15 +377,33 @@ def register_dashboard_items(self):
...
+            )
+            labels = json.dumps(
+                {
+                    "ok": app_settings.HEALTH_STATUS_LABELS["ok"],
+                    "problem": app_settings.HEALTH_STATUS_LABELS["problem"],
+                    "critical": app_settings.HEALTH_STATUS_LABELS["critical"],
+                    "unknown": app_settings.HEALTH_STATUS_LABELS["unknown"],
+                    "deactivated": app_settings.HEALTH_STATUS_LABELS["deactivated"],
+                }
+            )
```
pandafy: Is it necessary to do a `json.dump` here ?

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -1,38 +1,37 @@
...
+  border: 1px solid #eaecf0;
 }
 .health-ok {
-  color: green;
+  color: #067647;
+  background-color: #dcfae6;
+  border: 1px solid #75e0a7;
 }
 .health-problem {
   color: orange;
```
pandafy:
```suggestion
    color: #B54708;
```

I think you missed updating this color.

openwisp_monitoring/device/apps.py
pandafy: I tried these changes 

```diff 
diff --git a/openwisp_monitoring/device/apps.py b/openwisp_monitoring/device/apps.py
index d2347bf..53f7b78 100644
--- a/openwisp_monitoring/device/apps.py
+++ b/openwisp_monitoring/device/apps.py
@@ -1,4 +1,3 @@
-import json
 from urllib.parse import urljoin
 
 from django.apps import AppConfig
@@ -382,15 +381,6 @@ class DeviceMonitoringConfig(AppConfig):
                 urlconf=MONITORING_API_URLCONF,
                 args=["000"],
             )
-            labels = json.dumps(
-                {
-                    "ok": app_settings.HEALTH_STATUS_LABELS["ok"],
-                    "problem": app_settings.HEALTH_STATUS_LABELS["problem"],
-                    "critical": app_settings.HEALTH_STATUS_LABELS["critical"],
-                    "unknown": app_settings.HEALTH_STATUS_LABELS["unknown"],
-                    "deactivated": app_settings.HEALTH_STATUS_LABELS["deactivated"],
-                }
-            )
             if MONITORING_API_BASEURL:
                 device_list_url = urljoin(MONITORING_API_BASEURL, str(device_list_url))
                 loc_geojson_url = urljoin(MONITORING_API_BASEURL, str(loc_geojson_url))
@@ -420,7 +410,7 @@ class DeviceMonitoringConfig(AppConfig):
                     "monitoring_device_list_url": device_list_url,
                     "monitoring_location_geojson_url": loc_geojson_url,
                     "monitoring_indoor_coordinates_list": indoor_coordinates_list_url,
-                    "monitoring_labels": labels,
+                    "monitoring_labels": app_settings.HEALTH_STATUS_LABELS,
                 },
             )
 
 ```
 
It generated the following HTML 

```
<script type="text/javascript">
  window._owGeoMapConfig = {
    geoJsonUrl: '/api/v1/monitoring/geojson/',
    locationDeviceUrl: '/api/v1/monitoring/location/000/device/',
    indoorCoordinatesUrl: '/api/v1/monitoring/location/000/indoor-coordinates/',
    labels: {'unknown': 'unknown', 'ok': 'ok', 'problem': 'problem', 'critical': 'critical', 'deactivated': 'deactivated'}
  }
</script>
```

Which is a valid JSON object.

File: openwisp_monitoring/device/static/monitoring/css/netjsongraph.css [hunk 14]
```diff
@@ -9,9 +9,11 @@
   border: 1px solid #ccc;
   color: #6d6357;
   line-height: 20px;
-  max-width: 400px;
+  /* max-width: 400px;
   min-width: 200px;
-  padding: 10px 15px;
+  padding: 10px 15px; */
```
pandafy: Is this a leftover?

openwisp_monitoring/settings.py
```diff
@@ -29,3 +29,10 @@ def get_settings_value(option, default=None):
     24 * 60 * 60,  # 24 hours in seconds
 )
 DEFAULT_CHART_TIME = get_settings_value("DEFAULT_CHART_TIME", "7d")
+HEALTH_STATUS_LABELS = {
+    "ok": "Ok",
+    "problem": "Problem",
+    "critical": "Critical",
+    "unknown": "Unknown",
+    "deactivated": "Deactivated",
+}
```
pandafy: Why do we require this? We already have a setting for this which we discussed on our 1-on-1.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
-              `<a class="next" href="#next" data-url="${data.next}">${gettext("next")} &#8250;</a>`,
-            );
-          }
-          pagination = `<p class="paginator">${parts.join(" ")}</div>`;
-        }
+        let devices = data.results;
+        let nextUrl = data.next;
+        const statusLabelsMap = labels;
+        let statusFilterButtons = "";
+        console.log(statusLabelsMap);
```
pandafy: 👀

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+          ([status_key, status_label]) => {
+            const label = gettext(status_label);
+            statusFilterButtons += `<span 
+              class="health-status health-${status_key} status-filter" 
+              data-status="${status_key}"
+            >
+              ${label}
+            </span>`;
+          },
+        );
```
pandafy: This is good!

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+                          <div class="map-detail">
+                            <h2>${layer.feature.properties.name} (${
+                              data.count
+                            })</h2>
+                            <div class="input-container">
+                              <input id="device-search" placeholder="Search for devices" />
+                            </div>
+                            <div class="label-container">
+                              ${statusFilterButtons}
+                              <input id="status-filter" style="display: none;" type="text" />
```
pandafy:
```suggestion
                              <input id="status-filter" type="hidden" />
```

pandafy: More on hidden input field https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/hidden

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+                                    <tr>
+                                        <th>${gettext("name")}</th>
+                                        <th><span class ="health-status-heading">${gettext(
+                                          "status",
+                                        )}</span></th>
+                                    </tr>
+                                </thead>
+                                <tbody>
+                                    
+                                </tbody>
```
pandafy:
```suggestion
                                <tbody></tbody>
```

In favour of brevity, I would write it like this

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+          el.find("tbody").html(rows);
+        }
+        let fetchDevicesTimeout;
+        let loading = false;
+        function fetchDevices(url, ms = 0) {
+          if (!url || loading) return;
+          clearTimeout(fetchDevicesTimeout);
+          loading = true;
+          const spinner = el.find(".table-spinner");
+          spinner.show();
```
pandafy: The spinner is rendered over the existing rows. Is this expected? 


https://github.com/user-attachments/assets/80ca7852-09b9-4c6e-9414-7e4f8358bde7

I was expecting the code to hide the table when loading the data. This is what we do everywhere in OpenWISP.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+            const statusParam = el.find("#status-filter").val();
+            if (searchParam) {
+              params.append("search", searchParam);
+            }
+
+            if (statusParam) {
+              statusParam.split(",").forEach((status) => {
+                params.append("status", status);
+              });
+            }
```
pandafy: This is good!

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+            const queryString = params.toString();
+            let fetchUrl;
+            // if nextUrl is the same as url, that means we are fetching for infinite scroll
+            if (url === nextUrl) fetchUrl = url;
+            else fetchUrl = queryString ? `${url}?${queryString}` : url;
```
pandafy: Can you please explain why do we need to check if the request is for infinite scroll or not? 

And, **always use braces for conditionals!!!**

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+
+            $.ajax({
+              dataType: "json",
+              url: fetchUrl,
+              xhrFields: { withCredentials: true },
+
```
pandafy:
```suggestion
```

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+                  devices = devices.concat(data.results);
+                } else {
+                  devices = data.results;
+                }
+                nextUrl = data.next;
+                renderRows(devices);
+              },
+              error() {
+                console.error("Could not load more devices from", url);
+              },
```
pandafy: This is nice. 

We should also inform the user about this error. You can use alert for now.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
 
+          if (btn.hasClass("active")) {
+            btn.removeClass("active").html(label);
+            activeStatuses = activeStatuses.filter((s) => s !== status);
+          } else {
+            btn
+              .addClass("active")
+              .html(`${label} <span class="remove-icon">&times;</span>`);
+            activeStatuses.push(status);
+          }
```
pandafy: With a little bit of CSS we can get rid of adding this close button on the fly each time the filter is added or removed. 

```diff
$ git diff openwisp_monitoring/device/static/monitoring/css/device-map.css openwisp_monitoring/device/static/monitoring/js/device-map.js
diff --git a/openwisp_monitoring/device/static/monitoring/css/device-map.css b/openwisp_monitoring/device/static/monitoring/css/device-map.css
index 164b096..381ca5e 100644
--- a/openwisp_monitoring/device/static/monitoring/css/device-map.css
+++ b/openwisp_monitoring/device/static/monitoring/css/device-map.css
@@ -130,6 +130,10 @@
 }
 .remove-icon {
   font-weight: bold;
+  display: none;
+}
+.status-filter.active .remove-icon{
+  display: inline-block;
 }
 .table-container {
   height: 120px;
diff --git a/openwisp_monitoring/device/static/monitoring/js/device-map.js b/openwisp_monitoring/device/static/monitoring/js/device-map.js
index ddb15b1..ce58550 100644
--- a/openwisp_monitoring/device/static/monitoring/js/device-map.js
+++ b/openwisp_monitoring/device/static/monitoring/js/device-map.js
@@ -74,11 +74,12 @@
         Object.entries(statusLabelsMap).forEach(
           ([status_key, status_label]) => {
             const label = gettext(status_label);
-            statusFilterButtons += `<span 
-              class="health-status health-${status_key} status-filter" 
+            statusFilterButtons += `<span
+              class="health-status health-${status_key} status-filter"
               data-status="${status_key}"
             >
               ${label}
+              <span class="remove-icon">&times</span>
             </span>`;
           },
         );
@@ -111,7 +112,7 @@
                                     </tr>
                                 </thead>
                                 <tbody>
-                                    
+
                                 </tbody>
                                 </table>
                                 <div class="ow-loading-spinner table-spinner"></div>
@@ -214,12 +215,9 @@
           const label = gettext(status);
 
           if (btn.hasClass("active")) {
-            btn.removeClass("active").html(label);
-            activeStatuses = activeStatuses.filter((s) => s !== status);
+            btn.removeClass("active");
           } else {
-            btn
-              .addClass("active")
-              .html(`${label} <span class="remove-icon">&times;</span>`);
+            btn.addClass("active");
             activeStatuses.push(status);
           }
           $(`#status-filter`).val(activeStatuses.join(","));
```

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -62,62 +66,174 @@
...
+          $(`#status-filter`).val(activeStatuses.join(","));
+          fetchDevices(url);
+        });
+        el.find(".table-container").on("scroll", function () {
+          if (this.scrollTop + this.clientHeight >= this.scrollHeight - 10) {
+            fetchDevices(nextUrl, 100);
+          }
+        });
```
pandafy, line 257: Any specific reason to add a delay while scrolling?

openwisp_monitoring/device/static/monitoring/js/floorplan.js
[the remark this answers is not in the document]
pandafy: As always, AI is hallucinating.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -21,16 +23,24 @@
...
 Chart = load_model("monitoring", "Chart")
 Check = load_model("check", "Check")
+Location = load_model("geo", "Location")
+DeviceLocation = load_model("geo", "DeviceLocation")
+Floorplan = load_model("geo", "Floorplan")
 
 
 class SeleniumTestMixin(BaseSeleniumTestMixin):
+    # Todo: Added for testing
+    retry_max = 1
```
pandafy: Thanks for the comment.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+            health_ok = self.find_element(By.CSS_SELECTOR, ".map-detail .health-ok")
+            health_ok.click()
+            self.wait_for_invisibility(
+                By.CSS_SELECTOR, ".map-detail .ow-loading-spinner"
+            )
+            sleep(0.5)
+            table_entries = self.find_elements(By.CSS_SELECTOR, ".map-detail tbody tr")
+            self.assertEqual(len(table_entries), 1)
+            self.assertIn(
+                "Test-Device2",
```
pandafy:
```suggestion
               device2.name,
```

Again, please use descriptive variable names.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+            "g path.leaflet-interactive",
+            timeout=5,
+        )
+        location_point = self.find_element(By.CSS_SELECTOR, "g path")
+        location_point.click()
+        self.wait_for_visibility(By.CSS_SELECTOR, ".map-detail")
+        table_entries = self.find_elements(By.CSS_SELECTOR, ".map-detail tbody tr")
+        self.assertEqual(len(table_entries), 2)
+
+        with self.subTest("Test filtering by status"):
```
pandafy: I'd also verify the following things: 

- Multiple filters can be used at the same time 
- The close button appears next to the filter 
- Removing all the filters will hide the close button from all filter and show all the devices.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+        with self.subTest("Test filtering by input field"):
+            input_field = self.find_element(By.CSS_SELECTOR, "#device-search")
+            input_field.send_keys("Test-Device1")
+            self.wait_for_invisibility(
+                By.CSS_SELECTOR, ".map-detail .ow-loading-spinner"
+            )
+            sleep(0.5)
+            table_entries = self.find_elements(By.CSS_SELECTOR, ".map-detail tbody tr")
+            self.assertEqual(len(table_entries), 1)
+            self.assertIn("Test-Device1", table_entries[0].text)
```
pandafy:
```suggestion
            self.assertIn(device1.name, table_entries[0].text)
```

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+            )
+            health_ok.click()
+            self.wait_for_invisibility(
+                By.CSS_SELECTOR, ".map-detail .ow-loading-spinner"
+            )
+            sleep(0.5)
+            table_entries = self.find_elements(By.CSS_SELECTOR, ".map-detail tbody tr")
+            self.assertEqual(len(table_entries), 2)
+
+        with self.subTest("Test filtering by input field"):
```
pandafy: I'd also test that clearing the input field will show all the devices.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+            input_field = self.find_element(By.CSS_SELECTOR, "#device-search")
+            input_field.send_keys("Test-Device1")
```
pandafy: Instead of using the complete name, let's only use "device1" for searching (all in lowercase)

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+            self.wait_for_invisibility(
+                By.CSS_SELECTOR, ".map-detail .ow-loading-spinner"
+            )
+            sleep(0.5)
+            table_entries = self.find_elements(By.CSS_SELECTOR, ".map-detail tbody tr")
+            self.assertEqual(len(table_entries), 1)
+            self.assertIn("Test-Device1", table_entries[0].text)
+
+        with self.subTest("Test filtering to get no results"):
```
pandafy: Can we also add a test for infinite scroll?

Django allows loading [fixtures](https://docs.djangoproject.com/en/5.2/topics/testing/tools/#fixture-loading) for tests. We can use it to populate the database with many devices. 

Or if you have a better solution in mind, you can use that too.

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -284,3 +304,165 @@ def test_device_wifi_session_inline_change(self):
...
+        self._create_object_location(
+            content_object=device2,
+            location=location,
+            floorplan=floor2,
+            organization=org,
+        )
+        self.login()
+        self.wait_for_visibility(By.CSS_SELECTOR, "#device-map-container")
+
+        with self.subTest("Test floorplan redering"):
```
pandafy: Add a test for "floorplan button shouldn't be shown if the location does not have any floors"

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
   font-weight: bold;
   text-decoration: none !important;
 }
+.leaflet-popup-content {
+  display: flex;
+  min-width: fit-content;
+  margin: 24px !important;
```
pandafy: Let's use padding instead of margin. 

Do we neeed to set display and min-width properties here?

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
...
+  box-sizing: border-box;
+}
+.map-detail h2 {
+  margin: 35px 0 10px 0;
+  font-size: 18px;
+}
+.input-container {
+  margin: 10px 0;
+  position: relative;
+  display: inline-block;
```
pandafy: This is repeated.

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
...
+}
+.map-detail input {
+  outline: none;
+  width: 100%;
+  padding: 10px 12px;
+  padding-left: 2rem;
+  box-sizing: border-box;
+  border-radius: 6px;
+  font-size: 16px;
+}
```
pandafy: Please double check the padding and the border color of the input field. The padding looks off from the figma on the input field.

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
...
+.label-container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  margin: 10px 0;
+  gap: 10px;
```
pandafy:
```suggestion
  gap: 8px;
```

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
...
+}
+.health-status {
+  padding: 3px 10px;
+  border-radius: 12px;
+  white-space: nowrap;
+  font-size: 12px;
+}
+.remove-icon {
+  font-weight: bold;
+}
```
pandafy: Screenshot from 2025-07-28 21-44-50

The render icon renders differently from the figma design 

image

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -64,3 +63,119 @@
...
+.map-detail h2 {
+  margin: 35px 0 10px 0;
+  font-size: 18px;
+}
+.input-container {
+  margin: 10px 0;
+  position: relative;
+  display: inline-block;
+  width: 100%;
+  display: flex;
```
pandafy: Do we really need flexbox here?

openwisp_monitoring/device/static/monitoring/css/device-map.css
pandafy: The spacing between the status filters and the table and between the table and the floorplan button are also different.

openwisp_monitoring/device/apps.py
dee077: If the order of showing labels does not matter, then we can do this
or we can change the order in the `settings.py` itself

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/netjsongraph.css
dee077: This was the fix made to show the full width of the map container 
Updated now!

openwisp_monitoring/settings.py
dee077: This way we can check the custom label as well
let me know if it is not required

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Used braces for conditionals
If we are fetching for infinte scroll new need concat the results otherwise not
Added comment as well

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Thought good to have some delay 
Let me know if it is not needed

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Yes we dont need flex here
Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated the spacing between  status filters and the table
The same close icon is used multiple locations here can we download it from figma and include it in `openwisp-utils`
Let me know your thoughts on this

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Updated the padding

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: There is a defualt `margin: 13px 24px 13px 20px;` in `.leaflet-popup-content`
If we add padding we still need to overwrite the above margin to 0 and add padding to the parent div

cestercian:
https://github.com/openwisp/netjsongraph.js/pull/407

pandafy (commented):
@nemesifier @dee077 and I discussed the following changes to the UI today 

- [x] Device popup heading margin should be set to 0 
- [x] Heading and close button should be at the same level
- [x] Device table height is 180px (max height)
    - [x] If there's only one device, than it shall adapt accordingly 
    - [x] If there are more than 3 devices, then show a scroll bar. Otherwise, just show the three devices. 
- [x] Make the filters highlighted when they are selected. 
    - [x]  The default opacity should be 0.8
    - [x]  After clicking, the filter the opacity should be 1 (`.health-status.active {opacity: 1}`)
- [x] Reduce the opacity of the disabled navigation arrow to 0.3 
    - [x] Set `{cursor: default}` on the disabled navigation arrow 
- [x] Set `{cursor: default}` to active floor button `floor-btn.active.selected`
- [x] Change the color of active arrow to #737985
- [ ] Changes to the floorplan overlay 
    - [x] Remove padding on floorplan heading 
    - [ ] Add `margin-bottom: 10px` to the heading 
    - [ ] Update the position of the close button
      ```css
        #floorplan-heading {
            margin-bottom: 10px; 
            // remove padding 
        }
        #floorplan-close-btn {
            top: 8px; 
            right: 8px;
        }
      ```
- [x] Add hover effect on the close icons (set `color: black` on hover)
- [ ] Floorplan navigation widget - if there are more floors available, and the selected floor is in the middle, then clicking on the navigation button should add new floors to the navigation widget instead of moving the selected floor. 
- [ ] Increase the margin of the floorplan navigation arrow (@dee077 I missed to write down the value of the margin, I guess it was 8px)

cestercian:
https://github.com/openwisp/netjsongraph.js/pull/407

openwisp_monitoring/device/static/monitoring/css/device-map.css
```diff
@@ -73,14 +73,17 @@
   right: 25px !important;
   font-size: 20px !important;
 }
+.leaflet-popup-close-button:hover {
```
pandafy, line 77:
```suggestion 
.leaflet-popup-close-button:hover, 
.leaflet-popup-close-button:active, 
.leaflet-popup-close-button:focus {
```

For completeness, let's also include active and focus pseudo class.

pandafy:
In our debugging session today, we found that the fullscreen can only be toggled by user gesture (not by JS from an async function). 

### Problem: Fullscreen mode can only be triggered by user gesture

Possible solution for the fullscreen navigation: 
  
1. Don't show the floorplan navigation in the fullscreen mode 

   (Quick fix until we find a way to solve it properly)
2. Show the floorplan navigation in fullscreen, but clicking on the floor will exit the fullscreen 
    and then the floor will be changed. The user will see the changed floor in non fullscreen window 
  
   (Almost implemented, con: clunky experience for the user)
3. (Uncertain) **Redo the implementation**, use only one NetJSONGraph instance to render the 
  data. When the user clicks on a floor, the background will be changed and the data will be 
  rendered again. 

   (I think this may not worth the effort right now. We can come back to this later if required)

nemesifier:
I think we need to go with 2. Not allowing to navigate floors is not acceptable. In the meantime we need to investigate option 3.

Please @pandafy @dee077 think on whether we could add logic in netjsongraph.js to facilitate option 3 and if it's technically feasible we can create a follow up issue to work on it in the coming months.

openwisp_monitoring/device/apps.py
dee077: Removed!

openwisp_monitoring/settings.py
dee077: Updated to custom label names.

openwisp_monitoring/tests/test_selenium.py
dee077: Updated!

openwisp_monitoring/tests/test_selenium.py
dee077: Added these scenarios

openwisp_monitoring/tests/test_selenium.py
dee077: Updated!

openwisp_monitoring/tests/test_selenium.py
dee077: Added the test

openwisp_monitoring/tests/test_selenium.py
dee077: Updated!

openwisp_monitoring/tests/test_selenium.py
dee077: Added using simple for loop

openwisp_monitoring/tests/test_selenium.py
dee077: We are not showing the button at all when there is no floor plan, so how can we search for an element that does not exist in the DOM? It was causing issues. 
Let me know if a solution exists.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,410 @@
...
+        <span class="njg-tooltip-key">status</span>
+        <span class="popup-status health-${node?.monitoring.status} ">
+          ${node?.monitoring.status_label}
+        </span>
+      </div>
+      <div class="open-device-btn-container">
+        <button class="default-btn open-device-btn" data-url="${node?.admin_edit_url}"">
+          <span class="ow-device floor-icon"></span>
+          Open Device
+        </button>
```
pandafy: Can this be an anchor element? The job of this button is to redirect to the device admin page, right? I think, it will be better to add a hyperlink here. This will also remove the need for adding click event handlers.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,410 @@
...
+    }
+    $floorDiv.show();
+    const floorNavigation = $("#floorplan-navigation");
+    if (isFullScreen && maps[floor]) {
+      document.exitFullscreen();
+      floorNavigation.addClass("fullscreen");
+      $(`#floor-content-${floor} .leaflet-container`).append(floorNavigation);
+    }
+    maps[currentFloor]?.invalidateSize();
+  }
```
pandafy: Bug: Empty floorplan container if the floor is not present in the location 

**Steps to replocate**
1. Open any floor in a location (here floor 8) and close the overlay
2. Choose any other location which does not have the same floor and open the floorplan overlay 

**Expected result**

The floorplan overlay should show the nearest floor to the ground 

**Actual outcome** 

Blank floorplan overlay 



https://github.com/user-attachments/assets/51e5b4ce-b805-4152-8c16-a002ea352c44

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,410 @@
...
+      .off("click", ".open-device-btn")
+      .on("click", ".open-device-btn", function () {
+        const url = $(this).data("url");
+        window.location.href = url;
+      });
+    currentPopup = L.popup({ closeOnClick: false })
+      .setLatLng(node?.properties.location)
+      .setContent(popupContent)
+      .openOn(map);
+  }
```
pandafy: Currently, both pop-up and tool-tip are showing at the same time. 

IIRC, we discussed to hide the tooltip whenever pop is shown. 


https://github.com/user-attachments/assets/e6c8f515-04f5-4a61-a111-c19b2b60f982

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,410 @@
...
+            const nodeProjected = L.point(topLeft.x + px, topLeft.y + py);
+            const nodeLatLng = map.unproject(nodeProjected, zoom);
+            node.properties = {
+              ...node.properties,
+              location: nodeLatLng,
+            };
+          });
+          // Update the map options with new node locations
+          const mapOptions = this.utils.generateMapOption(this.data, this);
+          this.echarts.setOption(mapOptions);
```
pandafy: Can we perform this operation in the prepareData method?

pandafy: I suspect these lines are override the options that you set which initializing netjsongraph. Thus, all nodes on the floorplan have grey color irrespective of their status. 


Screenshot from 2025-08-25 20-28-02

openwisp_monitoring/device/static/monitoring/js/floorplan.js
```diff
@@ -0,0 +1,410 @@
...
+
+          // Unproject the topLeft and bottomRight points to get northWest and southEast latlngs
+          const nw = map.unproject(topLeft, zoom);
+          const se = map.unproject(bottomRight, zoom);
+          const bnds = L.latLngBounds(nw, se);
+          L.imageOverlay(imageUrl, bnds).addTo(map);
+          map.fitBounds(bnds);
+          map.setMaxBounds(bnds);
```
pandafy:
```suggestion
                    map.setMaxBounds(bnds.pad(0.3));
```

I have some devices which are situated the edge of the map. Without the padding for max bound, the pop-up was getting clipped. 

Screenshot from 2025-08-25 20-47-13

Adding padding to the map fixes this issue for me.

pandafy: With padding 
Screenshot from 2025-08-25 20-48-32

openwisp_monitoring/device/static/monitoring/js/floorplan.js
pandafy: Question: Did we discuss to only show the name of the device in the tooltip? 

Am I hallucinating? (XD)

.github/workflows/ci.yml
```diff
@@ -103,13 +103,13 @@ jobs:
...
-          TIMESERIES_UDP=1 coverage run runtests.py --exclude-tag=selenium_tests
-          coverage combine
-          coverage xml
+          ./tests/manage.py test openwisp_monitoring.tests.test_selenium.TestDashboardMap
+          # # the next test runs focus on specific backend features, therefore
+          # # we can skip selenium tests to improve speed and test consistency
+          # SAMPLE_APP=1 coverage run runtests.py --exclude-tag=selenium_tests
+          # TIMESERIES_UDP=1 coverage run runtests.py --exclude-tag=selenium_tests
+          # coverage combine
+          # coverage xml
```
nemesifier: These will have to be restored, let's not forget.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/1.2
+openwisp-controller @ git+https://github.com/dee077/openwisp-controller.git@feature/828-api-indoor-map-coordinates#egg=openwisp-controller
```
nemesifier: Shouldn't this be updated? Didn't we merge that PR?

openwisp_monitoring/device/static/monitoring/js/floorplan.js
nemesifier: Yeah I also noticed the gray color of the circles regardless of their status.

nemesifier (commented):
I did a round of testing, here's my observations.

First of all, I am really happy to see the coordinates issue resolved, thanks!

### Floorplan view device status

I noticed the same thing Gagan brought up here:
https://github.com/openwisp/openwisp-monitoring/pull/688/files#r2298359942

### Map height

I noticed the map height in the dashboard page to be a lot higher than the current master, is that on purpose?

It's taking the entire space on my screen, not leaving room for the main stats to show up.

### Floorplan view hover

I think you're already aware of this right? Is this what you were discussing with @cestercian?

Screenshot from 2025-08-25 16-42-24

### Full screen vertical scroll quirk

This happens when going to full screen on the indoor view, scrolling down with the mouse and then closing the full screen.

indoor-scroll-quirk

I would expect to see the map again, but the page has scrolled down.

## Loading overlay

As discussed, let's add a loading overlay while the floorplan data and view is loading.

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: No, we can't we need an instance of a map to convert the data, which is not available at the time of prepare data.

pandafy: In prepareData, we will only need the dimensions of the image to update the co-ordinates.

pandafy: You can ignore my last comment. On closer inspection, I see that we are using map.project and map.unproject. You were correct, we can't do this in prepareData. 


Can you update the comment to also mention why we perform this operation here and not in prepareData?

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Added!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Added label

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Fixed!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: Fixed!

openwisp_monitoring/device/static/monitoring/js/floorplan.js
dee077: This was because the variable `currentFloor` is not set to null after closing the overlay.

openwisp_monitoring/device/static/monitoring/css/device-map.css
dee077: Added!

requirements.txt
dee077: Updated!

.github/workflows/ci.yml
dee077: Updated!

dee077:
## Updates:
- Updated netjsongraph.min.js fixes the label color issue, but it makes the image squeezed vertically and the nodes are appearing in the wrong place again, so keep updating nodes lat, lng on the fly. 
- Make the popup show 2 table entries at a time.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -3,15 +3,15 @@ name: Bug report
 about: Open a bug report
 title: "[bug] "
 labels: bug
-assignees: ''
-
+assignees: ""
```
nemesifier: This should have already been handled on the target branch, have you updated/rebased on that branch? 

See https://github.com/openwisp/openwisp-monitoring/commit/e53cd57e39e1aa0c7087a865f5808e7a5daa05a8.

openwisp_monitoring/device/api/views.py
```diff
@@ -246,6 +250,8 @@ class MonitoringGeoJsonLocationList(GeoJsonLocationList):
 
 class MonitoringLocationDeviceList(LocationDeviceList):
     serializer_class = MonitoringLocationDeviceSerializer
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = MonitoringLocationDeviceFilter
```
nemesifier, line 254: Are there basic tests for this?

dee077: Added tests for both endpoints
`LocationDeviceList` and `IndoorCoordinatesList`

dee077: Added tests for both endpoints
`LocationDeviceList` and `IndoorCoordinatesList`

.github/ISSUE_TEMPLATE/bug_report.md
dee077: Rebased!

openwisp_monitoring/device/api/urls.py
```diff
@@ -40,4 +40,9 @@
         views.wifi_session_detail,
         name="api_wifi_session_detail",
     ),
+    path(
+        "api/v1/monitoring/location/<str:pk>/indoor-coordinates/",
```
nemesifier, line 44: This needs to be added to the docs.

openwisp_monitoring/device/static/monitoring/css/floorplan.css
```diff
@@ -0,0 +1,188 @@
+.menu-backdrop.active {
```
nemesifier: Can we move the contents of this file into `device-map.css` to avoid one extra HTTP request for a few KB of CSS?

I'll simply separate it with a CSS comment.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -1,23 +1,26 @@
 "use strict";
 
-/* jshint esversion: 8 */
+/*jshint esversion: 8 */
```
nemesifier: I think we can remove this comment.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -116,29 +112,148 @@
...
-        $el.find(".next").click(function (e) {
-          e.preventDefault();
-          loadPopUpContent(nodeData, netjsongraphInstance, $(this).data("url"));
+        const el = $(currentPopup.getElement());
+        function renderRows() {
+          if (devices.length === 0) {
+            el.find("tbody").html(`
+              <tr>
+                <td class="no-devices" colspan="2">
+                  ${gettext("No devices found!")}
```
nemesifier:
```suggestion
                  ${gettext("No devices found")}
```

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -116,29 +112,148 @@
...
+                  devices = devices.concat(data.results);
+                } else {
+                  devices = data.results;
+                }
+                nextUrl = data.next;
+                renderRows(devices);
+              },
+              error() {
+                console.error("Could not load more devices from", url);
+                alert("Could not load more devices.");
```
nemesifier: Flag as translatable.

.github/workflows/pypi.yml
```diff
@@ -29,4 +29,4 @@ jobs:
       - name: Build package
         run: python -m build
       - name: Publish package distributions to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.13.0
+        uses: pypa/gh-action-pypi-publish@v1.12.4
```
nemesifier: For some reason the versions here are being downgraded, that's wrong, please fix.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -116,29 +112,148 @@
...
+              <input id="device-search" placeholder="Search for devices" />
+            </div>
+            <div class="label-container">
+              ${statusFilterButtons}
+              <input id="status-filter" type="hidden" />
+            </div>
+            <div class="table-container">
+              <table>
+                <thead>
+                    <tr>
```
nemesifier: Indentation is not consistent, 4 spaces are used here while we are using 2.

openwisp_monitoring/device/static/monitoring/js/device-map.js
```diff
@@ -116,29 +112,148 @@
...
+                </td>
+              </tr>
+            `);
+            return;
+          }
+          const rows = devices
+            .map(
+              (device) => `
+            <tr>
+                <td class="col-name"><a href="${device.admin_edit_url}">${device.name}</a></td>
```
nemesifier: Same here, please double check all indentation.

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Removed!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Updated!

openwisp_monitoring/device/static/monitoring/css/floorplan.css
dee077: Updated!

.github/workflows/pypi.yml
dee077: Updated!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Fixed!

openwisp_monitoring/device/static/monitoring/js/device-map.js
dee077: Fixed!

openwisp_monitoring/tests/test_selenium.py
```diff
@@ -38,7 +38,6 @@
 
 
 class SeleniumTestMixin(BaseSeleniumTestMixin):
-    browser = "chrome"
```
pandafy: I accidentally added this in my previous commit

.github/workflows/ci.yml
```diff
@@ -89,6 +89,8 @@ jobs:
           pip install -r requirements-test.txt
           pip install -U -I -e .
           pip uninstall -y Django
+          # TODO: Remove before merging - install branch build of openwisp-utils for CI testing
+          pip install -U -I "openwisp-utils @ https://github.com/openwisp/openwisp-utils/tarball/gsoc25-map-adjustments"
```
pandafy, line 93: Blocker https://github.com/openwisp/openwisp-utils/pull/509

openwisp_monitoring/device/api/urls.py
nemesifier: This is still pending. Let's not wait for this as it can be addressed very quickly and I am worried if we delay this we may forget.

PR #748, opened 2026-02-26
[tests] Fixed hardcoded config_app_label and app_label in device test_admin

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.

## Description of Changes

This adds config_app_label attrs in device test_admin tests to make them derivable in other django apps.

I currently also need to set app_label, since the derived TestImportExportMixin requires it... we could probably devise a slightly nicer way of dealing with it, like f.i. renaming the variable config_app_label (in controller) in TestImportExportMixin to make it more explicit which app it is referring to...

### DIFF
--- a/openwisp_monitoring/device/tests/test_admin.py
+++ b/openwisp_monitoring/device/tests/test_admin.py
@@ -49,7 +49,6 @@ class TestAdmin(
 
     resources_fields = TestImportExportMixin.resource_fields
     resources_fields.append("monitoring_status")
-    app_label = "config"
     _device_params = {
         "group": "",
         "management_ip": "",
@@ -912,7 +911,7 @@ class TestWifiSessionAdmin(
     TestWifiClientSessionMixin,
     TestCase,
 ):
-    config_app_label = "config"
+    config_app_label = Device._meta.app_label
     wifi_session_app_label = WifiSession._meta.app_label
     wifi_session_model_name = WifiSession._meta.model_name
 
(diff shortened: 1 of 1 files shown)


openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -43,6 +43,10 @@ class TestAdmin(
 ):
     """Test the additions of openwisp-monitoring to DeviceAdmin"""
 
+    # Since TestImportExportMixin expects app_label for config app
+    app_label = "config"
+    config_app_label = "config"
+
     resources_fields = TestImportExportMixin.resource_fields
     resources_fields.append("monitoring_status")
     app_label = "config"
```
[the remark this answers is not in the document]
asmodehn: how about replacing this app_label by resolving them from the model _meta nested class instead ?
Also resolving from model in TestImportExportmixin will make app_label definition here redundant ?

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -917,6 +912,7 @@ class TestWifiSessionAdmin(
     TestWifiClientSessionMixin,
     TestCase,
 ):
+    config_app_label = "config"
```
nemesifier: I think this is missing right?

```suggestion
    config_app_label = Device._meta.app_label
```

asmodehn: correct. I ll fix it asap.

PR #775, opened 2026-04-18
Influxdb2.x port on master

Baboux98:
## Checklist

- [ ] I have manually tested the changes proposed in this pull request.

Not applicable

## Description of Changes

This pull request revives and modernizes the InfluxDB 2.x support work originally proposed in #584, adapting it to the current master branch and current project conventions.

## Screenshot

Please include any relevant screenshots.


nemesifier (commented):
Thanks @Baboux98 for picking up the InfluxDB 2.x effort (#274). This is clearly the most complete and most recent of the attempts, so I think it should be the canonical one.

It is a draft and very large (4700+ lines, 27 files), so a few framing points before a detailed review:

1. **Consolidate the InfluxDB 2.x PRs.** There are two older attempts at the same feature (#584 and #354), both stale drafts. Let's make this the single line of work and close the other two, lifting anything still useful from them into here, so reviewers are not splitting attention across three branches.
2. **Backend interface compatibility.** The current `db/backends/influxdb` (1.x) must keep working. The 2.x backend should slot in behind the same backend interface so existing deployments are unaffected and users opt in via setting. Please make that switch explicit and documented.
3. **CI must run against a real InfluxDB 2.x.** Get the test matrix running both 1.x and 2.x so we do not regress either.
4. Bring it out of draft with green CI when ready, and I will go through the query translation, retention policies, and migration path in detail.

This is the priority for the timeseries layer; let's focus it here.

PR #801, opened 2026-05-26
[feature] Added support for InfluxDB 2.x

### LINKED ISSUE #786: [feature:gsoc2026] Add influxDB2.0 backend - core implementation
The current openwisp-monitoring only supports InfluxDB 1.8 as its time-series backend. This issue tracks the implementation of a new `DatabaseClient` for InfluxDB 2.0 using the `influxdb-client` Python library.

The implementation must preserve the current database abstraction layer as much as possible, so higher-level monitoring logic can continue to work without database-specific changes.

## Scope

- Implement a new InfluxDB 2.0 backend under `openwisp_monitoring/db/backends/`.
- Use the `influxdb-client` Python library.
- Add any required optional dependency declarations for `influxdb-client` where this project expects backend dependencies to be declared.
- Support runtime backend selection through the existing `TIMESERIES_DATABASE` configuration mechanism.
- Keep the implementation compatible with the existing `DatabaseClient` interface expected by the rest of OpenWISP Monitoring.

## Required behavior

- Initialization must support InfluxDB 2.0 connection settings: URL, organization, token, and bucket name.
- The current database name concept should map to the InfluxDB 2.0 bucket name unless maintainers decide otherwise during implementation.
- `create_database` must create the target bucket through `buckets_api()`.
- `drop_database` must delete the target bucket through `buckets_api()`.
- `create_or_alter_retention_policy` must account for InfluxDB 2.0 bucket-level retention rules.
- InfluxDB 2.0 has no named retention policies in the same sense as InfluxDB 1.8, and only one bucket-level retention rule can be active. The final behavior for existing retention policy arguments must be documented in the implementation.
- `write` and `batch_write` must write to the configured bucket.
- Existing `database` and `retention_policy` write arguments need an explicit decision. One acceptable option is to keep the signature and log warnings when unsupported arguments are ignored.
- `query` must execute Flux queries and return data in a shape compatible with existing callers.
- InfluxDB 2.0 returns Flux tables and records, while current code expects InfluxDB 1.8-style `ResultSet` behavior with `get_points()`, `keys()`, and `items()`. The issue should remain open to either adapting the core code to a generic return format or wrapping InfluxDB 2.0 results to mimic the existing return type. The current proposal favors wrapping for compatibility.
- `read` must support basic key, field, and tag filtering.
- Unsupported `read` parameters such as complex `where`, `distinct_fields`, `count_fields`, precision, and retention policy need an explicit implementation decision. One acceptable option is raising `NotImplementedError` with migration guidance.
- `delete_metric_data` must provide equivalent deletion behavior through `delete_api()` predicates.
- `get_list_query` must handle manual grouping of Flux records and preserve the time-indexed format expected by charts.
- `get_list_retention_policies` must return retention policy-like data derived from bucket retention rules.
- Chart helper behavior needed by `get_query` must remain compatible with chart rendering code.

## Methods to cover

The backend must implement or adapt the behavior required by these existing public methods:

- `create_database`
- `drop_database`
- `create_or_alter_retention_policy`
- `query`
- `write`
- `batch_write`
- `read`
- `get_list_query`
- `get_list_retention_policies`
- `delete_metric_data`
- `get_query` and related chart helper behavior

## Acceptance criteria

- The InfluxDB 2.0 backend can be selected via `TIMESERIES_DATABASE` without code changes.
- Existing monitoring write, read, delete, and chart code paths keep working with the new backend.
- InfluxDB 2.0-specific limitations are handled explicitly and documented in code or docs.
- Existing callers that expect `get_points()`, `keys()`, and `items()` either keep working through a compatibility wrapper or are migrated deliberately to a new generic abstraction.
- The implementation is covered by the dedicated InfluxDB 2.0 tests tracked separately.

pushpitkamboj:
## Checklist

Closes #786

## Description of Changes

This branch adds InfluxDB 2 support behind the common TSDB interface, so the monitoring code can work with either InfluxDB 1 or InfluxDB 2 without changing the higher-level device and chart logic. It also updates the shared timeseries test setup and CI so the new backend is exercised alongside the existing one.

## Changes
- Added a new influxdb2 backend implementing the same TSDB contract used by the existing backends.
- Extended shared chart/query/test configuration so backend-specific queries exist for both InfluxDB 1 and InfluxDB 2.
- Updated test infrastructure to be backend-agnostic, so common monitoring/device tests can run against either backend.
- Added dedicated CI/test runner coverage for InfluxDB 2.
- Refined a few shared tests and helpers where backend timing/behavior differences needed stabilization.

### logs of openwisp-monitoring server
Screenshot 2026-07-06 at 8 06 53 AM

### logs of openwisp-monitoring dashboard
Screenshot 2026-07-06 at 8 06 25 AM

### logs of influxdb2 time series DB container
Screenshot 2026-07-06 at 8 06 09 AM


nemesifier (commented):
AI-assisted review (Claude Opus 4.7, extended thinking). I traced every public method on `DatabaseClient` through the monitoring, device, and check layers to find interface gaps that would break at runtime. I am not repeating what CodeRabbitAI already flagged (the `__iter__` duplication bug, Flux injection, `get_query` ignoring templates, `group_map` used as lookback instead of aggregation window). Everything below is additional.

## Deviations from the proposal

The proposal (Measurable Outcome #2) states: "Fully working InfluxDB 2.0 backend implementing the same DatabaseClient interface, with **Flux query equivalents written for all 19 chart queries** defined in queries.py, including default_chart_query and device_data_query, ensuring full feature parity with the existing InfluxDB 1.8." The project schedule puts this in Week 4-5: "Translate all 19 chart queries from InfluxQL to Flux."

The implementation has 5 explicit Flux query templates (uptime, packet_loss, rtt, wifi_clients, traffic). The other 14 chart types (general_wifi_clients, general_traffic, memory, cpu, disk, signal_strength, signal_quality, access_tech, bandwidth, transfer, retransmits, jitter, datagram, datagram_loss) use a generic fallback that does only `from(bucket) |> range |> filter(measurement)` with no aggregation, no field selection, no windowing. These will produce wrong chart data.

The PR description says "Added Flux-based query support for monitoring charts" and "Closes #786" without acknowledging this is partial. If this PR is intended as a first step with follow-ups to come, the description should say so explicitly and the issue should not be closed by it.

The proposal also promises "same DatabaseClient interface" but two methods that the monitoring layer calls directly (`delete_series`, `_get_top_fields`) are missing (see inline comments below). The PR description lists the methods that were added but does not mention these gaps.

The `read()` raising `NotImplementedError` for several parameters is documented both in the proposal and in the PR's "Compatibility Notes" section, so the approach itself is justified. The problem is that some of those parameters (`distinct_fields`, `count_fields`) are used by core checks (wifi_clients), so "not supported" translates to "feature broken at runtime." The proposal does not acknowledge this consequence.

## Summary

The general approach is sound: new backend module behind the abstraction layer, `QueryResultSet` for backward compatibility, docker-compose setup, migration guards, contract tests. The foundation is there.

The inline comments below flag things that would break at runtime. Items 1, 2, 3 will crash or return wrong data for any deployment. Items 4, 5, 6 will break specific monitoring checks. Item 7 is a design gap that needs a decision. Item 8 is minor.

I would suggest fixing 1 through 6 before the next review round, and opening a discussion on 7 (retention policy strategy for v2) since it affects the data model. The PR description should also be updated to reflect the actual scope: which chart queries are done, which are pending, and whether this is meant to close #786 or be followed by additional PRs.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,634 @@
...
+                " |> sum() |> map(fn: (r) => ({r with _value: r._value / 1000000000}))"
+            )
+        elif chart_type == "rtt":
+            # RTT: Multiple aggregations (mean of mean, max, min)
+            flux_query += " |> mean()"
+        else:
+            # Default handling: uptime, packet_loss, etc.
+            if summary:
+                flux_query += " |> mean()"
+        return flux_query
```
nemesifier, line 1448: **1. Missing `delete_series` method (will crash at runtime)**

`openwisp_monitoring/monitoring/tasks.py:95` calls `timeseries_db.delete_series(key=key, tags=tags)`. This class does not define `delete_series`. The existing `delete_metric_data` is a different method called from different call sites. This will raise `AttributeError` the moment a metric deletion task fires.

Simplest fix: add a `delete_series` method that delegates to `delete_metric_data`, or alias it.

**2. Missing `_get_top_fields` method (will crash at runtime)**

`openwisp_monitoring/monitoring/base/models.py:729` calls `timeseries_db._get_top_fields(...)`. This method does not exist here. Any chart configuration that sets `top_fields` (like the "applications" chart type) will raise `AttributeError`.

The v1 implementation runs a `SUM(*)` query, sorts results descending, and returns the top N field names. The v2 backend needs a Flux equivalent.

openwisp_monitoring/db/backends/influxdb2/queries.py
```diff
@@ -0,0 +1,97 @@
...
+    "wifi_clients": "num_clients",
+}
+
+default_chart_query = [_default_flux_query, ""]
+
+device_data_query = (
+    'from(bucket: "{0}") |> range(start: -24h) '
+    '|> filter(fn: (r) => r._measurement == "{1}" and r.pk == "{2}") '
+    "|> last()"
+)
```
nemesifier, line 282: **3. `device_data_query` will query a nonexistent bucket**

`openwisp_monitoring/device/base/models.py:155` calls:
```python
q = device_data_query.format(SHORT_RP, self.__key, self.pk)
```

`SHORT_RP = "short"`, so the resulting query becomes `from(bucket: "short") |> ...`. No bucket named "short" exists. The actual bucket is whatever `NAME` is set to (e.g. "openwisp2").

In v1, `{0}` was a retention policy prefix (`short.device_data`), valid InfluxQL. In v2, position `{0}` needs to hold the bucket name, not the retention policy name. This silently returns no data for every device, breaking the device detail page.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,634 @@
...
+        else:
+            flux_query += " |> range(start: -24h)"
+        # Filter by measurement
+        flux_query += f' |> filter(fn: (r) => r._measurement == "{key}")'
+        # Filter by tags
+        if tags:
+            for tag_key, tag_value in tags.items():
+                flux_query += f' |> filter(fn: (r) => r.{tag_key} == "{tag_value}")'
+        # Filter by fields
+        if isinstance(fields, str):
```
nemesifier, line 843: **4. `read()` does not handle `fields="*"` (wildcard)**

`openwisp_monitoring/check/classes/data_collected.py:31` passes `fields="*"` to `timeseries_db.read()`. This code converts it to a Flux filter `r._field == "*"`, which matches a literal asterisk character, not "all fields." The result is always empty, so the `data_collected` check will always report 0.

Fix: detect `fields == "*"` or `fields == ["*"]` and skip the field filter entirely.

**5. `read()` does not handle comma-separated measurement keys**

Same call site (`data_collected.py:30`) passes `key=",".join(set(passive_metrics))`, producing something like `"cpu,memory,disk"`. Line 408 does:
```python
flux_query += f' |> filter(fn: (r) => r._measurement == "{key}")'
```
This tries to match a measurement literally named `"cpu,memory,disk"`. In InfluxQL, `FROM cpu,memory,disk` queries multiple measurements. The Flux equivalent needs an `or` chain or `contains()` filter.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,634 @@
...
+            "extra_fields": "Flux requires explicit field selection instead of wildcard expansion",
+            "distinct_fields": "Use COUNT(DISTINCT) in Flux or custom queries",
+            "count_fields": "Use COUNT aggregation in Flux or custom queries",
+            "retention_policy": "InfluxDB 2.x uses buckets instead of retention policies",
+            "where": "Use custom Flux query instead of WHERE clause shortcuts",
+            "precision": "InfluxDB 2.x uses ISO8601 timestamps; precision parameter not supported",
+        }
+        for param, reason in unsupported_params.items():
+            if param in kwargs:
+                raise NotImplementedError(
```
nemesifier: **6. `read()` raises `NotImplementedError` for `distinct_fields`/`count_fields`, breaking the wifi_clients check**

`openwisp_monitoring/check/classes/wifi_clients.py:86-96` calls:
```python
timeseries_db.read(
    key="wifi_clients",
    fields=["clients"],
    distinct_fields=["clients"],
    count_fields=["clients"],
    ...
)
```

This will raise `NotImplementedError` here, breaking the wifi_clients check entirely on the influxdb2 backend. This is not an obscure edge case, it is one of the standard monitoring checks that runs on every device with wifi interfaces.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,634 @@
...
+    @cached_property
+    def use_udp(self):
+        return False
+
+    @retry
+    def create_or_alter_retention_policy(self, name, duration):
+        """Creates or alters retention policy on a bucket."""
+        from influxdb_client.domain import BucketRetentionRules
+
+        if name != "autogen":
```
nemesifier: **7. `create_or_alter_retention_policy` silently ignores the "short" retention policy**

`openwisp_monitoring/device/utils.py:15` calls `create_or_alter_retention_policy(SHORT_RP, duration)` where `SHORT_RP = "short"`. This implementation only acts on `name == "autogen"` and silently returns for anything else. The "short" retention policy controls how quickly device data snapshots expire. With v2, that configuration is silently lost and device data will follow the default bucket retention instead.

InfluxDB 2.x genuinely cannot do multiple retention policies within one bucket, so this needs a design decision: either route short-RP writes to a separate bucket, or document the limitation and accept that operators will need to set the bucket retention to the shorter value.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,634 @@
...
+            logger.warning(f"Error writing to InfluxDB2: {exception}")
+            raise TimeseriesWriteException
+
+    def batch_write(self, metric_data):
+        """Write multiple data points in batch."""
+        for data in metric_data:
+            if data.get("database") and data.get("database") != self.db_name:
+                logger.warning(
+                    f'Parameter "database" is ignored in InfluxDB 2.7'
+                    f'Writing to bucket "{self.db_name}"'
```
nemesifier, line 730: **8. Minor: `batch_write` warning message has a missing space**

```python
f'Parameter "database" is ignored in InfluxDB 2.7'
f'Writing to bucket "{self.db_name}"'
```

Python concatenates adjacent string literals, producing: `...ignored in InfluxDB 2.7Writing to bucket...`. Missing period and space between the two sentences.

openwisp_monitoring/db/backends/influxdb2/client.py
pushpitkamboj: this requires some serious discussion before implementation, leaving it for next week gsoc meet.

openwisp_monitoring/db/backends/influxdb2/queries.py
pushpitkamboj: I will address the query related issues later once the core implementation methods is being reviewed. As queries require a lot more queries to be written which I have not written yet

docker-compose.yml
```diff
@@ -27,6 +27,20 @@ services:
       INFLUXDB_USER: openwisp
       INFLUXDB_USER_PASSWORD: openwisp
 
+  influxdb2:
+    image: influxdb:2.7-alpine
```
cbeaujoin: Could it be the latest 2.9.1 ? https://github.com/influxdata/influxdb/releases

requirements.txt
```diff
@@ -1,4 +1,5 @@
 openwisp-controller @ https://github.com/openwisp/openwisp-controller/archive/refs/heads/1.3.tar.gz
 influxdb~=5.3.2
```
cbeaujoin: As stated here http://docs.influxdata.com/influxdb/v1/tools/api_client_libraries/
`InfluxDB v2 client libraries are language-specific packages that integrate with the InfluxDB v2 API and support both InfluxDB 1.8+ and InfluxDB 2.x.`

So do we still need this dependency as you added `influxdb-client~=1.50.0` ?

pushpitkamboj:
@coderabbitai please review

.github/workflows/ci.yml
```diff
@@ -118,6 +117,70 @@ jobs:
...
+            gdal-bin \
+            libproj-dev \
+            libgeos-dev \
+            libspatialite-dev \
+            spatialite-bin \
+            libsqlite3-mod-spatialite
+          pip install -U pip wheel setuptools
+          pip install -r requirements-test.txt
+          pip install -U -I -e .
+          pip install -U django~=5.2.0
```
nemesifier: Let's use a basic matrix for consistency and future maintainability.

Can we run this in parallel with the other jobs?

tests/openwisp2/middleware.py
```diff
@@ -0,0 +1,15 @@
+class OpenWispControllerHeaderMiddleware:
```
nemesifier: seems this shouldn't be here

nemesifier (changes requested):

nemesifier (changes requested):
Please sync these branches (gsoc26-add-more-tsdb and feature/influxdb2) with master, you can use merge to make it easier, and we'll rebase later.

docs/developer/installation.rst
```diff
@@ -71,6 +77,19 @@ Install development dependencies:
     pip install -r requirements-test.txt
     npm install -g prettier
 
+If you are using InfluxDB 2.0, export the following environment variables
+before running migrations, celery, or the development server:
+
+.. code-block:: shell
+
+    export TIMESERIES_BACKEND=influxdb2
```
nemesifier, line 92: Can we simplify this and only require this variable (`TIMESERIES_BACKEND`)?

The other variables below can be set but we should auto-default them in the `settings.py` so most developers using the docker containers will only need to export one var during development, which is easier.

The rest of vars below can be commented out and a comment can precede them saying something like "Uncomment these variables if you are not using the influxdb2 container ...etc... provided in the docker compose of this repo ...etc..."

openwisp_monitoring/check/classes/wifi_clients.py
```diff
@@ -83,17 +84,21 @@ def may_execute(cls):
...
                 "content_type": self.related_object._meta.label_lower,
                 "object_id": str(self.related_object.pk),
             },
-            since=(timezone.localtime() - timezone.timedelta(minutes=time_interval)),
-        )
+            "since": (timezone.localtime() - timezone.timedelta(minutes=time_interval)),
+        }
+        values = timeseries_db.read(**read_options)
+        if not values and timeseries_db.use_udp:
+            sleep(0.12)
```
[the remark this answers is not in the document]
nemesifier: @pushpitkamboj this is not looking good

openwisp_monitoring/db/backends/influxdb/tests.py
```diff
@@ -25,15 +27,22 @@
...
+from ... import device_data_query
 from ...exceptions import TimeseriesWriteException
 from .. import timeseries_db
 
 Chart = load_model("monitoring", "Chart")
 Notification = load_model("openwisp_notifications", "Notification")
 
 
-@tag("timeseries_client")
+@tag("timeseries_client", "tsdb_influxdb")
```
nemesifier: What about:

```suggestion
@tag("timeseries_client", "influxdb1")
```

And use `influxdb2` for the new tests.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1090 @@
...
+        if match:
+            return match.group(1)
+        return None
+
+    def _ensure_bucket_exists(self, bucket_name):
+        if bucket_name == self.db_name:
+            self.create_database()
+            return
+        if bucket_name == self._get_bucket_name("short"):
+            from openwisp_monitoring.device import settings as device_settings
```
nemesifier: move to top

openwisp_monitoring/monitoring/migrations/0006_migrate_influxdb_structure.py
```diff
@@ -3,6 +3,11 @@
 
 
 def forward_migration(apps, schema_editor):
+    from openwisp_monitoring.db import timeseries_db
```
nemesifier: move to top

openwisp_monitoring/monitoring/migrations/0012_migrate_signal_metrics.py
```diff
@@ -16,6 +16,11 @@
 
 def forward_migration(apps, schema_editor):
     update_metric_timeseries_structure_forward_migration(apps, schema_editor)
+    from openwisp_monitoring.db import timeseries_db
```
nemesifier: same

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -221,13 +290,20 @@ def setUpClass(cls):
     def tearDownClass(cls):
         timeseries_db.drop_database()
         for metric_name in metrics.keys():
-            unregister_metric(metric_name)
+            try:
+                unregister_metric(metric_name)
+            except ImproperlyConfigured:
+                pass
```
nemesifier: why pass?

openwisp_monitoring/monitoring/tests/__init__.py
```diff
@@ -221,13 +290,20 @@ def setUpClass(cls):
...
         for key in charts.keys():
-            unregister_chart(key)
+            try:
+                unregister_chart(key)
+            except ImproperlyConfigured:
+                pass
```
nemesifier: why pass?

openwisp_monitoring/monitoring/tests/test_charts.py
```diff
@@ -352,3 +344,59 @@ def test_clean_dashboard_traffic_chart(self):
             self.assertDictEqual(
                 DEFAULT_DASHBOARD_TRAFFIC_CHART, {"__all__": ["wan", "eth1", "eth0_2"]}
             )
+
+
+class _TestChartsBackendMixin(TestMonitoringMixin, TestCase):
```
nemesifier: For consistency:
```suggestion
class TestChartsBackendMixin(TestMonitoringMixin, TestCase):
```

We don't generally name classes with leading underscores except rare cases.

openwisp_monitoring/monitoring/tests/test_db_creation.py
```diff
@@ -1,29 +1,60 @@
...
-            mocked_logger.assert_called_with(
-                'Error while executing method "create_database":'
-                f"\n\nAttempt {max_retries} out of {max_retries}.\n"
-            )
-        self.assertEqual(mock.call_count, max_retries)
-        self.assertEqual(sleep_mock.call_count, max_retries - 3)
-        sleep_mock.assert_called_with(delay)
+        with patch(self.patch_target, side_effect=ConnectionError()) as mock:
+            with patch("openwisp_monitoring.utils.sleep") as sleep_mock:
+                with patch("logging.Logger.info") as mocked_logger:
```
nemesifier: can we avoid this nested hell by using method decorators please?

tests/openwisp2/settings.py
```diff
@@ -324,3 +337,6 @@
     from openwisp2.local_settings import *
 except ImportError:
     pass
+
+if "EXTRA_MIDDLEWARE" in globals():
+    MIDDLEWARE = [*EXTRA_MIDDLEWARE, *MIDDLEWARE]
```
nemesifier: What's this?

requirements.txt
nemesifier: @pushpitkamboj please follow up

nemesifier (changes requested):

requirements.txt
pushpitkamboj: We still need influxdb~=5.3.2 because the legacy InfluxDB 1.x backend still imports and uses the old client directly, including the current UDP-specific behavior (use_udp / udp_port): 
- https://github.com/openwisp/openwisp-monitoring/blob/feature/influxdb2/openwisp_monitoring/db/backends/influxdb/client.py#L13-L15
- https://github.com/openwisp/openwisp-monitoring/blob/feature/influxdb2/openwisp_monitoring/db/backends/influxdb/client.py#L95-L134.

nemesifier:
@coderabbitai full review

.github/workflows/ci.yml
```diff
@@ -45,58 +45,25 @@ jobs:
...
-            spatialite-bin \
-            libsqlite3-mod-spatialite
-          pip install -U pip wheel setuptools
-          pip install -r requirements-test.txt
-          pip install -U -I -e .
-          pip install -U ${{ matrix.django-version }}
-          sudo npm install -g prettier
+          django-version: ${{ matrix.django-version }}
 
       - name: Start InfluxDB
```
nemesifier: This could be "Start Timeseries DB"

.github/workflows/ci.yml
```diff
@@ -117,11 +84,61 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           fail-on-error: false
 
+  influxdb2:
```
nemesifier: let's make the tsdb a variable of the same matrix so we have 1 job with muitlple TSDBs

tests/openwisp2/settings.py
```diff
@@ -28,18 +28,34 @@
...
+elif TIMESERIES_BACKEND == "influxdb2":
+    # These defaults mirror the influxdb2 and redis containers defined in the
+    # repository docker-compose.yml, so most local development setups only need
+    # to export TIMESERIES_BACKEND=influxdb2.
+    TIMESERIES_DATABASE = {
+        "BACKEND": "openwisp_monitoring.db.backends.influxdb2",
+        "NAME": os.getenv("INFLUXDB2_BUCKET", "openwisp2"),
+        "HOST": os.getenv("INFLUXDB2_HOST", "localhost"),
+        "PORT": os.getenv("INFLUXDB2_PORT", "8087"),
+        "ORG": os.getenv("INFLUXDB2_ORG", "openwisp"),
```
nemesifier: This is inconsistent, I think we need to keep thing abstract:

```suggestion
        "USER": os.getenv("INFLUXDB2_ORG", "openwisp"),
```

Let's map "organization" to "user".

tests/openwisp2/settings.py
```diff
@@ -28,18 +28,34 @@
...
+        "TOKEN": os.getenv("INFLUXDB2_TOKEN", "openwisp-token"),
```
nemesifier: Same here:
```suggestion
        "PASSWORD": os.getenv("INFLUXDB2_TOKEN", "openwisp-token"),
```

nemesifier (commented):
We need to keep things abstract enough so we can change DBs without going crazy having to remember how each DB calls different things.

Let's stick to:

- name (db name)
- user (db user or organization or whatever the hell they call it)
- password (db password or token or whatever they call it)

nemesifier (changes requested):
Problem found during testing:

```
# open django shell
./manage.py shell_plus
>>> 
# close it with CTRL+D

Exception ignored in: <function InfluxDBClient.__del__ at 0x7f9b4a7a6a20>
  File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/client/influxdb_client.py", line 320, in __del__
  File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 84, in __del__
  File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 661, in _signout
TypeError: 'NoneType' object is not callable
Exception ignored in: <function ApiClient.__del__ at 0x7f9b3d682340>
  File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 84, in __del__
  File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 661, in _signout
TypeError: 'NoneType' object is not callable

```

nemesifier (commented):
EDIT: I found a problem in a staging env but it was due to https://github.com/openwisp/openwisp-radius/issues/749.

requirements.txt
nemesifier: @pushpitkamboj please add comments to help future maintainers remember these details at glance without having to spend energy looking for this info

openwisp_monitoring/db/backends/influxdb/tests.py
```diff
@@ -32,8 +34,14 @@
 Notification = load_model("openwisp_notifications", "Notification")
 
 
-@tag("timeseries_client")
+@tag("timeseries_client", "influxdb1")
 class TestDatabaseClient(TestMonitoringMixin, TestCase):
+    @classmethod
+    def setUpClass(cls):
```
nemesifier: this method is being duplicated in a lot of places, please DRY

openwisp_monitoring/db/backends/influxdb2/queries.py
```diff
@@ -0,0 +1,231 @@
...
+]
+
+
+class DeviceDataQuery:
+    def format(self, retention_policy, measurement, pk):
+        from openwisp_monitoring.db import timeseries_db
+
+        bucket = timeseries_db._get_bucket_name(retention_policy)
+        return (
+            f'from(bucket: "{bucket}") |> range(start: -24h) '
```
nemesifier: This still interpolates `bucket`, `measurement`, and `pk` directly into Flux. A quote or backslash in any of these values can break the query or alter the predicate. Please format these through the backend Flux string helper and add regression tests for escaped device data queries.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+    def _clean_operator(self, op):
+        if op not in self._OPERATORS:
+            raise self.client_error(
+                f'Invalid operator "{op}" passed.\n'
+                f"Valid operators are: {', '.join(self._OPERATORS)}"
+            )
+        return "==" if op == "=" else op
+
+    @staticmethod
+    def _escape_flux_string(value):
```
nemesifier: This helper uses JSON escaping, which can emit escapes such as `\uXXXX`. Flux string literals do not follow JSON escaping rules. Please replace this with Flux-specific escaping and keep all query construction going through that helper.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+                    f"r._measurement == {self._format_flux_string(measurement)}"
+                    for measurement in measurements
+                ]
+            )
+            flux_query += f" |> filter(fn: (r) => {measurement_filter})"
+        # Filter by tags
+        if tags:
+            for tag_key, tag_value in tags.items():
+                flux_query += (
+                    f" |> filter(fn: (r) => r.{tag_key} == "
```
nemesifier: `tag_key` is interpolated as a Flux identifier without validation. Tags can come from metric JSON fields and device-derived data, so this should either validate a strict identifier pattern or use a safe Flux access form. Please cover unsafe tag keys in tests.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+    def _backfill_expected_fields(self, points, query):
+        expected_fields = self._extract_expected_fields(query)
+        if not expected_fields:
+            return points
+        for point in points:
+            for field in expected_fields:
+                point.setdefault(field, None)
+        return points
+
+    def _build_delete_predicate(self, key: str = None, tags: dict = None):
```
nemesifier: The delete predicate escapes values, but not predicate keys. Please validate tag keys before using them in delete predicates, otherwise a crafted tag key can change the delete predicate semantics.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+    def _normalize_chart_window(self, time_value, group_map=None):
+        if group_map and time_value in group_map:
+            return group_map[time_value]
+        if isinstance(time_value, (int, float)):
+            return f"{max(int(time_value), 1)}m"
+        if isinstance(time_value, str) and re.fullmatch(r"\d+", time_value):
+            return f"{max(int(time_value), 1)}m"
+        return time_value
+
+    def _normalize_chart_start_range(self, time_value, group_map=None):
```
nemesifier: This uses `group_map` while normalizing the range start, so `get_query(time="1h", group_map={"1h": "5m"})` becomes `range(start: -5m)` instead of `range(start: -1h)`. The selected chart range and aggregation window must remain separate. Please use `group_map` only for the `aggregateWindow(every: ...)` window.

openwisp_monitoring/db/backends/influxdb2/tests.py
```diff
@@ -0,0 +1,1516 @@
...
+    def test_get_query_basic(self):
+        """Test basic Flux query generation for charts."""
+        query = self.timeseries_db.get_query(
+            chart_type="line",
+            params={"key": "cpu"},
+            time="1h",
+            group_map={"1h": "5m"},
+        )
+        self.assertIn('from(bucket: "', query)
+        self.assertIn("range(start: -5m)", query)
```
nemesifier: This assertion currently locks in the range/window bug. For `time="1h"`, the query should keep `range(start: -1h)` while using `5m` only as the aggregation window. Please assert both pieces separately.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+        self,
+        query,
+        params,
+        chart_type,
+        group_map,
+        number,
+        time,
+        timezone=settings.TIME_ZONE,
+    ):
+        del query, chart_type, timezone
```
nemesifier: `_get_top_fields()` ignores the supplied chart query and chart type, then ranks every field in the base measurement. That can include unrelated fields for charts with custom templates or field constraints. Please preserve the InfluxDB 1 semantics by ranking through the supplied query and chart parameters.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+        return "ResultSet({%s})" % ", ".join(items)
+
+
+class DatabaseClient(BaseTimeseriesClient):
+    """InfluxDB 2.0 client for timeseries database operations."""
+
+    _AGGREGATE = ["mean", "sum", "count", "max", "min", "mode", "stddev"]
+    backend_name = "influxdb2"
+    client_error = InfluxDBError
+    required_settings = ("BACKEND", "TOKEN", "NAME", "ORG", "HOST", "PORT")
```
nemesifier: The client supports `TIMESERIES_DATABASE["URL"]`, but settings validation still requires `HOST` and `PORT`. Please accept either `URL` or the `HOST` plus `PORT` pair, so URL-only InfluxDB 2 deployments do not need unused settings.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+                logger.debug(f'InfluxDB2 bucket "{bucket_name}" not found')
+
+    @cached_property
+    def db(self):
+        """Returns an InfluxDBClient instance."""
+        url = (
+            TIMESERIES_DB.get("URL")
+            or f"http://{TIMESERIES_DB['HOST']}:{TIMESERIES_DB['PORT']}"
+        )
+        if TIMESERIES_DB.get("URL", "").startswith("http://"):
```
nemesifier: This warning checks only the explicit `URL` setting. When `URL` is omitted, the fallback builds `http://HOST:PORT` and sends the token over cleartext without warning. Please check the final computed `url` instead.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+
+        Args:
+            db_name: Bucket name (equivalent to database in InfluxDB 1.x)
+        """
+        self._db = None
+        self.db_name = db_name or TIMESERIES_DB["NAME"]
+        self.org = TIMESERIES_DB.get("ORG", "openwisp")
+        self.token = TIMESERIES_DB.get("TOKEN", "")
+
+    def reset(self, db_name=None):
```
nemesifier: `reset()` drops the cached InfluxDB client and API handles without closing the underlying HTTP client. Please close the cached client before removing it, or add a reusable `close()` method used by `reset()` and short-lived test or management command paths.

openwisp_monitoring/db/backends/base.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+        formatter = getattr(self.device_data_query, "format", None)
+        if not callable(formatter):
+            raise ImproperlyConfigured(
+                "Backend query bundle must define device_data_query.format()."
+            )
+        return self
+
+
+class BaseTimeseriesClient(ABC):
```
nemesifier, line 82: This is the highest-value place to add type hints. Please type the abstract TSDB contract method parameters and return values so future backends implement the same interface consistently. A small `TypedDict` for batch write payloads would also make this boundary much clearer.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+    "_start",
+    "_stop",
+    "_time",
+    "_measurement",
+    "_field",
+    "_value",
+}
+
+
+class QueryResultSet:
```
nemesifier, line 61: Please add focused type hints to this compatibility wrapper, especially `points`, `get_points()`, `keys()`, and `items()`. This is where the InfluxDB 2 Flux response is made to look like the InfluxDB 1 result shape, so explicit types would be useful documentation and would catch shape mismatches.

nemesifier (changes requested):
There are still correctness, security, documentation, and resource-lifecycle issues in the InfluxDB 2 backend. The public settings docs also need to be updated: `docs/user/settings.rst` still documents only the InfluxDB 1 `TIMESERIES_DATABASE` shape, while this PR adds a new supported backend. The type hints request should be focused on the TSDB boundary and compatibility wrapper rather than applied everywhere.

pushpitkamboj:
> Problem found during testing:
> 
> ```
> # open django shell
> ./manage.py shell_plus
> >>> 
> # close it with CTRL+D
> 
> Exception ignored in: <function InfluxDBClient.__del__ at 0x7f9b4a7a6a20>
> Traceback (most recent call last):
>   File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/client/influxdb_client.py", line 320, in __del__
>   File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 84, in __del__
>   File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 661, in _signout
> TypeError: 'NoneType' object is not callable
> Exception ignored in: <function ApiClient.__del__ at 0x7f9b3d682340>
> Traceback (most recent call last):
>   File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 84, in __del__
>   File "/opt/openwisp2/env/lib/python3.11/site-packages/influxdb_client/_sync/api_client.py", line 661, in _signout
> TypeError: 'NoneType' object is not callable
> ```

I didn't noticed this error when I was testing. I diagnosed the error and the failure is in the library’s _signout() teardown path: it calls the module-level helper _requires_expire_user_session(), and during late shutdown that imported global was no longer safely callable, so cleanup aborted with TypeError: 'NoneType' object is not callable. I fixed this by explicitly registering timeseries_db.close() with [atexit](https://docs.python.org/3/library/atexit.html#atexit.register) for the InfluxDB2 backend, so the client is closed earlier, before the destructor reaches the shutdown.

tests/openwisp2/settings.py
pushpitkamboj: done, renamed as u said in settings and other parts of code as well.

openwisp_monitoring/db/backends/influxdb/tests.py
pushpitkamboj: done

openwisp_monitoring/db/backends/influxdb2/queries.py
pushpitkamboj: done

openwisp_monitoring/db/backends/influxdb2/client.py
pushpitkamboj: I am using the [Flux string literal spec](https://github.com/influxdata/flux/blob/master/docs/SPEC.md#string-literals) as the source of truth for this helper. The implementation now follows Flux string escaping rules directly instead of JSON escaping, so query construction is aligned with the language grammar expected by Flux.

openwisp_monitoring/db/backends/influxdb2/client.py
pushpitkamboj: done

tests/openwisp2/settings.py
pushpitkamboj: done

.github/workflows/ci.yml
pushpitkamboj: done.
added a single `include` row for `influxdb2` on Python 3.12 / Django 5.2, so we avoid expanding the full cartesian product while still keeping the dedicated InfluxDB2 coverage.

pushpitkamboj:
> EDIT: I found a problem in a staging env but it was due to [openwisp/openwisp-radius#749](https://github.com/openwisp/openwisp-radius/issues/749).

okay, I don't know how to solve this as I have less context about the mentioned openwisp-radius issue.

pushpitkamboj:
FYI, while adding type hints I hit a Python compatibility issue: typing.Self is available only in Python 3.11+, but our CI still runs on Python 3.10. So I replaced it with a TypeVar in `/db/backends/base.py`

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1117 @@
...
+        timezone=settings.TIME_ZONE,
+    ):
+        del query, chart_type, timezone
+        if number <= 0:
+            return []
+        flux_query = self._build_chart_base_query(params, time, group_map)
+        flux_query += ' |> group(columns: ["_field"]) |> sum()'
+        result = list(self.query(flux_query).get_points())
+        result.sort(key=lambda point: point.get("_value", 0), reverse=True)
+        return [point["_field"] for point in result[:number] if point.get("_field")]
```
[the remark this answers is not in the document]
pushpitkamboj: already resolved

openwisp_monitoring/db/backends/influxdb2/tests.py
```diff
@@ -0,0 +1,1516 @@
...
+        self.assertIn('_measurement == "cpu"', query)
```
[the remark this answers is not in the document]
pushpitkamboj: already solved.

nemesifier:
> FYI, while adding type hints I hit a Python compatibility issue: typing.Self is available only in Python 3.11+, but our CI still runs on Python 3.10. So I replaced it with a TypeVar in `/db/backends/base.py`

We can drop support for Python <= 3.10, as we're targeting to officially release this work in late 2026 / early 2027, by that time Python 3.10 will be EOL.

.github/workflows/ci.yml
```diff
@@ -35,68 +35,43 @@ jobs:
...
+        tsdb:
+          - influxdb
         exclude:
           # Python 3.13 supported only in Django >=5.1.3
           - python-version: "3.13"
             django-version: django~=4.2.0
+        include:
+          - python-version: "3.12"
+            django-version: django~=5.2.0
+            tsdb: influxdb2
```
nemesifier, line 44: Let's test on python 3.13, remove python 3.10, django 4.2 and 5.1, the exclude above can be removed as well.

.github/actions/setup-test-environment/action.yml
```diff
@@ -0,0 +1,54 @@
...
+    required: true
+  django-version:
+    description: Django version constraint to install
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Cache APT packages
+      uses: actions/cache@v6
```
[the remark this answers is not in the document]
pushpitkamboj: @nemesifier I am not sure whether we want to pin GitHub Actions to full commit SHAs or keep using the major version tags here. need you opinion before I change in the PR

openwisp_monitoring/db/backends/tests.py
```diff
@@ -0,0 +1,318 @@
...
+        base_config,
+        clear=True,
+    )
+    @patch("openwisp_monitoring.db.backends.influxdb2.client.InfluxDBClient")
+    def test_db_uses_http_fallback_url(self, mock_client):
+        client = DatabaseClient()
+        client.db
+        mock_client.assert_called_once_with(
+            url="http://localhost:8086", token="token", org="openwisp"
+        )
```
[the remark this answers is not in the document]
pushpitkamboj: we have different test cases to assert the log warning

docs/user/settings.rst
```diff
@@ -59,6 +63,9 @@ The following table describes all keys available in
 ``USER``     User for logging into the timeseries database
 ``PASSWORD`` Password of the timeseries database user
 ``NAME``     Name of the timeseries database
+``URL``      Optional connection URL. For the
```
nemesifier: does this work also for influxdb1?

docs/user/settings.rst
nemesifier: Seems this has been wrong for a long time 🥲 

```suggestion
**type**:    ``dict``
```

docs/user/settings.rst
```diff
@@ -67,6 +74,22 @@ The following table describes all keys available in
...
+    TIMESERIES_DATABASE = {
+        "BACKEND": "openwisp_monitoring.db.backends.influxdb2",
+        "NAME": "openwisp2",
+        "USER": "openwisp",  # InfluxDB organization
+        "PASSWORD": "openwisp-token",  # InfluxDB API token
+        "URL": "http://localhost:8087",
+    }
+
+For the ``openwisp_monitoring.db.backends.influxdb2`` backend, you must
+define either ``URL`` or both ``HOST`` and ``PORT``.
```
nemesifier: This whole section is a bit messy, we have:

1. the default value showing influxdb1
2. the reference table
3. an example showing influxdb2 which also repeats info already stated in the table.

I'd keep the examples close together, avoid repeating things multiple times as it makes the text verbose and avoid writing ``openwisp_monitoring.db.backends.influxdb2`` when mentioning the inflxudb2 backend, just call it "``influxdb2`` backend".

.github/workflows/ci.yml
```diff
@@ -27,76 +27,44 @@ jobs:
...
       - name: QA checks
         run: |
           ./run-qa-checks
 
       - name: Tests
-        if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
+        if: ${{ !cancelled() && steps.setup_env.conclusion == 'success' }}
         run: |
-          ./runtests
+          ./runtests-${{ matrix.tsdb }}
```
nemesifier: The InfluxDB2 matrix entry starts only the `influxdb2` service, but this step runs plain `./run-qa-checks`. The failing job shows QA loading the default InfluxDB 1 settings and trying to create the InfluxDB 1 database on `localhost:8086`, which is not running in this matrix entry. Please make this QA step backend-aware too, for example by exporting `TIMESERIES_BACKEND=${{ matrix.tsdb }}` for `./run-qa-checks`, or otherwise ensure QA uses the same backend as the matrix entry.

docs/user/settings.rst
```diff
@@ -49,6 +49,10 @@ To ensure consistent alerting behavior, this value should match the
         },
     }
 
+The example above shows the default InfluxDB 1.8 configuration.
```
nemesifier: The settings page now shows the InfluxDB2 configuration shape, but the public docs still need the related backend-specific limitations. Please update the existing UDP-related docs to make clear that UDP write support applies only to the InfluxDB 1 backend, not InfluxDB2. Also update `docs/user/intro.rst` to mention InfluxDB2 as a supported timeseries backend. The inline comments mapping `USER` to the InfluxDB2 organization and `PASSWORD` to the API token are fine.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1208 @@
...
+            if "T" not in value and " " in value:
+                value = value.replace(" ", "T", 1)
+            if value.endswith("Z"):
+                return f'time(v: "{value}")'
+            if (
+                "+" not in value[10:]
+                and "-" not in value[10:]
+                and not value.endswith("Z")
+            ):
+                value = f"{value}Z"
```
nemesifier: For datetime inputs this returns `since + timedelta(days=3650)`. If `since` is more than roughly ten years old, the stop time is already in the past, so `read(since=...)` silently excludes recent data. The non-datetime branch already uses a fixed future timestamp. Please use the same fixed future stop time for datetime inputs too.

nemesifier (changes requested):
I am requesting changes because the InfluxDB2 matrix still fails QA with the wrong backend settings, and the public docs still need to mention InfluxDB2 in `docs/user/intro.rst`. The UDP docs should also clarify that UDP writes apply only to the InfluxDB 1 backend, not InfluxDB2.

docs/user/settings.rst
pushpitkamboj: No, url is only for influxdb2

.github/actions/setup-test-environment/action.yml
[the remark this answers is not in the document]
nemesifier: @coderabbitai version tags are fine

.github/actions/setup-test-environment/action.yml
```diff
@@ -0,0 +1,54 @@
...
+          libproj-dev \
+          libgeos-dev \
+          libspatialite-dev \
+          spatialite-bin \
+          libsqlite3-mod-spatialite
+        pip install -U pip wheel setuptools
+        pip install -r requirements-test.txt
+        pip install -U -I -e .
+        pip install -U "${{ inputs.django-version }}"
+        sudo npm install -g prettier
```
nemesifier: Is there any other need for splitting this logic into a separate file? If not, let's unify the YAML file again (easier to read and maintain).

.github/workflows/ci.yml
```diff
@@ -27,76 +27,46 @@ jobs:
...
-          - django~=5.1.0
           - django~=5.2.0
-        exclude:
-          # Python 3.13 supported only in Django >=5.1.3
-          - python-version: "3.13"
-            django-version: django~=4.2.0
+        tsdb:
+          - influxdb
+        include:
+          - python-version: "3.12"
```
nemesifier:
```suggestion
          - python-version: "3.13"
```

openwisp_monitoring/db/backends/influxdb2/tests.py
```diff
@@ -0,0 +1,1741 @@
...
+                "_field": "usage",
+                "_value": 70,
+                "time": "2024-03-25T12:00:00Z",
+                "host": "server1",
+            },
+        ]
+        resultset = QueryResultSet(points)
+
+        iterated_points = list(resultset)
+
```
nemesifier: avoid needless blank lines please

openwisp_monitoring/db/backends/influxdb2/tests.py
```diff
@@ -0,0 +1,1741 @@
...
+    def test_write_uses_retention_policy_bucket(self):
+        """Test writing with a retention policy uses the mapped bucket."""
+        with patch.object(self.timeseries_db, "_write_api") as mock_write_api:
+            self.timeseries_db.write(
+                name="test_measurement",
+                values={"field1": 10},
+                retention_policy=SHORT_RP,
+            )
+
```
nemesifier: avoid needless blank lines

openwisp_monitoring/db/backends/tests.py
```diff
@@ -0,0 +1,382 @@
...
+        }
+        for backend_path, attrs in backend_cached_attrs.items():
+            with self.subTest(backend=backend_path):
+                backend_module = import_module(backend_path)
+                client = backend_module.DatabaseClient(db_name="initial-db")
+                for attr in attrs:
+                    client.__dict__[attr] = object()
+
+                client.reset(db_name="reset-db")
+
```
nemesifier: Here and there's many other instances like this, make the code more compact and only keep blank lines when necessary or where they do help make the code more readable.

runtests
```diff
@@ -1,12 +1,10 @@
...
-coverage run runtests.py
-
-# the next test runs focus on specific backend features, therefore
-# we can skip selenium tests to improve speed and test consistency
-SAMPLE_APP=1 coverage run runtests.py --exclude-tag=selenium_tests
-TIMESERIES_UDP=1 coverage run runtests.py --exclude-tag=selenium_tests
+# full InfluxDB1 coverage lives in runtests-influxdb
+# full InfluxDB2 coverage lives in runtests-influxdb2
+./runtests-influxdb
+./runtests-influxdb2
```
nemesifier: Is it possible to run tests for all the tsdbs in one go? Wouldn't influxdb1 and 2 conflict with one another?
Developers like us use this command to run tests.

I propose that we clean things up, ideally both users and CI run the same script: `./runtests`.
By default the script runs influxdb1, to make it run other tsdbs (eg: influxdb2 only for now, elasticsearch later), we can do:

```
TSDB=influxdb2 ./runtests
```

And we need to document this in the developer section of the docs.

This way the runtests script stay consistent for both CI, developers and AI agents.

There must be one entry point to run tests or it becomes a mess.

nemesifier (commented):
## Unresolved CodeRabbit Comments

### `openwisp_monitoring/db/backends/influxdb2/client.py`, `_delete_range()`

CodeRabbit said the delete stop bound still uses `datetime.now(timezone.utc)`, so future-dated points can survive `delete_series()` or `delete_metric_data()`.

**My opinion**: Valid and still unaddressed. The current code still uses `stop = datetime.now(timezone.utc)`. Since this backend already uses a fixed future stop for open reads, deletes should use the same far-future bound to avoid leaving points created by clock skew or bad device timestamps.

### `runtests-influxdb2`

CodeRabbit said the InfluxDB2 runner only executes the default run and does not exercise `SAMPLE_APP=1` or `TIMESERIES_UDP=1` variants.

**My opinion**: This is fine for now, I wouldn't run SAMPLE_APP=1 on every tsdb because that logic does not depend on the tsdb used, it's pure python/django stuff, let's run those tests with the default tsdb, however we should add a comment in the bash test script to reflect this and avoid ambiguity, @pushpitkamboj please add this.

openwisp_monitoring/db/backends/influxdb2/queries.py
```diff
@@ -0,0 +1,235 @@
...
+    "{ifname_filter}"
+    "{organization_id_filter}"
+    "{location_id_filter}"
+    "{floorplan_id_filter}"
+)
+
+
+def _window(fn):
+    return (
+        f' |> aggregateWindow(every: {{window}}, fn: {fn}, timeSrc: "_start")'
```
nemesifier: This window is still aligned by Flux in UTC. I reproduced this through a real chart read with a 30 day range in Asia/Kolkata: a point written at 2024-03-24 20:00 UTC, which is 2024-03-25 01:30 locally, is shown in a bucket labelled 2024-03-24 05:30 instead of 2024-03-25 00:00. The range boundaries are now correct, but the aggregation bucket boundaries still need to use the request timezone for 24h and larger windows.

nemesifier (changes requested):
I found one remaining timezone issue that I think should be fixed before merging. The start and end range conversion added in e87b8d8c works, but daily Flux aggregation windows are still UTC-aligned. This can put non-UTC users daily chart points under the wrong local day.

openwisp_monitoring/db/backends/influxdb2/queries.py
pushpitkamboj: done

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -1084,12 +1084,29 @@ def _format_chart_query(
             )
             end_range = f", stop: {formatted_end}"
         window = self._normalize_chart_window(time, group_map)
+        try:
+            large_window = self._duration_to_seconds(str(window)) >= 24 * 60 * 60
```
nemesifier: Let's store that multiplication in a constant somewhere.

nemesifier:
@coderabbitai full review

requirements.txt
```diff
@@ -1,4 +1,5 @@
 openwisp-controller @ https://github.com/openwisp/openwisp-controller/archive/refs/heads/1.3.tar.gz
-influxdb~=5.3.2
+influxdb~=5.3.2 # python client for InfluxDB1.8 backend
+influxdb-client~=1.50.0 # python client for influxdb2 backend
```
nemesifier: On one line we use InfluxDB1.8, the next is influxdb2, please be consistent in the case (either InfluxDB or influxdb).

runtests
```diff
@@ -1,12 +1,24 @@
...
+    TIMESERIES_BACKEND=influxdb coverage run runtests.py \
+        "$@" --exclude-tag=selenium_tests
+    SAMPLE_APP=1 TIMESERIES_BACKEND=influxdb coverage run runtests.py \
+        "$@" --exclude-tag=selenium_tests
+    TIMESERIES_UDP=1 TIMESERIES_BACKEND=influxdb coverage run runtests.py \
+        "$@" --exclude-tag=selenium_tests
+elif [ "$TSDB" = "influxdb2" ]; then
+    # SAMPLE_APP and UDP variants are covered by the default backend path.
+    TIMESERIES_BACKEND=influxdb2 coverage run runtests.py \
+        "$@" --exclude-tag=selenium_tests
```
nemesifier: it looks like selenium tests are always excluded.

Before, selenium tests were ran in the first `coverage run runtests.py` and hence skipped in subsequent runs dedicated to testing extensibility (`SAMPLE_APP=1`) and UDP.

In the version being proposed here, it looks to me as if selenium tests are ALWAYS excluded, which is wrong. Please clarify.

nemesifier (changes requested):

runtests
nemesifier: Let's not exclude selenium tests from influxdb2.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1260 @@
...
+        bucket = self._get_bucket_name(kwargs.get("retention_policy"))
+        flux_query = f'from(bucket: "{bucket}")'
+        # Add time range
+        since = kwargs.get("since")
+        if since:
+            timestamp = self._format_flux_time(since)
+            stop = self._format_flux_time(self._get_open_range_stop(since))
+            flux_query += f" |> range(start: {timestamp}, stop: {stop})"
+        else:
+            flux_query += " |> range(start: -24h)"
```
nemesifier: InfluxDB1 does not add a time filter when `since` is omitted, but this limits InfluxDB2 reads to the last 24 hours. That means stored data older than 24 hours can disappear from callers using the compatibility API. I pushed a failing regression test, `test_metric_read_omit_since`, which writes an older point and then calls `Metric.read()` without `since`: InfluxDB1-style behavior would return the point, but the current InfluxDB2 implementation returns an empty result.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1260 @@
...
+                    'You may pass "time" / "-time" to get result sorted '
+                    "in ascending /descending order respectively."
+                ) % {"order": order}
+                raise self.client_error(message)
+        else:
+            flux_query += ' |> sort(columns: ["_time"])'
+        # Apply limit
+        limit = kwargs.get("limit")
+        if limit:
+            flux_query += f" |> limit(n: {limit})"
```
nemesifier: This `limit()` can apply per Flux table, not to the whole result set. In practice, `metric.read(limit=1, order="-time", extra_fields="*")` can return one row per field/table instead of one latest metric point. I pushed a failing test, `test_metric_read_limit_applies_to_latest_point_only`, which writes an older optional field and a newer main field: the current result contains two points instead of the single latest point.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1260 @@
...
+        for bucket in self._get_known_bucket_names():
+            self._delete_range(predicate, bucket=bucket)
+
+    def delete_metric_data(
+        self, key: str | None = None, tags: TimeseriesTags | None = None
+    ) -> None:
+        """Deletes metric data based on measurement and tags."""
+        predicate = self._build_delete_predicate(key=key, tags=tags)
+        if not predicate:
+            self._reset_known_buckets()
```
nemesifier: This unfiltered cleanup path deletes and recreates the buckets. That is a bigger operational change than InfluxDB1, which uses `DROP SERIES FROM /.*/` and keeps the database and retention policies in place. In production, recreating buckets can change bucket IDs and may break external dashboards, bucket-scoped tokens, or anything else tied to bucket metadata. Can we avoid recreating buckets here?

nemesifier (changes requested):
I pushed some changes to improve a few aspects before leaving this review: added failing regression tests for the two InfluxDB2 read compatibility issues, added developer docs for custom Flux timezone placeholders, and added a small comment explaining why the timezone placeholder is needed in built-in Flux windows.

The latest timezone fix looks good, but I found a few production-impact issues in the broader InfluxDB2 backend implementation that should be addressed before merging.

nemesifier:
@coderabbitai is there any remaining issue?

nemesifier:
@coderabbitai resolve the threads that are clearly already addressed by the latest commits

nemesifier:
@coderabbitai also full review, any new major issue apart from the ones already reported?

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1260 @@
...
+            ),
+            floorplan_id_filter=self._format_filter(
+                "floorplan_id", params.get("floorplan_id")
+            ),
+            field_filter=self._format_field_filter(
+                fields, params.get("field_name", "")
+            ),
+        )
+        if summary:
+            formatted = re.sub(
```
nemesifier: I am concerned about summary queries being rewritten with regex.

The main risk is returning seemingly valid but wrong data. These substitutions can turn a windowed chart query into a different Flux pipeline that still executes, but no longer preserves the intended summary semantics, for example grouped summaries, unique client counts, field remapping, or timestamp handling.

InfluxDB1 does use regex in its query builder too, but in a much more limited and safer way. It rewrites or strips the small `GROUP BY time(...)` clause, while preserving the rest of the query. Here the regex rewrites whole Flux pipeline fragments such as `aggregateWindow(...) |> map(...)` and `window() |> unique() |> count() |> duplicate(...)`. That is a broader semantic transformation, so it is more fragile and harder to reason about than the InfluxDB1 case.

This also makes custom chart queries risky. `get_query(summary=True)` can receive arbitrary Flux templates, and a template that happens to contain `aggregateWindow()` may be rewritten even if that aggregation is not the part that should be replaced for summary mode.

The current code is also hiding compatibility decisions inside regex substitutions. For example, `mode` becoming `last()` and the WiFi clients `window |> unique |> count` pipeline becoming `unique |> count` are semantic choices. They should be explicit in the query definitions or query-building logic, not implicit side effects of text replacement.

A better approach would be to remove these regex rewrites and make summary query generation explicit. We can keep the Flux definitions DRY by sharing fragments in `queries.py`, but define normal and summary variants separately for all built-in InfluxDB2 chart queries that currently depend on the summary regex behavior.

For example, `rtt` could be split like this:

```python
_rtt_base = (
    _range
    + _object_filters
    + " |> filter(fn: (r) => r._field =~ /^rtt_(avg|max|min)$/)"
)

_rtt_map = (
    " |> map(fn: (r) => ({r with _field: "
    'if r._field == "rtt_avg" then "RTT_average" '
    'else if r._field == "rtt_max" then "RTT_max" '
    'else "RTT_min"}))'
)

chart_query["rtt"]["influxdb2"] = _rtt_base + _window_mean + _rtt_map
summary_query["rtt"]["influxdb2"] = _rtt_base + " |> mean()" + _rtt_map
```

The same pattern should be applied to the other built-in InfluxDB2 chart queries using `_window_mean`, `_window_sum`, `_window_count`, or `_window_mode`.

`wifi_clients` also needs an explicit summary variant, because its normal query uses a different `window() |> unique() |> count()` pipeline. The normal query should keep that windowed pipeline, while the summary query should explicitly perform the whole-range unique count.

This avoids parsing Flux with regex, keeps the behavior inspectable during review, and still avoids duplicating the full query bodies.

.github/workflows/ci.yml
```diff
@@ -58,16 +57,15 @@ jobs:
           echo 'set man-db/auto-update false' | sudo debconf-communicate >/dev/null
           sudo dpkg-reconfigure man-db
 
-      - name: Set up and Cache Python ${{ matrix.python-version }}
+      - name: Set up and cache Python ${{ matrix.python-version }}
         uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
```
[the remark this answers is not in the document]
nemesifier: No thanks, we're fine with tags, stop with this @coderabbitai.

docs/developer/utils.rst
```diff
@@ -245,6 +245,44 @@ If you don't need to register a new chart but need to change a specific
...
+                ' |> filter(fn: (r) => r._field == "{field_name}")'
+                " |> aggregateWindow(every: {window}, fn: mean, "
+                'timeSrc: "_start"{window_timezone})'
+                " |> map(fn: (r) => "
+                "({{r with _time: date.truncate(t: r._time, "
+                "unit: {window}{window_timezone})}}))"
+            )
+        },
+    }
+
```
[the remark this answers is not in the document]
nemesifier: should be fixed

openwisp_monitoring/db/backends/influxdb2/queries.py
```diff
@@ -0,0 +1,262 @@
...
+_wifi_clients_base = (
+    _range + _object_filters + ' |> filter(fn: (r) => r._field == "{field_name}")'
+)
+_wifi_clients_map = ' |> map(fn: (r) => ({{r with _field: "wifi_clients"}}))'
+_wifi_clients_query = (
+    _wifi_clients_base
+    + " |> window(every: {window}, createEmpty: true{window_timezone})"
+    + ' |> unique(column: "_value")'
+    + " |> count()"
+    + ' |> duplicate(column: "_start", as: "_time")'
```
nemesifier, line 106: This WiFi clients path still leaves the first bucket timestamp unnormalized. With a custom range starting off a 10-minute boundary, for example `15:09`, `window()` creates the first window with `_start = 15:09`, then this duplicate makes that value the chart `_time`. The rest of the series then continues on normal 10-minute boundaries (`15:10`, `15:20`, ...), so the API can return an x-axis like `15:09`, `15:10`, `15:20`. InfluxDB1 does not show this because `GROUP BY time(10m) tz(...)` aligns buckets for us. Please truncate `_time` after this duplicate, like the `_window()` helper does with `date.truncate(...)`, and add a regression test for `wifi_clients` or `general_wifi_clients` using a custom range such as `10:07` to assert all non-null points are on 10-minute boundaries.

nemesifier (changes requested):
I found one remaining InfluxDB2 chart bucketing issue that can show inconsistent x-axis timestamps for WiFi clients charts.

image

nemesifier:
@coderabbitai full review

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1346 @@
...
+                    timezone.utc
+                )
+            value = self._get_timestamp(value)
+        if isinstance(value, str):
+            if re.fullmatch(r"\d{4}-\d{2}-\d{2}", value):
+                value = f"{value}T00:00:00"
+            if "T" not in value and " " in value:
+                value = value.replace(" ", "T", 1)
+            if value.endswith("Z"):
+                return f'time(v: "{value}")'
```
nemesifier: This still interpolates a caller-controlled string into a Flux time literal. A malicious `end` value can break out of `time(v: "...")` and append another Flux stage. The API only validates custom dates when both `start` and `end` are supplied, so a lone `end` can reach this path. The failing regression is `TestInfluxDb2Client.test_get_query_does_not_interpolate_unsafe_end_date`. Please strictly parse and reserialize date strings, or otherwise reject invalid time strings before generating Flux.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1346 @@
...
+            if not self.use_udp or len(payload) > 65000:
+                return self._write_api.write(
+                    bucket=bucket, org=self.user, record=points
+                )
+            self._write_udp(payload, retention_policy=retention_policy)
+        except Exception as exception:
+            self._handle_write_exception(exception)
+
+    def validate_query(self, query: str) -> bool:
+        for word in self._FORBIDDEN:
```
nemesifier, line 643: The validator blocks InfluxQL-style keywords, but Flux write-side-effect functions such as `to()` still pass. If another bug ever allows query text or template input to be injected, this can be chained into TSDB data tampering because chart reads can write forged points to buckets the token can access. The failing regression is `TestInfluxDb2Client.test_validate_query_rejects_flux_write_side_effects`. Please reject Flux side-effect functions such as `to()` and `experimental.to()` for chart queries.

nemesifier (changes requested):
I found two remaining security issues that should be fixed before merging. Both have failing regression tests in this PR now.

nemesifier (commented):
As shared in the chat group, please double check the access technology chart grouping which seems incorrect.

nemesifier:
@coderabbitai we're almost ready to merge, is there any issue which remained unaddressed? Please do one last full review.

openwisp_monitoring/db/backends/influxdb2/queries.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+_window_mean = _window("mean")
+_window_sum = _window("sum")
+_window_count = _window("count")
+# Mode is computed manually because count() overwrites _value with the
+# occurrence count; mode_value keeps the original field value to restore later.
+_window_mode = (
+    _field_group
+    + " |> window(every: {window}, createEmpty: false{window_timezone})"
```
nemesifier, line 47: This is the root cause of the access technology chart inconsistency. `access_tech` is the only built-in InfluxDB2 chart using this manual mode window with `createEmpty: false`, so the query returns only buckets that contain access-tech points. When the selected range has sparse access-tech samples, the chart x-axis collapses to the data segment instead of covering the requested range, unlike the other chart queries. The failing regression test is `TestInfluxDb2ClientIntegration.test_access_tech_chart_keeps_sparse_range_buckets`. Please make this query preserve aligned range buckets while keeping the mode value semantics.

nemesifier (changes requested):
@pushpitkamboj there is one remaining InfluxDB2 chart consistency issue to fix before merge. The failing regression test has already been pushed and demonstrates the problem.

nemesifier:
@coderabbitai any remaining issue?

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1421 @@
...
+            r"(?:\d+[smhdw])+",
+            duration,
+        ):
+            raise ValueError(f'Invalid duration "{duration}"')
+        duration_parts = re.findall(r"(\d+)([smhdw])", duration)
+        return sum(int(value) * mapping[unit] for value, unit in duration_parts)
+
+    def _get_timestamp(self, timestamp=None):
+        """Returns ISO format timestamp."""
+        timestamp = timestamp or now()
```
nemesifier: `timestamp=0` is a valid value, but this line silently throws it away and writes the point at the current time instead. Since we are touching this code, let's fix it properly here and in the influxdb1 client too: change `timestamp or now()` to `if timestamp is None: timestamp = now()`, and add a regression test that writes with `timestamp=0`.

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -0,0 +1,1421 @@
...
+                f"contains(value: r._field, set: [{field_list}]))"
+            )
+        if field_name:
+            return (
+                " |> filter(fn: (r) => "
+                f"r._field == {self._format_flux_string(field_name)})"
+            )
+        return ""
+
+    def _get_summary_chart_query(self, query):
```
nemesifier, line 1270: Custom chart queries that use `aggregateWindow()` or `window()` but do not define a `summary_query` end up with a summary that is just the first time bucket, because unmatched queries fall through here unchanged. With influxdb1 the backend strips the `GROUP BY time`, so the same custom query produces a whole-range summary.

Built-in charts are fine because their summaries live in the backend bundle, but custom charts are not.

### Proposed solution

Enforce this at registration time instead of documenting it:

1. Add a default no-op `validate_chart_config(self, chart_config)` method to `BaseTimeseriesClient` in `openwisp_monitoring/db/backends/base.py`.
2. Implement it in the influxdb2 client:
   - skip if the chart config has no `query` for `influxdb2`;
   - if the `influxdb2` query matches a built-in chart query in `self.queries.chart_query`, verify that `self.queries.summary_query` has a matching summary (this guards the bundle, built-in charts do not define `summary_query` inside the chart config);
   - if the query is custom and the stripped Flux source contains `aggregateWindow(` or `window(`, require `chart_config["summary_query"]["influxdb2"]` to be set; otherwise raise `ImproperlyConfigured`.
3. Call `timeseries_db.validate_chart_config(chart_config)` from `_validate_chart_configuration` in `openwisp_monitoring/monitoring/configuration.py`.
4. Add tests in `openwisp_monitoring/db/backends/influxdb2/tests.py` for the built-in, valid custom, and missing `summary_query` cases.

That way the app fails early and loud on startup when a custom influxdb2 chart query is misconfigured.

nemesifier (changes requested):
Two small issues left, everything else from the previous rounds looks good to me.

nemesifier (approved):
@coderabbitai this should be ready.

nemesifier:
@coderabbitai resolve

PR #815, opened 2026-06-09
[feature] Device list: allowed inspecting and filtering problematic metrics #744

### LINKED ISSUE #774: [feature] Device list: highlight specific issue types for devices in "PROBLEM" state
⚠️ not suited to beginners.

**Is your feature request related to a problem? Please describe.**

Currently, the OpenWISP device list shows a "PROBLEM" status as a catch-all. It is not possible to distinguish between different types of issues (e.g., a device being unreachable vs. a device having high memory usage) without navigating to the individual device detail page, which shows the detail as in the screenshot below:

Image

Additionally, the interface lacks the ability to filter the device list by specific problematic metrics, making it difficult to prioritize troubleshooting tasks.

**Describe the solution you'd like**

We are looking for a solution to the two problems, here's my thoughts.

1.  **Contextual Issue Visibility (Problem 1):**
    * We need a way to show which specific health checks are failing directly from the list view.
    * **Design Goal:** The solution should be accessible to mouse, touch, and keyboard users. While tooltips or hovers are common, we should consider how to handle this for mobile/handheld users and ensure it doesn't clutter the UI for healthy devices.
    * **Data Loading:** To keep the dashboard fast, we suggest this data be loaded on-demand rather than pre-loading details for every row.

2.  **Granular Filtering (Problem 2):**
    * We would like to enable filtering by the type of metric causing the "PROBLEM" state.
    * **Design Goal:** If a user selects "PROBLEM" in the health status filter, they should ideally be able to drill down into specific categories (e.g., Ping, Memory, etc.). I'd like to explore the best way to present these sub-options without overcomplicating the existing filter sidebar.
      A possible solution could be to show an additional filter when users select the PROBLEM state which allows to filter devices which have a specific unealthy metric. Eg:

Image

In the example above, "By monitoring metric" could appear only after "PROBLEM" is selected as health status.

**Describe alternatives you've considered**

* **Static Column:** Adding a dedicated column for "Top Issues." It’s straightforward but might take up too much horizontal real estate.
* **Expandable Rows:** Using a "drawer" or accordion for each device to show health status.
* **Hover-based Tooltips:** These are often the easiest to implement but present significant accessibility and mobile-usability challenges (handheld devices don't have the concept of hover).

pandafy:
## Blockers 

- [x] https://github.com/openwisp/openwisp-utils/pull/700

## Checklist

Closes #774 

## Description of Changes

Added support for inspecting unhealthy metrics directly from the
device changelist through an inline accordion and filtering devices
by specific problem types when the health status is PROBLEM.

This makes it easier to identify and triage problematic devices
without leaving the list view.

## Screenshot

[Screencast from 06-10-26 01:19:59.webm](https://github.com/user-attachments/assets/a91bde56-ad7c-4c09-9a83-a7389dcc10e2)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,7 @@ jobs:
           pip install -U pip wheel setuptools
           pip install -r requirements-test.txt
           pip install -U -I -e .
+          pip install --no-cache-dir --no-deps --force-reinstall https://github.com/openwisp/openwisp-utils/tarball/sub-filter
           pip install -U ${{ matrix.django-version }}
           sudo npm install -g prettier
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,6 @@ jobs:
           pip install -U pip wheel setuptools
           pip install -r requirements-test.txt
           pip install -U -I -e .
-          pip install --no-cache-dir --no-deps --force-reinstall https://github.com/openwisp/openwisp-utils/tarball/sub-filter
           pip install -U ${{ matrix.django-version }}
           sudo npm install -g prettier
 
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -331,7 +331,7 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
         if status == "problem":
             html += format_html(
                 '<div class="device-issues-accordion">'
-                '<div class="issues-content" style="display:none;"></div>'
+                '<div class="issues-content"></div>'
                 '<a href="#" class="issues-toggle" data-device-id="{0}">'
                 "{1}"
                 "</a>"
--- a/openwisp_monitoring/device/admin.py
+++ b/openwisp_monitoring/device/admin.py
@@ -33,7 +33,12 @@ from ..monitoring.admin import MetricAdmin
 from ..settings import MONITORING_API_BASEURL, MONITORING_API_URLCONF
 from . import settings as app_settings
 from .exportable import DeviceMonitoringResource
-from .filters import DeviceFilter, DeviceGroupFilter, DeviceOrganizationFilter
+from .filters import (
+    DeviceFilter,
+    DeviceGroupFilter,
+    DeviceOrganizationFilter,
+    UnhealthyMetricFilter,
+)
 
 DeviceData = load_model("device_monitoring", "DeviceData")
 WifiSession = load_model("device_monitoring", "WifiSession")
@@ -251,7 +256,10 @@ class MetricInline(
 
 class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
     change_form_template = "admin/monitoring/device/change_form.html"
-    list_filter = ["monitoring__status"] + BaseDeviceAdmin.list_filter
+    list_filter = [
+        "monitoring__status",
+        UnhealthyMetricFilter,
+    ] + BaseDeviceAdmin.list_filter
     list_select_related = ["monitoring"] + list(BaseDeviceAdmin.list_select_related)
     list_display = list(BaseDeviceAdmin.list_display)
     list_display.insert(list_display.index("config_status"), "health_status")
@@ -268,11 +276,13 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
             + ("monitoring/js/chart-utils.js",)
             + ("monitoring/js/lib/moment.min.js",)
             + ("monitoring/js/lib/daterangepicker.min.js",)
+            + ("monitoring/js/device-changelist.js",)
         )
         css = {
             "all": (
                 "monitoring/css/percircle.min.css",
                 "monitoring/css/daterangepicker.css",
+                "monitoring/css/device-changelist.css",
             )
             + MetricAdmin.Media.css["all"]
         }
@@ -304,7 +314,7 @@ class DeviceAdmin(BaseDeviceAdmin, NestedModelAdmin):
             health = "yes" if metric.is_healthy else "no"
             icon_url = static(f"admin/img/icon-{health}.svg")
             metric_rows.append(
-                f'<li><img src="{icon_url}" ' f'alt="health"> {metric.name}</li>'
+                f'<li><img src="{icon_url}" alt="health"> {metric.name}</li>'
             )
         return format_html(
             mark_safe(f'<ul class="health_checks">{"".join(metric_rows)}</ul>')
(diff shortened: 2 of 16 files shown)


openwisp_monitoring/device/admin.py
```diff
@@ -313,11 +323,23 @@ def health_checks(self, obj):
...
-            obj.monitoring.get_status_display(),
+        status = obj.monitoring.status
+        status_display = obj.monitoring.get_status_display()
+        html = format_html(
+            '<span class="health-{0}">{1}</span>', status, status_display
         )
+        if status == "problem":
+            html += format_html(
+                '<div class="device-issues-accordion">'
+                '<div class="issues-content" style="display:none;"></div>'
```
nemesifier: can we use css to set display:none?

openwisp_monitoring/device/api/urls.py
```diff
@@ -10,6 +10,11 @@
         views.monitoring_device_list,
         name="api_monitoring_device_list",
     ),
+    re_path(
+        r"^api/v1/monitoring/device/(?P<pk>[^/]+)/metrics/$",
+        views.device_unhealthy_metrics,
+        name="api_device_unhealthy_metrics",
```
nemesifier: What about a more generic name?

```suggestion
        name="api_device_metrics",
```

And allow fetching any metric for the device.

openwisp_monitoring/device/api/views.py
```diff
@@ -333,6 +335,35 @@ def get_queryset(self):
...
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["is_healthy"]
+    queryset = Metric.objects.all()
+
+    def get_parent_queryset(self):
+        return Device.objects.filter(pk=self.kwargs["pk"])
+
+    def get_queryset(self):
+        device_pk = self.kwargs["pk"]
+        device_ct = ContentType.objects.get_for_model(Device)
```
nemesifier, line 367: can we cache this to save the query as this never changes?

openwisp_monitoring/device/filters.py
```diff
@@ -1,10 +1,42 @@
...
+        choices = []
+        for key, config in sorted(get_metric_configuration().items()):
+            if "alert_settings" in config:
+                name = config.get("label", key)
+                choices.append((key, str(name)))
+        return choices
+
+    def filter_queryset(self, request, queryset):
+        if self.value():
+            device_ct = ContentType.objects.get_for_model(Device)
```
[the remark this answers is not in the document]
nemesifier: 👍

openwisp_monitoring/device/static/monitoring/js/device-changelist.js
```diff
@@ -0,0 +1,66 @@
...
+      } else {
+        if ($content.children().length > 0) {
+          $content.slideDown(200, function () {
+            $accordion.addClass("expanded");
+          });
+          $btn.text(gettext("hide issues"));
+          $btn.data("expanded", true);
+        } else {
+          var apiUrl =
+            "/api/v1/monitoring/device/" + deviceId + "/metrics/?is_healthy=false";
```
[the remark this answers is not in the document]
pandafy: @coderabbitai this will add extra work which little benefits.

pandafy:
@coderabbitai full review!

openwisp_monitoring/device/tests/test_admin.py
```diff
@@ -925,6 +927,215 @@ def test_device_admin_recover_button_visibility(self):
...
+        org = self._create_org()
+        self._create_device(organization=org, name="test-device")
+        url = reverse(f"admin:{self.config_app_label}_device_changelist")
+        response = self.client.get(url, {"monitoring__status": "problem"})
+        self.assertContains(response, "ow-filter by-problematic-metric ow-sub-filter")
+        self.assertNotContains(response, "ow-sub-filter hidden")
+        self.assertContains(response, "Ping")
+
+    def test_unhealthy_metric_filter_queryset(self):
+        org = self._get_org()
```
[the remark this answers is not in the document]
pandafy: @coderabbitai this suggestion is bogus

nemesifier (changes requested):
Detal device page on current master:

Image

Detal device page on this branch:

Image

There's a worsening of a couple off aspects:

- show issues is not expected here as the information is redundant with the health checks field
- health checks used to show all metrics, not just  the problematic ones

nemesifier:
@coderabbitai any outstanding issue? Please perform one full review, if no issues are found please resolve.

openwisp_monitoring/device/static/monitoring/css/device-changelist.css
```diff
@@ -0,0 +1,67 @@
...
+}
+.device-issues-accordion .issues-content li {
+  list-style: none;
+  font-size: 12px;
+}
+
+.device-issues-accordion.expanded .issues-content {
+  max-height: 500px;
+  opacity: 1;
+}
```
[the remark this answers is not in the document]
pandafy: @nemesifier I don't think this is useful.

nemesifier:
@coderabbitai any outstanding issue?

openwisp_monitoring/device/api/views.py
```diff
@@ -333,6 +335,33 @@ def get_queryset(self):
 monitoring_device_list = MonitoringDeviceList.as_view()
 
 
+class DeviceMetricsView(BaseProtectedAPIMixin, FilterByParentManaged, ListAPIView):
+    """
+    Returns metrics for a device, optionally filtered by ``is_healthy``.
+    Used by the changelist accordion to show why a device is in PROBLEM state.
+    """
+
+    serializer_class = DeviceMetricSerializer
```
nemesifier, line 355: The new changelist accordion calls `DeviceMetricsView`, but this view uses `Metric.objects.all()` with `DjangoModelPermissions` from `BaseProtectedAPIMixin`, so a GET request requires `view_metric`

A user who can view the device changelist but does not have metric model permission will see the `show issues` link and then get a denied AJAX request. The selenium test logs in as a superuser, and the API multitenancy test explicitly grants `view_metric`, so this regression is not covered.

I am not sure we're giving this permission to non staff users routinely, I don't think we do, so this should be a problem for non-staff users.

I would make the endpoint authorization match the feature: if the user can view the parent device in the device admin, the endpoint should return that device's metrics. The smallest fix is to adjust the permission check for this view to be based on the parent `Device` rather than the `Metric` model, and add a test for an org operator who can view the device changelist but does not have `view_metric`.

nemesifier (changes requested):

.github/workflows/ci.yml
```diff
@@ -82,6 +82,7 @@ jobs:
           pip install -U pip wheel setuptools
           pip install -r requirements-test.txt
           pip install -U -I -e .
+          pip install --no-cache-dir --no-deps --force-reinstall https://github.com/openwisp/openwisp-utils/tarball/sub-filter
```
pandafy: TODO: Remove this before merging

nemesifier:
@coderabbitai full review or resolve if no more major issues are found

openwisp_monitoring/device/api/urls.py
```diff
@@ -10,6 +10,11 @@
         views.monitoring_device_list,
         name="api_monitoring_device_list",
     ),
+    re_path(
+        r"^api/v1/monitoring/device/(?P<pk>[^/]+)/metrics/$",
```
nemesifier: I found an inconsistency here:


```suggestion
        r"^api/v1/monitoring/device/(?P<pk>[^/]+)/metric/$",
```

We use the singular form except on rare occasions.

nemesifier (changes requested):

nemesifier:
@coderabbitai review or resolve

nemesifier:
@coderabbitai full review or resolve if not more issues are found

openwisp_monitoring/device/api/views.py
pandafy: This is already coming from a cache implement in the ContentType model.

nemesifier:
@coderabbitai ping, waiting for you.

nemesifier:
@coderabbitai AFAIK all those issues are already resolved.

nemesifier:
@coderabbitai resolve

nemesifier (approved):

PR #818, opened 2026-06-19
[tests] Fixed iperf3 tests failing in isolation

### LINKED ISSUE #817: [tests] Iperf3 tests fail in isolation due to incorrect handling of seeded monitoring data
While running the iperf3 test suite, I noticed that it fails when executed in isolation.

The test suite contains incorrect assumptions regarding seeded monitoring metrics and charts present in the test database.

**Steps To Reproduce**
1. Run:

python tests/manage.py test openwisp_monitoring.check.tests.test_iperf3

2. Observe failures in test_iperf3_check.

**Expected behavior**
The iperf3 test suite should pass when executed in isolation.

**Actual behavior**
Several assertions fail because seeded monitoring metrics and charts are not excluded correctly.

**Root cause**
- charts_qs excludes "general_clients" while the seeded chart configuration is "gen_wifi_clients"
- metrics_qs filters on the key field instead of configuration
- one assertion uses Chart.objects.count() instead of the filtered queryset

**Proposed fix**
- Exclude gen_wifi_clients instead of general_clients from charts_qs
- Filter metrics_qs using the configuration field
- Use the filtered queryset consistently in assertions

vijetanarwal:
### Problem

The iperf3 test suite passes when executed together with the rest of the test suite but fails when executed in isolation.

The failure is caused by incorrect assumptions regarding seeded monitoring metrics and charts created during migrations:

* `charts_qs` excludes `general_clients`, while the actual chart configuration is `gen_wifi_clients`.
* `metrics_qs` filters on the `key` field, while the seeded objects are identified by their `configuration`.
* One assertion uses `Chart.objects.count()` instead of the filtered queryset, causing seeded charts to be counted.

### Solution

* Update `charts_qs` to exclude `gen_wifi_clients`.
* Update `metrics_qs` to exclude objects using the `configuration` field.
* Replace the raw `Chart.objects.count()` assertion with `self.charts_qs.count()`.

### Testing

```bash
python tests/manage.py test openwisp_monitoring.check.tests.test_iperf3
```

Result:

```text
Ran 11 tests

OK
```

### DIFF
--- a/openwisp_monitoring/check/tests/test_iperf3.py
+++ b/openwisp_monitoring/check/tests/test_iperf3.py
@@ -59,9 +59,11 @@ class TestIperf3(
         "iperf3.openwisptestserver2.com",
     ]
     charts_qs = Chart.objects.exclude(
+        configuration__in=["gen_wifi_clients", "general_traffic"]
+    )
+    metrics_qs = Metric.objects.exclude(
         configuration__in=["general_clients", "general_traffic"]
     )
-    metrics_qs = Metric.objects.exclude(key__in=["general_clients", "general_traffic"])
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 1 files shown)


nemesifier (approved):
Thanks @vijetanarwal 👍

PR #821, opened 2026-06-23
[chores:fix] Fixed collection of metrics via hex UUID path

nemesifier:
## Checklist

Fixing regression introduced in #816.

## Description of Changes

The recent change to url patterns introduced a regression. A similar fix to the same issue was done recently to OpenWISP Controller, the best solution is to reuse the same solution also in OpenWISP Monitoring.

### DIFF
--- a/openwisp_monitoring/device/api/urls.py
+++ b/openwisp_monitoring/device/api/urls.py
@@ -16,7 +16,7 @@ urlpatterns = [
         name="api_device_metric_list",
     ),
     path(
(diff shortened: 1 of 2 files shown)


pandafy (approved):
LGTM!

BTW, I didn't find any documentation for `uuid_any` in openwisp-controller.

nemesifier:
> LGTM!
> 
> BTW, I didn't find any documentation for `uuid_any` in openwisp-controller.

It's not documented, we should not needed it outside of controller and monitoring, so it can remain for internal use only.

Issue #803, opened 2026-05-26
[feature: gsoc26] Add Elasticsearch backend support to OpenWISP Monitoring


pushpitkamboj:
This issue tracks adding Elasticsearch as a supported timeseries backend in `openwisp-monitoring`.

Previous Elasticsearch work already exists and should be used as reference:

- Issue [#99](https://github.com/openwisp/openwisp-monitoring/issues/99): `[timeseries] Add initial support for second timeseries db`
- PR [#164](https://github.com/openwisp/openwisp-monitoring/pull/164): `[timeseries] Add initial support for elasticsearch #99`

That work was never merged, but it is still useful prior art. This issue is meant to consolidate the remaining Elasticsearch work into a single, clearer implementation task.

## Scope

Implement a new backend under `openwisp_monitoring/db/backends/` using the official `elasticsearch` Python client and make it selectable through the existing `TIMESERIES_DATABASE` setting.

The backend should preserve the current abstraction layer as much as possible, so higher-level monitoring code can keep working without backend-specific changes.

### Backend implementation

Implement or adapt the behavior of the existing database client interface, including:

- `create_database`
- `drop_database`
- `create_or_alter_retention_policy`
- `write`
- `batch_write`
- `query`
- `read`
- `get_list_query`
- `get_list_retention_policies`
- `delete_metric_data`
- `get_query`

Notes:

- Elasticsearch index management should be used in place of InfluxDB database creation.
- Retention-related behavior should be mapped to ILM or another documented Elasticsearch mechanism.
- If some InfluxDB-specific arguments such as `retention_policy`, `distinct_fields`, or complex `where` clauses cannot be supported directly, the behavior should be handled explicitly and documented.

## Query parity

Implement Elasticsearch equivalents for the chart queries currently supported by the InfluxDB backend, including:

- all entries in `chart_query`
- `default_chart_query`
- `device_data_query`

The goal is functional parity with the existing backend for:

- regular monitoring charts
- WiFi client and traffic charts
- mobile signal charts
- iperf3 charts

Any unavoidable differences in query semantics should be documented clearly.

## Tests and CI

- Replicate the existing backend test coverage where behavior is expected to match InfluxDB 1.8.
- Add Elasticsearch-specific tests where needed.
- Add a dedicated GitHub Actions build for Elasticsearch so the test suite runs against it on every pull request.

## Acceptance criteria

[...]

PR #829, opened 2026-07-17
[feature] Add support for elasticsearch TSDB

### LINKED ISSUE #786: [feature:gsoc2026] Add influxDB2.0 backend - core implementation

pushpitkamboj:
## Checklist

Closes #786

## Description of Changes

This branch adds Elasticsearch 9 support behind the common TSDB interface, allowing OpenWISP Monitoring to use Elasticsearch as a time series backend without changing the higher-level device, metric, and chart logic.

The implementation follows the same abstraction used by the existing InfluxDB backends. Elasticsearch data is stored using time series data streams, with one document per metric point, ILM-based retention handling, and backend-specific query definitions for the existing monitoring charts.

## Changes

- Added a new `elasticsearch` backend implementing the shared TSDB client contract.
- Used the official Elasticsearch Python client with Elasticsearch 9-compatible connection and authentication options.
- Added TSDS data stream support with index templates, lifecycle policies, retention-specific streams, and namespace-safe cleanup.
- Implemented write, batch write, read, delete, retention policy, device data, and chart query compatibility methods.
- Added Elasticsearch chart and summary query definitions for the existing monitoring charts.
- Updated backend loading/configuration so Elasticsearch can be selected through `TIMESERIES_DATABASE`.
- Updated Docker/settings/docs so Elasticsearch can be started and used during local development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,8 +93,9 @@ jobs:
             spatialite-bin \
             libsqlite3-mod-spatialite
           pip install -U pip wheel setuptools
-          pip install -r requirements-test.txt
-          pip install -U -I -e .
+          # Resolve test dependencies against the matrix Django version from the start.
+          pip install -r requirements-test.txt "${{ matrix.django-version }}"
+          pip install -U -e .
           pip install -U "${{ matrix.django-version }}"
           sudo npm install -g prettier
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,9 @@ jobs:
             tsdb: influxdb2
           - python-version: "3.13"
             django-version: django~=5.2.0
+            tsdb: elasticsearch
+          - python-version: "3.13"
+            django-version: django~=5.2.0
             tsdb: influxdb
             udp: true
           - python-version: "3.13"
@@ -105,6 +108,8 @@ jobs:
 
           if [ "${{ matrix.tsdb }}" = "influxdb2" ]; then
             url="http://localhost:8087/health"
+          elif [ "${{ matrix.tsdb }}" = "elasticsearch" ]; then
+            url="http://localhost:9200"
           else
             url="http://localhost:8086/query?q=SHOW%20DATABASES"
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,14 +109,17 @@ jobs:
           if [ "${{ matrix.tsdb }}" = "influxdb2" ]; then
             url="http://localhost:8087/health"
           elif [ "${{ matrix.tsdb }}" = "elasticsearch" ]; then
-            url="http://localhost:9200"
+            url="http://localhost:9200/_cluster/health?wait_for_status=yellow&timeout=2s&filter_path=status"
           else
             url="http://localhost:8086/query?q=SHOW%20DATABASES"
           fi
 
           for attempt in $(seq 1 30); do
-            if curl -fsS "$url" >/dev/null 2>&1; then
-              exit 0
+            if response=$(curl -fsS "$url" 2>/dev/null); then
+              if [ "${{ matrix.tsdb }}" != "elasticsearch" ] || \
+                printf "%s" "$response" | grep -Eq '"status"[[:space:]]*:[[:space:]]*"(yellow|green)"'; then
+                exit 0
+              fi
             fi
             sleep 2
           done
--- a/docs/developer/installation.rst
+++ b/docs/developer/installation.rst
@@ -11,7 +11,7 @@ Dependencies
 ------------
 
 - Python >= 3.11
-- InfluxDB 1.x or InfluxDB 2.x
+- InfluxDB 1.x, InfluxDB 2.x, or Elasticsearch 9.x
 - fping
 - OpenSSL
 
@@ -64,6 +64,12 @@ well:
 Setup and activate a virtual-environment. (we'll be using `virtualenv
 <https://pypi.org/project/virtualenv/>`_)
 
+If you want to use Elasticsearch 9.x instead, start ``elasticsearch``:
+
+.. code-block:: shell
+
+    docker compose up -d redis elasticsearch
+
 .. code-block:: shell
 
     python -m virtualenv env
@@ -107,6 +113,60 @@ credentials, you can override the defaults with:
     export INFLUXDB2_TOKEN=openwisp-token
     export INFLUXDB2_BUCKET=openwisp2
     export REDIS_HOST=localhost
+
+If you are using Elasticsearch 9.x, export the following environment
+variable before running migrations, celery, or the development server:
+
+.. code-block:: shell
+
+    export TIMESERIES_BACKEND=elasticsearch
+
+If you are using the ``elasticsearch`` and ``redis`` containers provided
+in this repository's ``docker-compose.yml``, no additional variables are
+needed because the default values in ``tests/openwisp2/settings.py``
+already match that setup.
+
+If you are not using the provided containers, or if you changed ports or
+credentials, you can override the defaults with:
+
+.. code-block:: shell
+
+    # Optional overrides for non-default setups
+    export ELASTICSEARCH_URL=http://localhost:9200
+    export ELASTICSEARCH_NAME=openwisp2
+    export REDIS_HOST=localhost
+
+For Elastic Cloud, use ``ELASTICSEARCH_CLOUD_ID`` instead of
+``ELASTICSEARCH_URL``:
+
+.. code-block:: shell
+
+    export ELASTICSEARCH_CLOUD_ID=<cloud-id>
+    export ELASTICSEARCH_API_KEY=<api-key>
+
+If your cluster requires authentication, use one of these methods:
+
+.. code-block:: shell
+
+    # API key authentication
+    export ELASTICSEARCH_API_KEY=<api-key>
+
+    # Bearer token authentication
+    export ELASTICSEARCH_BEARER_AUTH=<bearer-token>
+
+    # Username and password authentication
+    export ELASTICSEARCH_USER=elastic
+    export ELASTICSEARCH_PASSWORD=<password>
+
+For TLS-enabled clusters, use:
+
+.. code-block:: shell
+
+    export ELASTICSEARCH_CA_CERTS=/path/to/http_ca.crt
+    export ELASTICSEARCH_SSL_ASSERT_FINGERPRINT=<fingerprint>
+
+Elasticsearch index templates, refresh behavior, retry policy, and
+lifecycle settings are managed by the backend.
 
 Install WebDriver for Chromium for your browser version from
 https://chromedriver.chromium.org/home and extract ``chromedriver`` to one
--- a/docs/developer/installation.rst
+++ b/docs/developer/installation.rst
@@ -114,8 +114,8 @@ credentials, you can override the defaults with:
     export INFLUXDB2_BUCKET=openwisp2
     export REDIS_HOST=localhost
 
-If you are using Elasticsearch 9.x, export the following environment
-variable before running migrations, celery, or the development server:
+If you are using Elasticsearch 9.x, the only required environment variable
+for local development is:
 
 .. code-block:: shell
 
@@ -126,25 +126,29 @@ in this repository's ``docker-compose.yml``, no additional variables are
 needed because the default values in ``tests/openwisp2/settings.py``
 already match that setup.
 
-If you are not using the provided containers, or if you changed ports or
-credentials, you can override the defaults with:
+Optional Elasticsearch connection overrides are:
 
 .. code-block:: shell
 
-    # Optional overrides for non-default setups
-    export ELASTICSEARCH_URL=http://localhost:9200
+    # Defaults to openwisp2
     export ELASTICSEARCH_NAME=openwisp2
-    export REDIS_HOST=localhost
 
-For Elastic Cloud, use ``ELASTICSEARCH_CLOUD_ID`` instead of
-``ELASTICSEARCH_URL``:
+    # Defaults to localhost and 9200
+    export ELASTICSEARCH_HOST=localhost
+    export ELASTICSEARCH_PORT=9200
+
+    # Defaults to http://<ELASTICSEARCH_HOST>:<ELASTICSEARCH_PORT>
+    export ELASTICSEARCH_URL=http://localhost:9200
+
+If ``ELASTICSEARCH_CLOUD_ID`` is set, it is used instead of
+``ELASTICSEARCH_URL`` or ``ELASTICSEARCH_HOST``/``ELASTICSEARCH_PORT``:
 
 .. code-block:: shell
 
     export ELASTICSEARCH_CLOUD_ID=<cloud-id>
-    export ELASTICSEARCH_API_KEY=<api-key>
 
-If your cluster requires authentication, use one of these methods:
+Advanced authentication and TLS settings are unset by default. If your
+cluster requires authentication, use one of these methods:
 
 .. code-block:: shell
 
@@ -158,15 +162,25 @@ If your cluster requires authentication, use one of these methods:
     export ELASTICSEARCH_USER=elastic
     export ELASTICSEARCH_PASSWORD=<password>
 
+When multiple authentication methods are configured, the backend uses them
+in this order: ``ELASTICSEARCH_API_KEY``, ``ELASTICSEARCH_BEARER_AUTH``,
+then ``ELASTICSEARCH_USER``/``ELASTICSEARCH_PASSWORD``.
+
 For TLS-enabled clusters, use:
 
 .. code-block:: shell
 
     export ELASTICSEARCH_CA_CERTS=/path/to/http_ca.crt
     export ELASTICSEARCH_SSL_ASSERT_FINGERPRINT=<fingerprint>
+    export ELASTICSEARCH_VERIFY_CERTS=true
+
+``ELASTICSEARCH_VERIFY_CERTS`` is passed to the Elasticsearch client only
+when explicitly set; otherwise the client default is used.
 
 Elasticsearch index templates, refresh behavior, retry policy, and
 lifecycle settings are managed by the backend.
+
+Elasticsearch uses HTTP/TCP only, so no UDP or Telegraf service is needed.
 
 Install WebDriver for Chromium for your browser version from
 https://chromedriver.chromium.org/home and extract ``chromedriver`` to one
--- a/docs/developer/installation.rst
+++ b/docs/developer/installation.rst
@@ -162,6 +162,9 @@ cluster requires authentication, use one of these methods:
     export ELASTICSEARCH_USER=elastic
     export ELASTICSEARCH_PASSWORD=<password>
 
+``ELASTICSEARCH_USER`` and ``ELASTICSEARCH_PASSWORD`` must be configured
+together.
+
 When multiple authentication methods are configured, the backend uses them
 in this order: ``ELASTICSEARCH_API_KEY``, ``ELASTICSEARCH_BEARER_AUTH``,
 then ``ELASTICSEARCH_USER``/``ELASTICSEARCH_PASSWORD``.
--- a/docs/developer/installation.rst
+++ b/docs/developer/installation.rst
@@ -169,16 +169,20 @@ When multiple authentication methods are configured, the backend uses them
 in this order: ``ELASTICSEARCH_API_KEY``, ``ELASTICSEARCH_BEARER_AUTH``,
 then ``ELASTICSEARCH_USER``/``ELASTICSEARCH_PASSWORD``.
 
-For TLS-enabled clusters, use:
+For self-managed TLS-enabled clusters, configure an HTTPS endpoint and the
+required certificate options:
 
 .. code-block:: shell
 
+    export ELASTICSEARCH_URL=https://<host>:<port>
     export ELASTICSEARCH_CA_CERTS=/path/to/http_ca.crt
     export ELASTICSEARCH_SSL_ASSERT_FINGERPRINT=<fingerprint>
     export ELASTICSEARCH_VERIFY_CERTS=true
 
 ``ELASTICSEARCH_VERIFY_CERTS`` is passed to the Elasticsearch client only
-when explicitly set; otherwise the client default is used.
+when explicitly set; otherwise the client default is used. Elastic Cloud
+deployments can use ``ELASTICSEARCH_CLOUD_ID`` instead of configuring the
+URL.
 
 Elasticsearch index templates, refresh behavior, retry policy, and
 lifecycle settings are managed by the backend.
@@ -223,13 +227,15 @@ Run tests with (make sure you have the :ref:`selenium dependencies
     TIMESERIES_UDP=1 ./runtests  # InfluxDB 1.x over UDP
     TSDB=influxdb2 ./runtests  # InfluxDB 2.x over HTTP
     TSDB=influxdb2 TIMESERIES_UDP=1 ./runtests  # InfluxDB 2.x over UDP (via Telegraf)
+    TSDB=elasticsearch ./runtests  # Elasticsearch over HTTP
 
 The ``./runtests`` script is the main test entry point. By default it runs
-the InfluxDB test flow. Set ``TSDB=influxdb2`` to run the InfluxDB 2.x
-test flow instead. Set ``TIMESERIES_UDP=1`` to run the UDP flow for the
-selected backend. When using ``TSDB=influxdb2 TIMESERIES_UDP=1``, Telegraf
-must be running because InfluxDB 2.x does not support UDP natively. Using
-``--parallel`` is not supported in this module.
+the InfluxDB test flow. Set ``TSDB=influxdb2`` or ``TSDB=elasticsearch``
+to run the corresponding backend test flow instead. Set
+``TIMESERIES_UDP=1`` to run the UDP flow for an InfluxDB backend. When
+using ``TSDB=influxdb2 TIMESERIES_UDP=1``, Telegraf must be running
+because InfluxDB 2.x does not support UDP natively. Using ``--parallel``
+is not supported in this module.
 
 Run quality assurance tests with:
 
@@ -285,9 +291,10 @@ Run the docker container:
 
     docker compose up
 
-By default, the Docker setup uses InfluxDB 1.8. To use InfluxDB 2.9
-instead, run:
+By default, the Docker setup uses InfluxDB 1.8. To use another supported
+backend instead, run one of the following:
 
 .. code-block:: shell
 
     TIMESERIES_BACKEND=influxdb2 docker compose up
+    TIMESERIES_BACKEND=elasticsearch docker compose up
--- a/docs/developer/installation.rst
+++ b/docs/developer/installation.rst
@@ -289,12 +289,12 @@ Run the docker container:
 
 .. code-block:: shell
 
-    docker compose up
+    docker compose up monitoring
 
 By default, the Docker setup uses InfluxDB 1.8. To use another supported
 backend instead, run one of the following:
 
 .. code-block:: shell
 
-    TIMESERIES_BACKEND=influxdb2 docker compose up
-    TIMESERIES_BACKEND=elasticsearch docker compose up
+    TIMESERIES_BACKEND=influxdb2 docker compose up monitoring
+    TIMESERIES_BACKEND=elasticsearch docker compose up monitoring
--- a/docs/user/settings.rst
+++ b/docs/user/settings.rst
@@ -69,6 +69,23 @@ Alternative ``influxdb2`` backend configuration:
         },
     }
 
+Alternative ``elasticsearch`` backend configuration:
+
+.. code-block:: python
+
+    TIMESERIES_DATABASE = {
+        "BACKEND": "openwisp_monitoring.db.backends.elasticsearch",
+        "NAME": "openwisp2",
+        "URL": "https://localhost:9200",
+        "API_KEY": "openwisp-api-key",
+        "CA_CERTS": "/etc/elasticsearch/certs/http_ca.crt",
+        "OPTIONS": {
+            "refresh": "wait_for",
+            "read_size": 10000,
+            "terms_size": 1000,
+        },
+    }
+
 The following table describes the keys available in the
 ``TIMESERIES_DATABASE`` setting:
 
@@ -88,6 +105,39 @@ The following table describes the keys available in the
              :ref:`timeseries_backend_options` for available options
 ============ =============================================================
 
+The ``elasticsearch`` backend uses ``URL``, or ``HOST`` and ``PORT``, to
+connect to the cluster, and ``NAME`` as the name of the data streams
+created by OpenWISP. It also supports the following additional keys:
+
+========================== ==============================================
+**Key**                    ``Description``
+``CLOUD_ID``               Elastic Cloud identifier, used instead of
+                           ``URL`` or ``HOST`` and ``PORT``
+``API_KEY``                Elasticsearch API key
+``BEARER_AUTH``            Elasticsearch bearer token
+``CA_CERTS``               Path to the CA certificate used to verify the
+                           TLS certificate of Elasticsearch
+``SSL_ASSERT_FINGERPRINT`` SHA-256 fingerprint of the TLS certificate of
+                           Elasticsearch, which can be used instead of
+                           ``CA_CERTS``
+``VERIFY_CERTS``           Boolean, passed to the Elasticsearch client
+                           only when configured, otherwise the default of
+                           the client is used
+========================== ==============================================
+
+``API_KEY``, ``BEARER_AUTH``, and ``USER`` with ``PASSWORD`` are
+alternative authentication methods, which are used in this order when more
+than one is configured. ``USER`` and ``PASSWORD`` must be configured
+together.
+
+.. important::
+
+    Use an ``https://`` URL whenever Elasticsearch credentials are
+    configured: API keys, bearer tokens, and passwords sent over
+    ``http://`` can be read by anyone with access to the network. The
+    backend logs a warning when credentials are configured on an
+    ``http://`` URL.
+
 .. _timeseries_backend_options:
 
 Timeseries Database Options
@@ -104,6 +154,15 @@ Timeseries Database Options
 ``udp_port``   Timeseries database port for writing data using UDP on the
                ``influxdb`` backend, or Telegraf listener port for the
                ``influxdb2`` backend
+``refresh``    Refresh policy applied to writes and deletions. Available
+               only for the ``elasticsearch`` backend. Defaults to
+               ``wait_for``
+``read_size``  Number of documents read from Elasticsearch in a single
+               request. Available only for the ``elasticsearch`` backend.
+               Defaults to ``10000``
+``terms_size`` Maximum number of groups returned by charts which group
+               data by tag. Available only for the ``elasticsearch``
+               backend. Defaults to ``1000``
 ============== ===========================================================
 
 The ``influxdb2`` backend supports UDP writes only through Telegraf.
@@ -181,6 +240,12 @@ forwards the data to InfluxDB 2.x over HTTP.
           bucket = "openwisp2"
           bucket_tag = "bucket"
           exclude_bucket_tag = true
+
+The ``elasticsearch`` backend writes over HTTP only and does not support
+UDP writes. It also accepts the ``http_compress``, ``max_retries``,
+``request_timeout``, and ``retry_on_timeout`` options, which are passed to
+the Elasticsearch client unchanged: when these are not configured, the
+defaults of the client are used.
 
 .. _openwisp_monitoring_default_retention_policy:
 
--- a/docs/user/settings.rst
+++ b/docs/user/settings.rst
@@ -156,7 +156,10 @@ Timeseries Database Options
                ``influxdb2`` backend
 ``refresh``    Refresh policy applied to writes and deletions. Available
                only for the ``elasticsearch`` backend. Defaults to
-               ``wait_for``
+               ``wait_for``, which keeps alert tolerance queries
+               consistent by waiting until new points are searchable.
+               Overriding it can improve write throughput, but may delay
+               or miss alert transitions
 ``read_size``  Number of documents read from Elasticsearch in a single
                request. Available only for the ``elasticsearch`` backend.
                Defaults to ``10000``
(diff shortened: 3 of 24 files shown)


pushpitkamboj:
@coderabbitai do full code review

pushpitkamboj:
@coderabbitai do a full review again

openwisp_monitoring/db/backends/elasticsearch/tests.py
```diff
@@ -0,0 +1,1953 @@
...
+        mock_ensure.assert_not_called()
+
+    @patch.object(
+        DatabaseClient,
+        "_ensure_data_stream_resources",
+        side_effect=[ConnectionError("temporary failure"), None],
+    )
+    def test_create_or_alter_retention_policy_retries_network_errors(self, mock_ensure):
+        self.timeseries_db.create_or_alter_retention_policy(SHORT_RP, "24h0m0s")
+        self.assertEqual(mock_ensure.call_count, 2)
```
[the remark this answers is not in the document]
pushpitkamboj: No change needed: `retry()` sleeps only after the fourth failed attempt, while this test succeeds on the second attempt and never calls `sleep()`

pushpitkamboj:
@coderabbitai please do a full review

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        stream_name = self._get_stream_name(retention_policy)
+        return {
+            "index_patterns": [stream_name],
+            "data_stream": {},
+            "priority": self._OPENWISP_INDEX_TEMPLATE_PRIORITY,
+            "template": {
+                "settings": {
+                    "index.lifecycle.name": self._get_policy_name(retention_policy),
+                },
+                "mappings": {
```
nemesifier, line 462: `fields.*` is shared by every measurement, but `fields.clients` is a MAC address for `wifi_clients` and a numeric count for `wifi_clients_min` and `wifi_clients_max`. Elasticsearch fixes the mapping from the first value: writing MAC data first makes threshold range queries lexical, while writing counts first can reject the later MAC value. Please use type-stable indexed paths or namespace fields by measurement, preserving `_source`, and add an integration test for both write orders and the resulting alert query.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+    def _handle_write_exception(self, exception) -> None:
+        logger.warning("Error writing to Elasticsearch: %s", exception)
+        raise TimeseriesWriteException from exception
+
+    def write(self, name: str, values: TimeseriesFields, **kwargs: Any) -> None:
+        self._check_database_kwarg(kwargs.get("database"))
+        retention_policy = kwargs.get("retention_policy")
+        document = self._build_document(name, values, **kwargs)
+        try:
+            self.db.index(
```
nemesifier, line 796: This write has no stable document ID. If Elasticsearch accepts it but the response is lost, the Celery retry inserts the point again. The bulk path has the same problem after a partial success. That changes sums, counts, charts, and tolerance history. Please create an operation ID before enqueueing the task, use it as `_id`, and treat a retry conflict for that operation as success. A content hash alone would collapse intentional identical samples.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+            return self.db_name
+        return self._validate_data_stream_name(f"{self.db_name}-{retention_policy}")
+
+    def _get_policy_name(self, retention_policy=None) -> str:
+        return f"{self.db_name}-{self._get_retention_policy_name(retention_policy)}-ilm"
+
+    def _get_template_name(self, retention_policy=None) -> str:
+        return f"{self._get_stream_name(retention_policy)}-template"
+
+    def _is_own_resource_name(self, name: str) -> bool:
```
nemesifier: This ownership check is unsafe for destructive operations. `openwisp2` also claims `openwisp2-prod`, so `drop_database()` can delete a separate valid deployment's streams, indices, templates, and ILM policies. Please record and verify an exact OpenWISP owner identifier in resource metadata instead of using a collidable prefix. Add a regression test with both names.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        }
+
+    def validate_query(self, query) -> bool:
+        if not isinstance(query, Mapping):
+            raise ValidationError(
+                {"configuration": _("Elasticsearch queries must be dictionaries.")}
+            )
+        if not query:
+            return False
+        if self._is_openwisp_query(query):
```
nemesifier, line 1569: `validate_query()` accepts native aggregation queries, but `get_list_query()` only parses hits for a query without an OpenWISP marker. A valid custom average, sum, histogram, or grouped aggregation is therefore accepted and rendered as an empty chart. Please either define a small result descriptor and parse supported aggregation responses, or reject unsupported custom aggregations during chart registration. Add Elasticsearch variants of the shared custom chart tests.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+                    tags=tags,
+                    since=kwargs.get("since"),
+                ),
+                self._build_field_exists_filter(
+                    self._get_read_exists_fields(fields, where)
+                ),
+            ),
+            "sort": sort,
+            "__retention_policy": retention_policy,
+            "size": int(self.options.get("read_size", 10000)),
```
nemesifier, line 1258: `read_size` is used as a silent total-result cap. A raw chart with 15,000 points loses the newest 5,000 because it sorts ascending. More importantly, `where` is applied after this capped search, so `limit=1` can return no match even when the next document matches. Please paginate with `search_after`, preserving same-timestamp merging across page boundaries, until the logical limit is satisfied or results are exhausted.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+                point[key] = value
+            if len(point) > 1:
+                points.append(point)
+        return points
+
+    def _merge_points_by_time(self, points):
+        merged = {}
+        order = []
+        for point in points:
+            time_key = point.get("__openwisp_time_key", point.get("time"))
```
nemesifier, line 1865: Merging by timestamp alone loses data. A `cpu,memory` read overwrites one `usage` value when both measurements have the same timestamp, and separate tag series collide too. PR #801 fixed this class of problem by retaining measurement and series identity. Please merge only points with the same timestamp, measurement, and normalized tags, then remove that internal metadata before returning results.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+                precision=precision,
+            )
+        if self._is_openwisp_query(query, "chart"):
+            response = self.query(query, precision=precision)
+            return self._get_chart_points(response, query, precision=precision)
+        if self._is_openwisp_query(query, "grouped_chart"):
+            response = self.query(query, precision=precision)
+            return self._get_grouped_chart_points(response, query, precision=precision)
+        response = self.query(query, precision=precision)
+        fields = query.get("__openwisp_fields") if isinstance(query, Mapping) else None
```
nemesifier, line 1929: Raw points omit fields absent at a timestamp. `AbstractChart.read()` appends every timestamp to `x`, but appends each trace only when its key exists, so sparse `upload` and `download` values become associated with the wrong timestamps. Normalize each point against the selected fields and return `None` for missing values. Please add a chart test asserting every trace has the same length as `x`.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+
+    def _round_chart_value(self, value):
+        if value >= 0:
+            return float(int(value + 0.5))
+        return float(int(value - 0.5))
+
+    def _extract_chart_metric_value(self, aggregation, metric):
+        metric_aggregation = aggregation.get(metric["name"], {})
+        if "doc_count" in metric_aggregation:
+            if metric_aggregation["doc_count"] == 0:
```
nemesifier, line 1721: An empty cardinality bucket has a valid value of zero, but this returns `None` before reading it. The existing WiFi chart contract expects zero for empty `COUNT(DISTINCT(...))` buckets, as asserted by `test_wifi_hostapd`. Preserve zero for empty count and cardinality buckets while retaining `None` for aggregates with no value.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        }
+
+    @retry
+    def _delete_by_query(self, query: Mapping[str, Any]) -> None:
+        refresh = self.refresh
+        if refresh == "wait_for":
+            refresh = True
+        for stream_name in self._get_data_stream_names() + self._get_index_names():
+            try:
+                self.db.delete_by_query(
```
nemesifier: `delete_by_query()` can return HTTP 200 with `timed_out`, `failures`, or version conflicts and leave matching documents behind. Discarding that response reports deletion as successful, especially with `conflicts=proceed`. Please inspect each response and raise on incomplete deletion so the existing retry can run. Cover timeouts, failures, and version conflicts.

runtests
```diff
@@ -7,27 +7,42 @@ TSDB="${TSDB:-influxdb}"
...
     exit 1
 fi
 
+if [ "$TSDB" = "elasticsearch" ] && [ -n "$UDP" ]; then
+    echo "Elasticsearch does not support UDP writes." >&2
+    exit 1
+fi
+
+TEST_ARGS=("$@")
+if [ "$TSDB" = "elasticsearch" ] && [ "$#" -eq 0 ]; then
```
nemesifier: The Elasticsearch CI job now replaces the full suite with only three modules. It skips shared monitoring, device, check, alert, API, chart, cleanup, management-command, and Selenium coverage which found most PR #801 compatibility bugs. The passing Elasticsearch job runs only 123 selected tests, so it is not a parity check. Please run the normal suite with `TIMESERIES_BACKEND=elasticsearch` and exclude only genuinely inapplicable tests.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        self.db_name = self._validate_data_stream_name(
+            TIMESERIES_DB["NAME"] if db_name is None else db_name
+        )
+        self._write_sequence = count()
+
+    def reset(self, db_name: str | None = None) -> None:
+        if db_name is not None:
+            self._validate_data_stream_name(db_name)
+        super().reset(db_name=db_name)
+        self.__dict__.pop("db", None)
```
nemesifier, line 279: Dropping the cached client without calling `Elasticsearch.close()` leaks its transport and connection pool until garbage collection. The shared test lifecycle calls `reset()` repeatedly, and downstream callers can do the same. Please close the cached client before removing it, expose a consistent `close()` method, and test that `reset()` closes the previous client.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        return self.options.get("refresh", "wait_for")
+
+    def _get_client_kwargs(self) -> dict[str, Any]:
+        kwargs = {}
+        if TIMESERIES_DB.get("CLOUD_ID"):
+            kwargs["cloud_id"] = TIMESERIES_DB["CLOUD_ID"]
+        else:
+            url = TIMESERIES_DB.get("URL")
+            if not url:
+                url = f"http://{TIMESERIES_DB['HOST']}:{TIMESERIES_DB['PORT']}"
```
nemesifier, line 309: The final fallback URL is `HTTP`, and the backend accepts API keys, bearer tokens, and basic credentials over it without a diagnostic. This sends reusable credentials in cleartext outside a trusted local network. PR #801 already fixed the equivalent InfluxDB2 issue. At minimum, warn when the final computed URL is `HTTP` and credentials are configured, while retaining unauthenticated loopback HTTP for development.

docs/developer/installation.rst
```diff
@@ -108,6 +114,81 @@ credentials, you can override the defaults with:
...
+    export ELASTICSEARCH_CA_CERTS=/path/to/http_ca.crt
+    export ELASTICSEARCH_SSL_ASSERT_FINGERPRINT=<fingerprint>
+    export ELASTICSEARCH_VERIFY_CERTS=true
+
+``ELASTICSEARCH_VERIFY_CERTS`` is passed to the Elasticsearch client only
+when explicitly set; otherwise the client default is used. Elastic Cloud
+deployments can use ``ELASTICSEARCH_CLOUD_ID`` instead of configuring the
+URL.
+
+Elasticsearch index templates, refresh behavior, retry policy, and
```
nemesifier, line 187: The public `TIMESERIES_DATABASE` documentation still describes only InfluxDB backends. This developer page lists test-environment variables, but users need the production Elasticsearch setting shape, authentication and TLS keys, and the supported option defaults. Please add this to `docs/user/settings.rst`, including intentional `read_size`, `terms_size`, refresh, and cardinality limitations.

nemesifier (changes requested):

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1574 @@
...
+                    self._get_read_exists_fields(fields, where)
+                ),
+            ),
+            "sort": sort,
+            "__retention_policy": retention_policy,
+            "size": int(self.options.get("read_size", 10000)),
+        }
+        response = self.query(query, precision=precision)
+        points = [
+            self._document_to_point(
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. `read()` adds `where` fields to the projection before filtering and removes predicate-only fields afterward at `client.py:1036-1061`. The regression test is at `elasticsearch/tests.py:765-782`. Resolving as stale.

runtests
```diff
@@ -5,27 +5,42 @@ TSDB="${TSDB:-influxdb}"
...
+    TEST_ARGS=(
+        openwisp_monitoring.db.backends.elasticsearch.tests
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. `openwisp_monitoring.db.backends.elasticsearch.tests` now exists and is part of the Elasticsearch test selection. Resolving as stale.

.github/workflows/ci.yml
```diff
@@ -105,6 +108,8 @@ jobs:
 
           if [ "${{ matrix.tsdb }}" = "influxdb2" ]; then
             url="http://localhost:8087/health"
+          elif [ "${{ matrix.tsdb }}" = "elasticsearch" ]; then
+            url="http://localhost:9200"
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. CI now waits for Elasticsearch cluster health to reach yellow or green at `.github/workflows/ci.yml:109-125`. Resolving as stale.

openwisp_monitoring/db/backends/base.py
```diff
@@ -127,6 +127,15 @@ def get_default_chart_query(self, has_object_scope: bool = False) -> str:
...
 
+    def _normalize_chart_window(self, time_value, group_map=None):
+        if group_map and time_value in group_map:
+            return group_map[time_value]
+        if isinstance(time_value, (int, float)) or (
+            isinstance(time_value, str) and time_value.isdigit()
+        ):
+            return f"{max(int(time_value), 1)}m"
+        return time_value
+
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. Numeric chart windows are normalized consistently and covered in both Elasticsearch and InfluxDB2 backend tests. Resolving as stale.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1561 @@
...
+        for field in self._CHART_FILTERS:
+            tag_filter = self._build_tag_filter(field, params.get(field))
+            if tag_filter:
+                filters.append(tag_filter)
+        if not filters:
+            return {"match_all": {}}
+        return {"bool": {"filter": filters}}
+
+    def _normalize_chart_window(self, time, group_map):
+        return str(group_map.get(time, time) or "1d")
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. The Elasticsearch chart-window override now matches the expected behavior and has regression coverage. Resolving as stale.

tests/openwisp2/settings.py
```diff
@@ -61,6 +61,43 @@
...
+                port=os.getenv("ELASTICSEARCH_PORT", "9200"),
+            ),
+        )
+    if os.getenv("ELASTICSEARCH_API_KEY"):
+        TIMESERIES_DATABASE["API_KEY"] = os.getenv("ELASTICSEARCH_API_KEY")
+    elif os.getenv("ELASTICSEARCH_BEARER_AUTH"):
+        TIMESERIES_DATABASE["BEARER_AUTH"] = os.getenv("ELASTICSEARCH_BEARER_AUTH")
+    elif os.getenv("ELASTICSEARCH_USER") and os.getenv("ELASTICSEARCH_PASSWORD"):
+        TIMESERIES_DATABASE["USER"] = os.getenv("ELASTICSEARCH_USER")
+        TIMESERIES_DATABASE["PASSWORD"] = os.getenv("ELASTICSEARCH_PASSWORD")
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. `DatabaseClient.validate_settings()` rejects a one-sided basic-auth pair while preserving API-key and bearer-auth precedence at `client.py:242-246`. Resolving as stale.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1582 @@
...
+                        {
+                            "field_doubles": {
+                                "path_match": "fields.*",
+                                "match_mapping_type": "double",
+                                "mapping": {
+                                    "type": "double",
+                                },
+                            }
+                        },
+                    ],
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. Large snapshots remain in `_source`, and the read filter includes documents where `fields.data` appears in `_ignored` at `client.py:821-839`. The integration test covers a snapshot larger than 8192 characters at `elasticsearch/tests.py:1366-1382`. Resolving as stale.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1582 @@
...
+            )
+        return sorted(policies, key=lambda item: (not item["default"], item["name"]))
+
+    def _get_timezone(self, timezone_name=None):
+        if not timezone_name:
+            return timezone.utc
+        try:
+            return ZoneInfo(str(timezone_name))
+        except Exception:
+            return timezone.utc
```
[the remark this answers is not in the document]
nemesifier: I agree this should be consistent. Request-facing views validate the timezone, but the client should not convert the range as UTC and then send the invalid original value to `date_histogram`. Please resolve the timezone once and use the resolved value for both timestamp conversion and histogram bucketing, either rejecting an invalid name or consistently falling back to UTC.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1582 @@
...
+                    since=kwargs.get("since"),
+                ),
+                self._build_field_exists_filter(
+                    self._get_read_exists_fields(fields, where)
+                ),
+            ),
+            "sort": sort,
+            "__retention_policy": retention_policy,
+            "size": int(self.options.get("read_size", 10000)),
+        }
```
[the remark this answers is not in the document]
nemesifier: I agree with the truncation finding, but setting `size=limit` or logging a warning is not sufficient. `where` is applied after same-timestamp merging in Python, so the first page can contain no match while a later page does. Raw charts also silently drop all points after `read_size`. Please paginate with `search_after`, preserve logical-point merging across page boundaries, and stop only after satisfying `limit` or exhausting the results.

openwisp_monitoring/db/backends/elasticsearch/tests.py
```diff
@@ -0,0 +1,1953 @@
...
+        original_db = timeseries_db.__dict__.get("db")
+        timeseries_db.__dict__["db"] = mock_db
+        try:
+            with self.assertRaises(TimeseriesWriteException):
+                timeseries_db.write("test_write", {"value": 1})
+        finally:
+            if original_db is None:
+                timeseries_db.__dict__.pop("db", None)
+            else:
+                timeseries_db.__dict__["db"] = original_db
```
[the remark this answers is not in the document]
nemesifier: Fixed on the current head. The test now uses `patch.dict(timeseries_db.__dict__, {"db": mock_db})` at `elasticsearch/tests.py:1878-1885`. Resolving as stale.

tests/openwisp2/settings.py
```diff
@@ -61,6 +63,46 @@
...
+            "http://{host}:{port}".format(
+                host=os.getenv("ELASTICSEARCH_HOST", "localhost"),
+                port=os.getenv("ELASTICSEARCH_PORT", "9200"),
+            ),
+        )
+    if os.getenv("ELASTICSEARCH_API_KEY"):
+        TIMESERIES_DATABASE["API_KEY"] = os.getenv("ELASTICSEARCH_API_KEY")
+    elif os.getenv("ELASTICSEARCH_BEARER_AUTH"):
+        TIMESERIES_DATABASE["BEARER_AUTH"] = os.getenv("ELASTICSEARCH_BEARER_AUTH")
+    elif os.getenv("ELASTICSEARCH_USER") or os.getenv("ELASTICSEARCH_PASSWORD"):
```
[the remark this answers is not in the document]
nemesifier: This finding is incorrect. The `or` intentionally preserves a one-sided credential so `DatabaseClient.validate_settings()` can reject an unmatched `USER` and `PASSWORD` pair at `openwisp_monitoring/db/backends/elasticsearch/client.py:242-246`. Changing this to `and` would hide the invalid configuration and bypass fail-fast validation. Resolving this thread.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        return {"bool": {"filter": filters}}
+
+    def _normalize_chart_window(self, time_value, group_map=None):
+        if group_map and time_value in group_map:
+            return group_map[time_value]
+        if isinstance(time_value, (int, float)):
+            return f"{max(int(time_value), 1)}m"
+        if isinstance(time_value, str) and re.fullmatch(r"\d+", time_value):
+            return f"{max(int(time_value), 1)}m"
+        return time_value
```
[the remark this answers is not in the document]
nemesifier: I agree this duplication should be removed. The implementations are identical and this project follows DRY. Please move the shared behavior to `BaseTimeseriesClient` and keep focused coverage for the base behavior.

openwisp_monitoring/db/backends/elasticsearch/queries.py
```diff
@@ -0,0 +1,164 @@
...
+        def replace_value(value):
+            if isinstance(value, str):
+                return value.replace(old, new)
+            if isinstance(value, dict):
+                return {key: replace_value(item) for key, item in value.items()}
+            if isinstance(value, list):
+                return [replace_value(item) for item in value]
+            return value
+
+        return ElasticsearchQuery(replace_value(query))
```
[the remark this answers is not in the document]
nemesifier: This finding is incorrect. `AbstractChart.get_top_fields()` calls `self._default_query.replace(...)` in `openwisp_monitoring/monitoring/base/models.py:720-728`. The method is required by the current top-fields path and must not be removed. Resolving this thread.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+        limit=None,
+        precision="s",
+    ) -> list[TimeseriesPoint]:
+        response = self.query(
+            {
+                "size": 0,
+                "query": self._build_base_query(
+                    key=key, tags=tags, since=since, where=where
+                ),
+                "aggs": {"count": {"cardinality": {"field": f"fields.{field}"}}},
```
nemesifier: This does not preserve the shared `COUNT(DISTINCT(...))` behavior. InfluxDB1 uses exact `COUNT(DISTINCT(...))`, and InfluxDB2 uses exact `distinct() |> count()`. Elasticsearch `cardinality` is an approximate HyperLogLog++ estimate. This result is used by `WifiClients._get_wifi_clients_count()`, so an estimate can change `wifi_clients_min` or `wifi_clients_max` health and alert decisions. The default threshold of 50 makes an error unlikely, but Elasticsearch does not guarantee an exact result. Please make `_count_distinct_read()` exact, for example by paging all unique values with a single-field composite aggregation. WiFi charts may keep approximate cardinality for scalability if that limitation is documented. Add a test large enough to verify exact health-check output.

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1655 @@
...
+    def db(self) -> Elasticsearch:
+        return Elasticsearch(**self._get_client_kwargs())
+
+    @property
+    def options(self) -> Mapping[str, Any]:
+        return TIMESERIES_DB.get("OPTIONS", {})
+
+    @property
+    def refresh(self):
+        return self.options.get("refresh", "wait_for")
```
nemesifier, line 299: Non-blocking documentation note: I checked this in detail and I do not want the default changed. Keep `refresh="wait_for"`. OpenWISP queries the time series database immediately after writes when evaluating alert tolerance. With `refresh=false`, that query may not see the point just written and may delay or miss an alert transition. `wait_for` normally waits for the next refresh rather than forcing one, but it can keep Celery workers occupied and reduce throughput on a slow cluster. Please document that this consistency versus throughput tradeoff is intentional.

nemesifier (changes requested):
The PR description needs two corrections:

1. `Closes #786` points to the InfluxDB2 issue handled by PR #801. This Elasticsearch work implements #803, so please use the correct issue reference.
2. The description still says TSDS, but the current implementation uses ordinary Elasticsearch data streams. Please update the wording so it describes what the code now does.

The inline distinct-count comment is the blocking issue in this review. The `refresh="wait_for"` comment is a non-blocking documentation note.

docs/developer/installation.rst
```diff
@@ -208,9 +291,10 @@ Run the docker container:
...
 
-By default, the Docker setup uses InfluxDB 1.8. To use InfluxDB 2.9
-instead, run:
+By default, the Docker setup uses InfluxDB 1.8. To use another supported
+backend instead, run one of the following:
 
 .. code-block:: shell
 
     TIMESERIES_BACKEND=influxdb2 docker compose up
+    TIMESERIES_BACKEND=elasticsearch docker compose up
```
nemesifier: This command does not select only the Elasticsearch stack. `docker compose up` starts every declared service, including InfluxDB 1 and Telegraf, which both publish UDP ports `8089-8092`; startup can fail on those collisions before the monitoring container is usable.

Please target the `monitoring` service so Compose starts only its selected backend and Redis dependencies, or introduce profiles which activate one backend stack at a time.

Severity: P2

docs/user/settings.rst
```diff
@@ -104,6 +154,15 @@ Timeseries Database Options
 ``udp_port``   Timeseries database port for writing data using UDP on the
                ``influxdb`` backend, or Telegraf listener port for the
                ``influxdb2`` backend
+``refresh``    Refresh policy applied to writes and deletions. Available
+               only for the ``elasticsearch`` backend. Defaults to
+               ``wait_for``
```
nemesifier: The `refresh=wait_for` default is intentional but its operational tradeoff is still missing. It keeps a write worker occupied until the next refresh so the immediate tolerance query sees the new point; overriding it improves throughput but can delay or miss an alert transition.

Please document this consistency-versus-throughput behavior next to the option rather than presenting only the default.

Severity: P3

openwisp_monitoring/db/backends/base.py
```diff
@@ -107,11 +110,20 @@ def reset(self, db_name: str | None = None) -> None:
...
+        if isinstance(time_value, str) and re.fullmatch(r"\d+", time_value):
+            return f"{max(int(time_value), 1)}m"
+        return time_value
+
     def get_default_chart_query(self, has_object_scope: bool = False) -> str:
         default_query = self.queries.default_chart_query
         resolver = getattr(default_query, "resolve", None)
         if callable(resolver):
-            return resolver(has_object_scope=has_object_scope)
+            return resolver()
```
nemesifier: This removes the existing `has_object_scope` argument from resolver calls. A custom backend descriptor whose `resolve()` requires that keyword now raises `TypeError`, and resolvers can no longer select an object-scoped default query.

Please preserve `resolver(has_object_scope=has_object_scope)` and let `ElasticsearchQuery.resolve()` accept the optional argument even if Elasticsearch does not use it. Add a base contract test for forwarding.

Severity: P2

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1998 @@
...
+                            "boolean_fields": {
+                                "path_match": "indexed_fields.boolean.*",
+                                "mapping": {
+                                    "type": "boolean",
+                                },
+                            }
+                        },
+                    ],
+                    "properties": {
+                        "@timestamp": {"type": "date"},
```
nemesifier: Elasticsearch `date` fields index only milliseconds, while OpenWISP writes and deletes microsecond timestamps. This makes exact deletion unsafe:

1. Write points at `T+100us` and `T+900us`.
2. Both are indexed at the same millisecond.
3. The exact range built by `delete_metric_data(timestamp=...)` matches and deletes both points.

Please map `@timestamp` as `date_nanos` and add an integration test with two same-millisecond points which verifies that deleting one preserves the other.

Severity: P2

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1998 @@
...
+                TIMESERIES_DB["PASSWORD"],
+            )
+        credentials = kwargs.keys() & {"api_key", "bearer_auth", "basic_auth"}
+        if credentials and url and urlparse(url).scheme == "http":
+            logger.warning(
+                'Sending Elasticsearch credentials over insecure HTTP at "%s". '
+                'Consider switching TIMESERIES_DATABASE["URL"] to https:// when '
+                "running outside local development.",
+                url,
+            )
```
nemesifier: The HTTP warning has two credential handling gaps:

- URL-only credentials such as `http://user:password@host` are sent without a warning.
- If an explicit API key or token is also configured, the warning logs the URL verbatim and exposes the embedded password.

Please treat URL userinfo as credentials and log only a redacted endpoint without the username or password.

Severity: P2

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1998 @@
...
+        }
+        if params.get("time"):
+            histogram["extended_bounds"] = {
+                "min": self._get_timestamp(params["time"], timezone_name=timezone),
+                "max": self._get_timestamp(
+                    params.get("end_date") or now(), timezone_name=timezone
+                ),
+            }
+        if timezone:
+            histogram["time_zone"] = timezone
```
nemesifier, line 1482: Invalid timezone names are converted to UTC when the range timestamps are serialized, but the original invalid name is still sent as `date_histogram.time_zone`. Aggregate charts then fail with an Elasticsearch 400 response, while raw charts silently use UTC for the same input.

Please resolve the timezone once and use the resolved value consistently for range conversion and every histogram, with aggregate, grouped, and raw regression coverage.

Severity: P2

openwisp_monitoring/db/backends/elasticsearch/client.py
```diff
@@ -0,0 +1,1998 @@
...
+    _MAX_DATA_STREAM_NAME_BYTES = 255
+    _OPERATORS = ("=", "!=", "<", ">", "<=", ">=")
+    _DURATION_PATTERN = re.compile(r"(?:\d+[smhdw])+")
+    _DURATION_PART_PATTERN = re.compile(r"(\d+)([smhdw])")
+    _DEFAULT_ROLLOVER_SECONDS = 30 * 24 * 60 * 60
+    _PIT_KEEP_ALIVE = "1m"
+    _DISTINCT_PAGE_SIZE = 1000
+    # highest precision supported by the cardinality aggregation,
+    # counts below this threshold are exact
+    _CARDINALITY_PRECISION = 40000
```
nemesifier: This comment promises exact cardinality below the precision threshold, but Elasticsearch documents those counts as approximate and says the observed low-cardinality accuracy is not guaranteed. This path is used by WiFi charts and summaries, although the health check correctly uses a separate exact composite aggregation.

Please correct the comment and document the chart-level approximation so users do not treat these values as exact.

Severity: P3

openwisp_monitoring/db/backends/influxdb/client.py
```diff
@@ -402,23 +402,36 @@ def get_device_data_query(self, retention_policy, measurement, pk):
...
+    def _delete_point(self, key=None, tags=None, timestamp=None):
+        conditions = [f"time = '{self._get_timestamp(timestamp)}'"]
+        for tag_key, tag_value in (tags or {}).items():
+            conditions.append(f"\"{tag_key}\" = '{self._escape_tag_value(tag_value)}'")
+        measurement = f'FROM "{key}" ' if key else ""
+        self.query(f'DELETE {measurement}WHERE {" AND ".join(conditions)}')
+
+    @staticmethod
+    def _escape_tag_value(value):
+        return str(value).replace("'", r"\'")
```
nemesifier: The new destructive InfluxQL query escapes single quotes but not backslashes. A tag value ending in a backslash produces a malformed literal, while backslash-plus-quote combinations can change how the remainder of the query is parsed.

Please reuse the existing `_format_string_value()` escaping or apply the same backslash-before-quote order here, and cover backslashes combined with quotes.

Severity: P2

openwisp_monitoring/db/backends/influxdb2/client.py
```diff
@@ -1061,11 +1080,33 @@ def _build_delete_predicate(
...
+        """
+        if timestamp is None:
+            return OPEN_RANGE_START, OPEN_RANGE_STOP
+        if not isinstance(timestamp, datetime):
+            timestamp, _value = self._parse_flux_time(str(timestamp))
+            if timestamp is None:
+                raise ValueError(f'Invalid timestamp: "{_value}"')
+        if timestamp.tzinfo is None:
+            timestamp = timestamp.replace(tzinfo=timezone.utc)
+        return timestamp, timestamp + timedelta(milliseconds=1)
```
nemesifier: This one-millisecond deletion range can remove neighboring points:

1. OpenWISP can write same-series points at `T` and `T+1us`.
2. Deleting `T` uses the range `[T, T+1ms)`.
3. InfluxDB therefore deletes both points, contrary to the single-point behavior required by issue #848.

Please use the smallest supported interval, which is already represented by `_RANGE_STOP_EPSILON`, and add an integration test proving that a point one microsecond later is preserved.

Severity: P2

nemesifier (changes requested):
Good progress, see my comments below

nemesifier (approved):

PR #866, opened 2026-09-10
[tests] Added query-count subtests to shared monitoring tests #855

### LINKED ISSUE #855: [tests] Use query-count subtests in shared test classes
**Is your feature request related to a problem? Please describe.**

Monitoring tests use Django's `assertNumQueries()`. When the query count is wrong, the test stops immediately and later assertions do not run.

OpenWISP provides `AssertNumQueriesSubTestMixin`, which records the query-count check as a subtest. This shows the query regression without hiding other failures in the same test.

**Describe the solution you'd like**

Identify shared Monitoring test mixins used by tests with query-count assertions.

Add `AssertNumQueriesSubTestMixin` to the appropriate shared test mixins, following the approach used by Firmware Upgrader's `TestAPIUpgraderMixin`.

Existing `self.assertNumQueries()` calls must remain unchanged. The mixin provides the same API and changes their behavior automatically.

Do not add the mixin to unrelated test classes which do not use query-count assertions.

**Describe alternatives you've considered**

Keep using Django's built-in `assertNumQueries()`. This works, but a query-count failure can hide later assertions.

Import the mixin in every individual test class. This duplicates inheritance declarations and makes the test structure less consistent.

**Additional context**

`AssertNumQueriesSubTestMixin` is provided by `openwisp_utils.tests`. No existing issue covers this change.

Chadokar:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

Closes #855.

## Description of Changes

Added `AssertNumQueriesSubTestMixin` to the shared `TestDeviceMonitoringMixin` and to `TestDashboardTimeseriesView`, while keeping all existing `assertNumQueries()` calls unchanged.

The mixin is placed before `CreateConfigTemplateMixin` so its `assertNumQueries()` implementation takes precedence in the method resolution order.

`TestDashboardTimeseriesView` uses the mixin directly because `CreateConfigTemplateMixin` appears before the general `TestMonitoringMixin` in its inheritance chain. Adding the query-count mixin only to `TestMonitoringMixin` would therefore not override the existing query-count implementation and would also affect unrelated monitoring tests.

Verification performed:

- `openwisp-qa-format`
- `./run-qa-checks`
- Focused Monitoring and Device test suites: 183 tests passed
- Full `./runtests`: 408, 299, and 265 tests passed
- Manual negative-path check confirmed that execution continues after an intentional query-count mismatch

## Screenshot

N/A. This change affects test behavior only.

### DIFF
--- a/openwisp_monitoring/device/tests/__init__.py
+++ b/openwisp_monitoring/device/tests/__init__.py
@@ -7,6 +7,7 @@ from django.utils.timezone import now
 from swapper import load_model
 
 from openwisp_controller.config.tests.utils import CreateConfigTemplateMixin
+from openwisp_utils.tests import AssertNumQueriesSubTestMixin
 
 from ...monitoring.tests import TestMonitoringMixin
 from ..utils import manage_short_retention_policy
@@ -20,7 +21,9 @@ Config = load_model("config", "Config")
 Device = load_model("config", "Device")
 
 
-class TestDeviceMonitoringMixin(CreateConfigTemplateMixin, TestMonitoringMixin):
+class TestDeviceMonitoringMixin(
+    AssertNumQueriesSubTestMixin, CreateConfigTemplateMixin, TestMonitoringMixin
+):
     device_model = Device
     config_model = Config
     _PING = "openwisp_monitoring.check.classes.Ping"
--- a/openwisp_monitoring/monitoring/tests/test_api.py
+++ b/openwisp_monitoring/monitoring/tests/test_api.py
@@ -7,6 +7,7 @@ from swapper import load_model
 
 from openwisp_controller.config.tests.utils import CreateConfigTemplateMixin
 from openwisp_controller.geo.tests.utils import TestGeoMixin
+from openwisp_utils.tests import AssertNumQueriesSubTestMixin
 
 from ..configuration import DEFAULT_DASHBOARD_TRAFFIC_CHART
 from . import TestMonitoringMixin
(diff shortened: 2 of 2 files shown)


nemesifier (changes requested):
Have you checked that all the classes that use `assertNumQueries` are inheriting either `AssertNumQueriesSubTestMixin` or  `TestDeviceMonitoringMixin`? At first glance it doesn't look like it.

Chadokar:
@nemesifier , Yes I checked this repository-wide.
There are 35 call for `assertNumQueries()` across 4 files, and they are contained in 6 test classes. In the PR:

- `TestDashboardTimeseriesView` - inherits `AssertNumQueriesSubTestMixin` directly.
- `TestDeviceMonitoring` - inherits it through `DeviceMonitoringTestCase` -> `TestDeviceMonitoringMixin`.
- `TestWifiClientSession` - inherits it through `TestWifiClientSessionMixin `-> `TestDeviceMonitoringMixin.`
- `TestDeviceApi` - inherits it through `DeviceMonitoringTestCase` -> `TestDeviceMonitoringMixin`.
- `TestWifiSessionApi` - inherits it through `TestWifiClientSessionMixin` / `DeviceMonitoringTestCase `-> `TestDeviceMonitoringMixin`.
- `TestAdmin` - inherits it through `TestWifiClientSessionMixin `/ `DeviceMonitoringTestCase` -> `TestDeviceMonitoringMixin`.
I also checked the effective method resolution at runtime, not only the inheritance declarations. For all 6 classes:

```
issubclass(cls, AssertNumQueriesSubTestMixin) == True
cls.assertNumQueries is AssertNumQueriesSubTestMixin.assertNumQueries == True
```

So all current `assertNumQueries()` call sites resolve to `AssertNumQueriesSubTestMixin.assertNumQueries`. I don't see another query count class that needs the _mixin_ added separately.

Chadokar:
@nemesifier, just following up on this. I verified all the current assertNumQueries() call sites and the effective method resolution in my comment above. Could you please take another look when you have time? Thanks!

